PR #22, opened 2023-04-06
feat: load a map view

### LINKED ISSUE #19: Load map in home screen
- [ ] a map should be displayed when opening the home screen

shrouxm:
## Description
Loads a map view as the default screen in the app scaffold.

### Checklist

### Related Issues
Closes #19 

### Verification steps
When opening the app, a map view should be shown.


dev-client/ios/Podfile
```diff
@@ -48,7 +48,12 @@ target 'DevClient' do
     # Pods for testing
   end
```
shrouxm, line 50:
```suggestion
  pod 'MapboxMaps', '10.12.0'
```

shrouxm, line 50:
```suggestion

  $RNMapboxMapsImpl = 'mapbox'
```

shrouxm (commented):
@paulschreiber your commits are loading maplibre instead of mapbox, i think the suggestions i made are what are needed to load mapbox, could you confirm?

dev-client/ios/Podfile
shrouxm: i don't know if this one is necessary

Issue #44, opened 2023-05-02
Set up oauth login


shrouxm:
# Description

A user must be authenticated in order to perform most queries against the Terraso backend. This means that as a prerequisite to querying the backend we must provide some sort of way for the user to sign in an obtain a Terraso authorization token. In order to keep things simple, this will originally just be done for Google.

# Details

Originally I thought that we could reuse the OAuth signin process we use for the Terraso backend. However, it is not quite that easy. The redirects used in OAuth are complicated by the fact that the mobile app needs to switch between the browser view to approve the app, but needs to redirect to a native view afterwards. Unfortunately this means that we can't use the Terraso OAuth callback as our callback here -- we wouldn't be able to force the user back to the app and the user would end up at the Terraso website in their browser.

Unforunately, the alternative will be more complicated. What Google suggests doing is the following:

1) Set up OAuth identification using a public authorization flow in the native app. At the end of this stage we will have opened the user's default browser, closed it, and redirected the user back to the app. We will also have the authorization token from the server. In OIDC, this is a JWT.
2) Use this authorization token to retrieve a Terraso token from the backend. The Terraso backend can verify the properties of the JWT to ensure that the authorization token was in fact created by Google, create an account, etc. We then would create a new Terraso JWT, encrypted with our private JWT key, that would be used by the frontend for authenticating with our backend.

# Useful Links

[Google Docs on authenticating with a backend server](https://developers.google.com/identity/sign-in/android/backend-auth)
[React Native Docs on Deep Linking](https://reactnative.dev/docs/security#authentication-and-deep-linking)
[Expo Link on Deep Linking](https://docs.expo.dev/guides/linking/#handling-links)

shrouxm:
ok i will try to clarify my understanding of the two login flows:

### web client
1. client retrieves login URLs from the backend with the backend's oauth client configuration, including a redirect URI to the backend
2. client navigates to login URL and logs in
3. oauth provider redirects the browser back to Terraso at one of the {provider}/callback endpoints, passing a login token in URL params
4. the callback endpoint on the backend sends the login token back to the oauth provider along with its secret token, who confirms it is valid, & gives the user's info

### proposed mobile client flow
1. client performs an oauth sign in without communicating with the backend, retrieving a signed ID token with their login info.
2. client sends signed ID token to backend.
3. backend validates the token's signature & logs the user in using the login info provided.

### questions
[...]

PR #54, opened 2023-05-10
feat: Create beautiful but functionless login button

david-code:
## Description

This button is just an experiment to learn how to create components, install nativebase and customize the theme. The next steps will be to update this button so it actually returns logs the user in.

I've temporarily commented out the map view. As part of this feature, I'd like to work out how to figure out navigation, so the user is redirected to the map view after being logged in.

Making this a separate PR to check out the following things:
- Does the theme setup look reasonable?
- folder structure for components

Related to #44

### DIFF
--- a/dev-client/App.tsx
+++ b/dev-client/App.tsx
@@ -6,22 +6,33 @@
  */
 
 import React from 'react';
-import {StyleSheet, View} from 'react-native';
+import { StyleSheet, View } from 'react-native';
 
 import Mapbox from '@rnmapbox/maps';
+import { Button, NativeBaseProvider, Text } from 'native-base';
+import { theme } from './theme';
+import LoginView from './components/Login';
+
+
 
 Mapbox.setAccessToken(
   'pk.eyJ1Ijoic2hyb3V4bSIsImEiOiJjbGY4bW8wbGEwbDJnM3FsN3I1ZzBqd2kzIn0.2Alc4o911ooGEtnObLpOUQ',
 );
 
+
 function App(): JSX.Element {
   return (
-    <View style={styles.container}>
-      <Mapbox.MapView style={styles.map} />
-    </View>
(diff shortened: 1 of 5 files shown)


dev-client/App.tsx
```diff
@@ -6,22 +6,33 @@
...
 function App(): JSX.Element {
   return (
-    <View style={styles.container}>
-      <Mapbox.MapView style={styles.map} />
-    </View>
+    <NativeBaseProvider theme={theme}>
+      <View style={[styles.container, {
+        justifyContent: "center"
+      }]}>
+        { /* <Mapbox.MapView style={styles.map} /> */}
```
david-code, line 29: Commented this out for now. Will implement some kind of routing as part of this feature and uncomment before we merge into main.

dev-client/package.json
```diff
@@ -11,8 +11,11 @@
   },
   "dependencies": {
     "@rnmapbox/maps": "^10.0.0",
+    "native-base": "^3.4.28",
     "react": "18.2.0",
-    "react-native": "0.71.6"
+    "react-native": "0.71.6",
+    "react-native-safe-area-context": "^3.3.2",
+    "react-native-svg": "^12.1.1"
```
david-code, line 18: As per suggested in the [nativebase docs](https://docs.nativebase.io/install-rn#h3-install-dependencies)

david-code:
Question about running the Apple `pod install` or whatever needed to complete the installation of a new package with native integration: how/when do we want to do this?

shrouxm (approved):
LGTM! we can change the folder structure if it doesn't seem to be working, but that seems reasonable

shrouxm:
re `pod install`, i'm really not sure, i've been leaning on @paulschreiber for iOS help so far but i think it's your call at this point, i just don't have any way to experiment and form opinions

Issue #48, opened 2023-05-02
Display user's (& their project's) sites on the map view


shrouxm:
Depends on #28

shrouxm:
this is working, david implemented it i think

PR #55, opened 2023-05-12
feat: Update map view to display a beautiful marker

david-code:
## Description

This is a bit of a messy commit, as I was trying out a bunch of new libraries to try and get this to work. The main changes are:

- Moved the map component into it's own separate component
- Added a vector icon library to display icons that match the mockups
- Tried doing some tests, but I think we need to install some coco pods through xcode or something. I will create a chore task and try and get someone to do this.
- Ran prettier on everything, so there's a lot of noise from just styling edits
- Displayed the user location on the map, which requires asking the user for permission on newer versions of Android.

I started working on this issue because I realized that the login for mobile is more difficult than I anticipated (see #44 for details). It would be nice to have some mockups implemented for our demo next week so I decided to focus on building out the frontend a bit more. My idea is to mock out the backend calls to start off with. This will also give us the benefit of being able to use the work that's being done on the GraphQL client when it's ready.

### Related Issues

Related to #46 #48

## Followup issues

- [x] Issue to add information on how to find material IO vector icons to the README #56
- [x] Change MarkerView to a better solution that is automatically clustered: https://github.com/rnmapbox/maps/blob/main/docs/Annotations.md #57
- [x] Set up iOS to use react-native-vector-icons as well

### DIFF
--- a/dev-client/App.tsx
+++ b/dev-client/App.tsx
@@ -5,34 +5,58 @@
  * @format
  */
 
-import React from 'react';
-import { StyleSheet, View } from 'react-native';
+import React, {useEffect} from 'react';
+import {PermissionsAndroid, StyleSheet, View} from 'react-native';
 
 import Mapbox from '@rnmapbox/maps';
-import { Button, NativeBaseProvider, Text } from 'native-base';
-import { theme } from './theme';
+import {Button, NativeBaseProvider, Text} from 'native-base';
+import {theme} from './theme';
 import LoginView from './components/Login';
-
-
+import SiteMap from './components/SiteMap';
+import {DisplaySite} from './datatypes/sites';
 
 Mapbox.setAccessToken(
   'pk.eyJ1Ijoic2hyb3V4bSIsImEiOiJjbGY4bW8wbGEwbDJnM3FsN3I1ZzBqd2kzIn0.2Alc4o911ooGEtnObLpOUQ',
 );
 
-
 function App(): JSX.Element {
+  useEffect(() => {
+    const locationGranted = PermissionsAndroid.request(
+      PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION,
+    )
+      .then(granted => {
+        if (granted == PermissionsAndroid.RESULTS.GRANTED) {
+          console.log('You can use the camera');
+        } else {
+          // TODO: What to do if rejected?
+          console.error('You cannot use the camera');
+        }
+      })
+      .catch(err => {
+        // TODO: What to do on error?
+        console.error(err);
+      });
+  });
+
   return (
     <NativeBaseProvider theme={theme}>
-      <View style={[styles.container, {
-        justifyContent: "center"
-      }]}>
-        { /* <Mapbox.MapView style={styles.map} /> */}
-        <LoginView />
+      <View
+        style={[
+          styles.container,
+          {
+            justifyContent: 'center',
+          },
+        ]}>
+        <SiteMap
+          sites={[new DisplaySite(0.1, 0.1, 'TEST SITE 1')]}
+          center={[0, 0]}
+        />
+        {/* <LoginView /> */}
       </View>
     </NativeBaseProvider>
   );
 }
 
-const styles = StyleSheet.create({ map: { flex: 1 }, container: { flex: 1 } });
+const styles = StyleSheet.create({map: {flex: 1}, container: {flex: 1}});
 
 export default App;
--- a/dev-client/App.tsx
+++ b/dev-client/App.tsx
@@ -6,7 +6,7 @@
  */
 
 import React, {useEffect} from 'react';
-import {PermissionsAndroid, StyleSheet, View} from 'react-native';
+import {PermissionsAndroid, Platform, StyleSheet, View} from 'react-native';
 
 import Mapbox from '@rnmapbox/maps';
 import {Button, NativeBaseProvider, Text} from 'native-base';
@@ -21,6 +21,10 @@ Mapbox.setAccessToken(
 
 function App(): JSX.Element {
   useEffect(() => {
+    if (Platform.OS !== 'android') {
+      // only need to request location permissions explicitly on Android
+      return;
+    }
     const locationGranted = PermissionsAndroid.request(
       PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION,
     )
@@ -48,7 +52,10 @@ function App(): JSX.Element {
           },
         ]}>
         <SiteMap
-          sites={[new DisplaySite(0.1, 0.1, 'TEST SITE 1')]}
+          sites={[
+            {lat: 0.0, lon: 0.3, name: 'TEST SITE 1'},
+            {lat: 0.1, lon: 0.5, name: 'TEST SITE 2'},
+          ]}
           center={[0, 0]}
         />
         {/* <LoginView /> */}
--- a/dev-client/android/app/build.gradle
+++ b/dev-client/android/app/build.gradle
@@ -170,3 +170,13 @@ dependencies {
 }
 
 apply from: file("../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesAppBuildGradle(project)
+
+// NOTE: This is to add vector icons to our app
+// We can add more here if we ever needed to, but as we are using the Material UI, probably won't need to
+// See https://github.com/oblador/react-native-vector-icons#android
+// REVIEW: Will need to do the same process with iOS
+project.ext.vectoricons = [
+    iconFontNames: [ 'MaterialIcons.ttf' ] // Name of the font files you want to copy
+]
+
+apply from: "../../node_modules/react-native-vector-icons/fonts.gradle"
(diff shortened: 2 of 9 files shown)


dev-client/android/app/build.gradle
```diff
@@ -170,3 +170,13 @@ dependencies {
 }
 
 apply from: file("../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesAppBuildGradle(project)
+
+// NOTE: This is to add vector icons to our app
+// We can add more here if we ever needed to, but as we are using the Material UI, probably won't need to
+// See https://github.com/oblador/react-native-vector-icons#android
+// REVIEW: Will need to do the same process with iOS
```
david-code, line 177: I want to make an issue for dealing with these kinds of chores. I think we should have a process for how to deal with this stuff.

dev-client/components/SiteMap.tsx
```diff
@@ -0,0 +1,47 @@
...
+// TODO: This vector icons module appears to be missing types
+import VIcon from 'react-native-vector-icons/MaterialIcons';
+
+type SiteMapProps = {
+  sites: DisplaySite[];
+  center: Position;
+};
+
+// REVIEW: This is a bad name, any suggestions?
+export default function SiteMap({sites, center}: SiteMapProps): JSX.Element {
```
david-code, line 16: Anyone have any better ideas for the name here?

dev-client/components/SiteMap.tsx
```diff
@@ -0,0 +1,47 @@
...
+      style={{
+        flex: 1,
+      }}>
+      <Camera ref={camera} />
+      <UserLocation />
+      {sites.map(site => {
+        return (
+          <MarkerView coordinate={[site.lon, site.lat]} key={site.key}>
+            <IconButton
+              icon={<Icon as={VIcon} name="location-on" />}
```
david-code, line 36: This is somewhat confusing. I don't worry about it here, but I think it would be nice to have a component to deal with this stuff for us, because we will be doing it a lot.

What this code is doing:
- `IconButton` is a nativebase component
- It uses `Icon` as the icon component, another nativebase component. The main advantage here is that we can pass our theme values to the icon.
- Finally, `VIcon` is the `react-native-vector-icon` component. It takes the `name` prop and displays the icon we want to display.

I think actually I should add the info about how to pick out the icon to the README.

dev-client/android/app/build.gradle
shrouxm: in https://github.com/techmatters/terraso-web-client/pull/934 i am keeping a list of issues to file in the PR itself, and i will file them as chore issues just before merging, does that seem reasonable?

dev-client/components/SiteMap.tsx
shrouxm: somewhat broader answer, but i was thinking that we should create a lower-level reusable component which encodes a lot of the logic that will be shared between the different map views (e.g. how to handle the camera, setting user location, etc)

i think this would be helpful because the rnmapbox docs are pretty lacking, and if we could wrap all of our usage of rnmapbox in one component with fairly generic props, and then just call out to that component, we will limit the amount we need to read those docs, and also if the library stops being maintained there will be a clear API boundary between our code & the library

this generic component might have props like:
```ts
type TerrasoMapViewProps<T> = {
  ...,
  annotationData: T[],
  renderIcon: (_:T) => JSX.Element,
  ...
}
```

but we could do that in a future issue

dev-client/components/SiteMap.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <Mapbox.MapView
+      style={{
+        flex: 1,
+      }}>
+      <Camera ref={camera} />
+      <UserLocation />
+      {sites.map(site => {
+        return (
+          <MarkerView coordinate={[site.lon, site.lat]} key={site.key}>
```
shrouxm, line 34: question: did you look into alternative implementations than `MarkerView`? (see https://github.com/rnmapbox/maps/blob/main/docs/Annotations.md). i think `MarkerView` is fine for now and we can just swap it out if we need to, just curious if you tried any of the others as i think we may need to upgrade from `MarkerView` eventually. (e.g. if we want to do point clustering without writing the clustering algorithm ourselves i don't think we could use `MarkerView`)

dev-client/components/SiteMap.tsx
shrouxm: suggestion: could we create a very thin abstraction over this and document what it's doing there? then our calls can be a bit less repetitive if they look like `<MyIconButton icon="location-on" size="sm" ... />

dev-client/App.tsx
```diff
@@ -5,34 +5,58 @@
...
+        } else {
+          // TODO: What to do if rejected?
+          console.error('You cannot use the camera');
+        }
+      })
+      .catch(err => {
+        // TODO: What to do on error?
+        console.error(err);
+      });
+  });
```
shrouxm, line 39: question: do we need to do this here as well as declaring the permissions in the manifest?

dev-client/datatypes/sites.ts
```diff
@@ -0,0 +1,17 @@
+export type Point2D = [number, number];
```
shrouxm, line 1: nit: i think this is unused?

dev-client/datatypes/sites.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    constructor(lat: number, lon: number, name: string) {
+        this.lat = lat;
+        this.lon = lon;
+        this.name = name;
+    }
+
+    get key() {
+        return `{this.lat},{this.lon},{this.name}`;
+    }
```
shrouxm, line 16: question: what is this for?

dev-client/datatypes/sites.ts
```diff
@@ -0,0 +1,17 @@
+export type Point2D = [number, number];
+
+export class DisplaySite {
+    lat: number;
+    lon: number;
```
shrouxm, line 5: question: can we re-use a `coordinate` type from https://www.npmjs.com/package/@types/geojson? `rnmapbox` also depends on this library and expects this format so it might make interop code a bit nicer

dev-client/datatypes/sites.ts
```diff
@@ -0,0 +1,17 @@
+export type Point2D = [number, number];
+
+export class DisplaySite {
```
shrouxm, line 3: question: in the web client we haven't been using classes, this type would've been written as
```ts
type DisplaySite = {
  lat: number;
  lon: number;
  name: string;
}
```
if you're indifferent, i think we should also avoid classes in this repo for consistency

shrouxm (approved):
looks good overall! i suspect there will be much trial & error in figuring out how to coax these libraries to do what we want. i left some questions & style suggestions, but if it's too much work to address all of them right now you can skip some since i'd rather we move fast right now.

dev-client/android/app/build.gradle
david-code: Yes, good idea Garo.

dev-client/components/SiteMap.tsx
david-code: Totally agree! Personally my idea was to just start out with a clunky component and refactor it when I have a bit more experience using it, seeing what works and what doesn't. However if you have a clear idea with what you want, I think there is a separate issue to work on the reusable map component you could grab.

dev-client/components/SiteMap.tsx
david-code: Yes, I ended up doing this in my next PR because it was getting annoying.

dev-client/App.tsx
david-code: Yes, for Android there are a few permissions that need to be requested explicitly: [AndroidPermissions](https://reactnative.dev/docs/permissionsandroid).

> PermissionsAndroid provides access to Android M's new permissions model. The so-called "normal" permissions are granted by default when the application is installed as long as they appear in AndroidManifest.xml. However, "dangerous" permissions require a dialog prompt. You should use this module for those permissions.

dev-client/components/SiteMap.tsx
david-code: No, good point. I'll add this as another issue.

dev-client/datatypes/sites.ts
david-code: Yes, I deleted this in my next PR.

dev-client/datatypes/sites.ts
david-code: This is to generate a unique key for rendering in a list with React. @shrouxm Any idea how this is currently done in the frontend?

dev-client/datatypes/sites.ts
david-code: Yes, agreed. I ended up doing this in my next PR.

dev-client/datatypes/sites.ts
david-code: Agreed, let's keep it consistent.

dev-client/datatypes/sites.ts
shrouxm: ohh right, hmm no idk. i would think it would eventually be the object's pk if possible, but not sure what we're doing currently

Issue #51, opened 2023-05-03
Create site from "project sites" list


shrouxm:
Depends on #33, #53

Issue #39, opened 2023-05-02
Implement flow for adding existing site to project


ltseng:

ltseng:
This task has been in review for two weeks. @JeancarloBarrios are you reviewing this, and if so, what's the blocker?

Issue #38, opened 2023-05-02
Implement flow for adding new site to project


ltseng:
Dependent on Creating a Site user story to be completed first

PR #81, opened 2023-06-09
feature: Change app bar

david-code:
## Description

This PR changes the app to use the [react-navigation Native Stack Navigator](https://reactnavigation.org/docs/native-stack-navigator) header instead of using a custom component.

Implementing the AppBar was one of the first things I did, before I tried using react-navigation. It's become clear to me that it makes a lot more sense to style the react-navigation header, which is quite flexible in this respect, than to have our own custom component.

Most of the changes here are altering the way Screens are created slightly and playing a bit with react navigation config to get the results I want. Doing this before I get started with #38, #39, #51 etc. because this will be needed there.


dev-client/screens/AppScaffold.tsx
```diff
@@ -1,55 +1,87 @@
...
+  const {colors} = useTheme();
 
-  function filterLogin([name, _]: ScreenMapArgs) {
+  function filterLogin([_name, config]: ScreenMapArgs) {
     if (user === null) {
-      return LoggedOut.has(name);
+      return config.loggedOut === true;
     } else {
-      return !LoggedOut.has(name);
+      return config.loggedOut !== true;
```
shrouxm, line 71: nit:
`return config.loggedOut`
`return !config.loggedOut`

shrouxm (approved):
LGTM!

PR #83, opened 2023-06-13
fix: add Material icon fonts to IOS build

paulschreiber:
## Description
Update the xcode build configuration to  include fonts for iOS. Also updates configuration files.


david-code (approved):
Looks good!

PR #114, opened 2023-06-22
fix: Update metro config

david-code:
I was getting the following warning when I ran `npm run start`:

warn From React Native 0.72, your metro.config.js file should extend'@react-native/metro-config'.

Metro would then crash. I obliged, and this commit is the result. Seems to fix things.

### DIFF
--- a/dev-client/metro.config.js
+++ b/dev-client/metro.config.js
@@ -1,17 +1,11 @@
-/**
- * Metro configuration for React Native
- * https://github.com/facebook/react-native
- *
- * @format
- */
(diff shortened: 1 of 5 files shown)


paulschreiber (approved):

david-code:
@paulschreiber Could you look at the ios failure here?

paulschreiber:
@david-code this is a GitHub Actions caching bug. You can ignore it (or delete the [CocoaPods cache](https://github.com/techmatters/terraso-mobile-client/actions/caches) and re-run.

PR #122, opened 2023-06-29
fix: Install library needed for uuid library

david-code:
## Description

Got an error trying to make a site on the map, so I followed the instructions displayed in the error and it fixed it.

### DIFF
--- a/dev-client/src/App.tsx
+++ b/dev-client/src/App.tsx
@@ -18,6 +18,8 @@ import {PermissionsAndroid} from 'react-native';
 import {Provider} from 'react-redux';
 import {createStore} from '../model/store';
 import {setAPIConfig} from 'terraso-client-shared/config';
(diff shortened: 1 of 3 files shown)


dev-client/src/App.tsx
```diff
@@ -18,6 +18,8 @@ import {PermissionsAndroid} from 'react-native';
 import {Provider} from 'react-redux';
 import {createStore} from '../model/store';
 import {setAPIConfig} from 'terraso-client-shared/config';
+// needed for uuid - https://github.com/uuidjs/uuid#react-native--expo
+import 'react-native-get-random-values';
```
shrouxm, line 22: suggestion: the docs state that the import should happen at the top of the import list

shrouxm (approved):
just one small comment! approving so you don't have wait on me once it's fixed

PR #123, opened 2023-06-29
fix: pass --fix to eslint

paulschreiber:
## Description
Update package.json's `lint` command to pas `--fix`.

### DIFF
--- a/dev-client/package.json
+++ b/dev-client/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "android": "react-native run-android --list-devices",
     "ios": "react-native run-ios",
-    "lint": "eslint .",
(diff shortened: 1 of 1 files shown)


david-code:
Will this mess up the CI lint checking if it fixes an issue in the CI instead of reporting it? Maybe it would make more sense to have another command `lint-fix`, like the `format` and `lint` commands on the backend.

shrouxm:
i personally don't like having `--fix` as the default option on linting, because it can easily be passed in as an option and the default of not rewriting files is less disruptive to me, but i don't feel hardline about it

paulschreiber:
The web client has this. It would be good to be consistent across projects. I am happy to have lint and format as separate commands (for both).

Issue #121, opened 2023-06-28
Set up iOS for OAuth signin


david-code:
[docs](https://github.com/FormidableLabs/react-native-app-auth#ios-setup)

PR #126, opened 2023-06-29
feature: Signin

david-code:
# Description

This PR changes the mobile client to 
1) actually show a login screen when the user has not signed in
2) implement
3) make a query to a backend endpoint to exchange the OpenID Connect ID token for Terraso access tokens

I am not merging directly to main, but am making a separate feature branch.
- Will need to finish #121 for this to work for iOS
- Until the login token is persisted (plan to do this next PR), developers will have to log in every time the app builds, which would be very annoying.
- Uses the login context I created for testing a while back. Should use Redux instead.
- [x] Need to change the app.gradle custom url schema work to use new package name (`app.terraso.landpks`), and update the Google API data as well.


dev-client/components/Login.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import Config from 'react-native-config';
+import {exchangeToken} from '../src/auth';
+import {useLoginDispatch} from '../src/context/LoginContext';
+
+// https://github.com/FormidableLabs/react-native-app-auth/blob/main/docs/config-examples/google.md
+const googleConfig = {
+  issuer: 'https://accounts.google.com',
+  clientId: `${Config.GOOGLE_OAUTH_APP_GUID}.apps.googleusercontent.com`,
+  redirectUrl: 'com.devclient:/oauth2redirect',
```
shrouxm, line 13: question: where do we configure how we handle this redirect URL?

dev-client/components/Login.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const onPress = useCallback(() => {
+    authorize(googleConfig)
+      .then(async result => {
+        let {atoken, rtoken, email, firstName, lastName} = await exchangeToken(
+          result.idToken,
+          'google',
+        );
+        loginDispatch({type: 'login', name: firstName});
+      })
+      .catch(e => console.error(e));
```
shrouxm, line 28: suggestion: can we move the call to `authorize` into `exchangeToken`?

dev-client/components/Login.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }, [googleConfig, authorize]);
```
shrouxm, line 29: nit: i think `googleConfig` and `authorize` will never change, can we omit them from the keys? or does linter complain even tho they're consts?

dev-client/src/auth/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+    body: JSON.stringify({
+      provider,
+      jwt: identityJwt,
+    }),
+  });
```
shrouxm, line 22: suggestion: the `request` method in `terraso-client-shared/terrasoApi/api` already does this request plumbing

dev-client/src/auth/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (
+    checkFields(['atoken', 'rtoken', 'user'], payload) ||
+    checkFields(['email', 'firstName'], payload.user)
+  ) {
+    // TODO: handle error
+    console.error(payload);
+    throw 'Bad token JSON';
+  }
+
+  // TODO: Just doing this to get the right types, has to be a better way
```
shrouxm, line 38: suggestion: you can give the type of the request JSON as a type argument to the `request` method i mentioned above. unfortunately there is a longstanding missing feature of TypeScript to validate that a runtime value conforms to a TypeScript type, so it won't actually _check_ the fields, but it at least lets TypeScript know what they're supposed to be

dev-client/src/auth/index.ts
shrouxm: note you will need to provide the `Config.TERRASO_BACKEND` value in the `setApiConfig` call, since that's where `request` gets the backend URL from

dev-client/src/auth/index.ts
```diff
@@ -0,0 +1,46 @@
+import Config from 'react-native-config';
```
shrouxm, line 1: suggestion: can we create a `config.ts` file like in the web client instead of directly accessing the env variables?

shrouxm (commented):
looks good overall, could we just merge to main so this code doesn't get too out of step if we do some other refactor/code style-y things on main? (i would like to do absolute URLs soon)

dev-client/components/Login.tsx
david-code: answer: the scheme name is handled by the changes to [build.gradle](https://github.com/techmatters/terraso-mobile-client/blob/50a3cd16e0042962b0f0c14c407b4bee94e64289/dev-client/android/app/build.gradle#L104). I need to update this actually so that the package name is `org.terraso.landpks`, I will add a task to do so. Ideally this would be at least a constant value for when we add other OAuth providers as well.

As to the route path: I am just following the react-native-app-auth docs I linked in the comment above. I checked the [Google docs](https://developers.google.com/identity/protocols/oauth2/native-app#step-2:-send-a-request-to-googles-oauth-2.0-server) as well, and it says the path part is optional, so idk. I might try and see if it works without the `oauth2redirect`, might not.

dev-client/src/auth/index.ts
david-code: OK, I'll try that out.

shrouxm:
@david-code did you forget to push changes? i see another review request but no new commits

PR #184, opened 2023-07-12
chore: use JSON for prettierrc

paulschreiber:
## Description
Use JSON for prettierrc. Match web client and shared client.


shrouxm (approved):
LGTM

PR #188, opened 2023-07-13
feat: home screen bottom sheet

### LINKED ISSUE #50: Create site from home screen sites list view
Depends on #33, #52

### LINKED ISSUE #52: Create home screen sites list view
If done before techmatters/terraso-client-shared#17, create an issue to wire it up to the API once techmatters/terraso-client-shared#17 is done.

shrouxm:
## Description
Adds a bottom sheet to the home screen.

### Checklist

### Related Issues
Closes #50 
Closes #52

### Verification steps
Sites should be visible in the bottom sheet. Clicking "show on map" should pan the map to the site. There is no location dashboard yet so clicking the site does nothing. Also updates the site map to not persist temporary sites, so there is at most one active at a time.

### DIFF
--- a/dev-client/src/auth/index.ts
+++ b/dev-client/src/auth/index.ts
@@ -27,6 +27,7 @@ export async function exchangeToken(
     headers: {'content-type': 'application/json'},
   });
 
+  console.log(payload);
   return {
     atoken: payload.atoken,
     rtoken: payload.rtoken,
--- a/dev-client/src/auth/index.ts
+++ b/dev-client/src/auth/index.ts
@@ -27,7 +27,6 @@ export async function exchangeToken(
     headers: {'content-type': 'application/json'},
   });
 
-  console.log(payload);
   return {
     atoken: payload.atoken,
     rtoken: payload.rtoken,
--- a/dev-client/src/components/home/BottomSheet.tsx
+++ b/dev-client/src/components/home/BottomSheet.tsx
@@ -6,45 +6,65 @@ import {Site} from 'terraso-client-shared/site/siteSlice';
 import {ScreenRoutes} from '../../screens/constants';
 import {TopLevelNavigationProp} from '../../screens';
 import {useSelector} from '../../model/store';
+import {useTranslation} from 'react-i18next';
+
+type SiteListSiteProps = {
+  site: Site;
+  showSiteOnMap: (site: Site) => void;
+};
+const SiteListSite = ({site, showSiteOnMap}: SiteListSiteProps) => {
+  const {t} = useTranslation();
+  const project = useSelector(state =>
+    site.projectId === undefined
+      ? undefined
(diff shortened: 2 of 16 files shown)


shrouxm:
@paulschreiber can you run `pod install`? i think that is the only necessary iOS install step for the dependencies i added

dev-client/src/auth/index.ts
```diff
@@ -27,6 +27,7 @@ export async function exchangeToken(
     headers: {'content-type': 'application/json'},
   });
 
+  console.log(payload);
```
paulschreiber, line 30:
```suggestion
```

dev-client/src/components/home/BottomSheet.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <Heading size="xl">Sites</Heading>
+          <Button size="lg" onPress={addSiteCallback}>
+            CREATE SITE
+          </Button>
+        </Flex>
+        {siteList.length === 0 && (
+          <>
+            <Text fontSize="md">
+              You don’t currently have any saved sites. Create a site by using
+              the button above or by finding the site location on the map.
```
paulschreiber, line 76: issue: text should be localized.

dev-client/src/components/home/BottomSheet.tsx
```diff
@@ -0,0 +1,95 @@
...
+            </Text>
+            <Text />
+            <Text fontSize="md">
+              A site allows you to collect data, learn more about your land, and
+              share with others. Learn more about using LandPKS...
```
paulschreiber, line 81: issue: text should be localized.

paulschreiber (changes requested):

dev-client/src/components/home/BottomSheet.tsx
```diff
@@ -0,0 +1,109 @@
...
+          <Heading size="xl">{t('site.list_title')}</Heading>
+          <Button size="lg" onPress={addSiteCallback}>
+            {t('site.create')}
+          </Button>
+        </Flex>
+        {siteList.length === 0 && (
+          <>
+            <Text fontSize="md">{t('site.none_existing_p1')}</Text>
+            <Text />
+            <Text fontSize="md">{t('site.none_existing_p2')}</Text>
```
paulschreiber, line 96: question what do `none_existing_p1` and `none_existing_p2` mean? Can we give them more descriptive names? If these are just different paragraphs for the same message, can we use <Trans> to consolidate strings?

shrouxm: they are two paragraphs of the same message. i couldn't find a way to get it to render properly when it was all one string, but that was before i i18n'd it so maybe there is a way i will check

shrouxm: done

PR #211, opened 2023-07-19
feat: refactor screen definitions

### LINKED ISSUE #119: refactor: componentize creation of app screens
## Description
Currently we factor out the common code of the different main app screens with config objects and functions spread around a few files. We should consolidate these and make the config objects props of a common `AppScreen` (or some such name) component for simplicity and ease of navigation.

shrouxm:
## Description
Refactors the root navigation stack to a more component oriented architecture. This enables us to use hooks when rendering the app bar's title to e.g. retrieve an objects name from the redux store. I added a site dashboard screen with no content just to make sure it works.

I did NOT change any navigators besides the root navigator, since this PR is already getting bigger. I will need to create another tab navigator when I actually create the site dashboard, so I will figure out what that would look like with this new architecture and then update other navigators to match.

### Checklist

### Related Issues
Fixes #119 

### Verification steps
Everything should work the same (except the new empty site dashboard screen).


dev-client/src/screens/AppScaffold.tsx
```diff
@@ -1,80 +1,137 @@
...
 
-  function filterLogin([_name, config]: ScreenMapArgs) {
-    if (currentUser === null) {
-      return config.loggedOut === true;
-    } else {
-      return config.loggedOut !== true;
-    }
-  }
+  const filterLogin = (name: ScreenName) =>
+    (currentUser === null) === (name === 'LOGIN');
```
david-code, line 134: nit: Do you think it makes sense to keep more than just one "logged out" screen? I think we will likely have several screens where the user doesn't have to log in - but maybe we can just deal with that as it arises.

dev-client/src/screens/AppScaffold.tsx
```diff
@@ -1,80 +1,137 @@
...
+    const CastTitle = HeaderTitle as HeaderTitleType<object> | undefined;
+    return {
+      ...origOptions,
+      headerTitle: ({tintColor}) =>
+        CastTitle !== undefined ? (
+          <CastTitle {...props.route.params} tintColor={tintColor} />
+        ) : (
+          <DefaultHeader name={name} tintColor={tintColor} />
+        ),
+    };
```
david-code, line 99: question: OK, do you think there is a benefit to being able to define the header like this? The way it was implemented previously, any params for the title name could be passed to the screen as route params. We could just have the translation function do `t("screen.name", route.params.titleParams)`. I think this is cleaner than having the screen itself create its own name by looking at global state, especially if we want to have some sort deep linking eventually. I'm curious what you think!

david-code (commented):
Looks good! I have a question about the whole Header/DefaultHeader and using the redux state to produce the screen name, curious what you think.

dev-client/src/screens/AppScaffold.tsx
shrouxm: i think we need to be able to compute it from the redux state regardless because e.g. if the user changes the site's name from the site's dashboard it won't update in the header if the name was passed as a screen parameter. same if someone changes the name remotely and they get synced to the client while they are on a screen with a dynamically computed title

dev-client/src/screens/AppScaffold.tsx
shrouxm: i'm not sure what this will look like eventually, because the actual intended workflow is that the whole app should be usable without logging in right?

shrouxm: (minus collaborative features)

david-code: Yeah, I was just thinking there will always be a bunch of screens where you have to login for them to make sense. But let's not worry about that now.

dev-client/src/screens/AppScaffold.tsx
david-code: OK good point. I agree, we can come back if this starts causing problems.

david-code (approved):

PR #214, opened 2023-07-24
fix: ios auth

jodiezhuu:
## Description
fixed iOS auth code crash with exception `Thread 1: "<AppDelegate: 0x600003df21c0> does not conform to RNAppAuthAuthorizationFlowManager"`


david-code (approved):

paulschreiber:
@david-code The CI error is a problem with the GitHub Acrtion cache. Deleting `macos-pods-xxxx` and re-running the action should clear it up.

PR #216, opened 2023-07-25
feat: location dashboard

### LINKED ISSUE #155: Create location dashboard
## Description
Create the location dashboard as the navigation target when clicking on a site from home screen (pin callout or bottom sheet list) or project sites list view. See designs for details.

shrouxm:
## Description
Implements much of the location dashboard design, as well as a random assortment of unrelated small UI improvements I was too lazy to pull out into separate PRs. I will annotate these.

### Checklist
- [x] Navigate to location dashboard from several places

### Related Issues
Fixes #155 

### Verification steps
Tapping a site in various locations


dev-client/src/components/projects/CreateProjectView/Form.tsx
```diff
@@ -1,9 +1,8 @@
-import {HStack, Input, Text, VStack} from 'native-base';
+import {Fab, HStack, Input, Text, VStack} from 'native-base';
```
shrouxm, line 1: note: the `SaveFAB` was mainly used for styling, i have done this styling in our theme instead of a component so we can directly use the `Fab` component. if we have need for more than one FAB style we can figure out how to use variants.

dev-client/src/components/sites/LocationDashboardScreen.tsx
```diff
@@ -0,0 +1,89 @@
...
+      const {navigate} = useNavigation();
+      return (
+        <IconButton
+          name="settings"
+          _icon={{color: tintColor}}
+          onPress={() => navigate('SITE_SETTINGS', {siteId})}
+        />
+      );
+    },
+    headerTitle: props => {
```
shrouxm, line 83: note: i realized i got myself in a bit of a twist while doing the screen refactor, so i'm no longer wrapping the `HeaderTitle` arg because it was unnecessary, and we would've had to wrap all the other args (like `headerRight`) in the same way which would've been quite annoying. this change has been propogated everywhere, the actual change is visible in the (now much simpler) `AppScaffold.tsx` code

dev-client/src/screens/AppScaffold.tsx
```diff
@@ -80,48 +72,28 @@ const defineScreen = <
   Props extends RootStackParamList[Name],
 >(
   name: Name,
-  {options, View}: ScreenDefinition<Props>,
+  {options = () => ({}), View: ScreenView}: ScreenDefinition<Props>,
```
shrouxm, line 75: todo: i think this variable alias is no longer necessary

dev-client/src/screens/AppScaffold.tsx
```diff
@@ -130,8 +102,16 @@ export default function AppScaffold() {
   // using theme hook here because react-navigation doesn't take nativebase utility props
   const {colors} = useTheme();
 
-  const filterLogin = (name: ScreenName) =>
-    (currentUser === null) === (name === 'LOGIN');
```
shrouxm, line 134: note: i was poking around here trying to improve navigation performance by trying not to pass different props into `RootStack.Navigator`. hasn't fixed everything but i've noticed some improvements i think? and i think this is more idiomatic. there is associated logic for this change in the `LoginScreen` (if the LoginScreen renders and there is a current user, navigates to the home screen)

dev-client/src/screens/HomeScreen.tsx
```diff
@@ -29,22 +28,17 @@ const HomeView = () => {
...
-        <BottomSheet
-          sites={sites}
-          showSiteOnMap={site => setMapCenter([site.longitude, site.latitude])}
-        />
-      </Box>
-      <Box flexBasis="10%">
-        <BottomNavigation />
-      </Box>
-    </VStack>
+    <ScreenScaffold>
```
shrouxm, line 31: note: i created a `ScreenScaffold` component to just do the wrapping of screens in a bottom sheet, since we were repeating that logic everywhere. perhaps there will be other useful parameters or logic for the screen scaffold to perform one day, perhaps not!

dev-client/src/screens/LoginScreen.tsx
```diff
@@ -1,13 +1,26 @@
...
-const LoginView = () => (
-  <Center height="100%">
-    <Login />
-  </Center>
-);
+  useEffect(() => {
+    if (loggedIn) {
+      navigation.replace('HOME');
+    }
+  });
```
shrouxm, line 17: note: we run this effect on every render

shrouxm: i should probably make this note a comment

shrouxm: i should probably have omitted the word "probably" in the previous comment

dev-client/src/components/sites/LocationDashboardScreen.tsx
```diff
@@ -0,0 +1,89 @@
+import {ScreenDefinition, useNavigation} from '../../screens/AppScaffold';
```
shrouxm, line 1: note: i am starting to make screen files not in the `screens` folder. there are going to be a lot of screens, especially when we get to e.g. data inputs and sub screens, so i think we were gonna have to split them up at some point to avoid that folder getting monstrous

shrouxm: i'm not sure if screen groups could be helpful here? i don't really understand what their point is so i'm not sure

dev-client/src/theme.ts
```diff
@@ -1,37 +1,94 @@
...
-const newColorTheme = {
-  primary: {
-    main: '#276749',
-    contrast: '#FFFFFF',
-    lightest: '#9AE6B4',
-    // TODO: This is used for the colorScheme value for Radio
-    // We should figure out how the color scheme stuff works and see if we can
-    // map our current variables to 100, 200 values etc.
-    600: '#276749',
+export const theme = extendTheme({
```
shrouxm, line 3: note: i had quite a bit of trouble figuring out the right format for all these theme fields. i'm not sure i understand well enough to write a helpful comment yet, but hopefully i will be able to at some point. lmk if you have any questions while modifying/adding to this code!

dev-client/src/screens/HomeScreen.tsx
david-code: Yes, good idea, I was thinking the same thing recently.

david-code (approved):
Looks good to me, lots of nice little refactorings too!

PR #238, opened 2023-08-04
feat: delete project

jodiezhuu:
## Description
project screen for delete project


david-code (approved):
Merged with main and tested locally, works well!

PR #250, opened 2023-08-07
feat: Add config vars to release steps

david-code:
We need to add some config variables to each deploy. Most of them, for now, have to do with setting up the OAuth login and connecting to the backend.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


paulschreiber (changes requested):
Please delete this PR and make a new one. It has 2925 files, and it slowing down GitHub.

To create the .env file, something like this is simpler:
```
echo "FOO=${{ secrets.FOO }}" >> $HOME/.env
```

PR #269, opened 2023-08-14
feat: home screen styling fixes

shrouxm:
## Description
A bunch of styling fixes for the home screen. Most notably, makes the site cards not look terrible.

### DIFF
--- a/dev-client/src/components/home/BottomSheet.tsx
+++ b/dev-client/src/components/home/BottomSheet.tsx
@@ -1,61 +1,11 @@
 import BottomSheet, {BottomSheetFlatList} from '@gorhom/bottom-sheet';
 import {useNavigation} from '../../screens/AppScaffold';
-import {Box, Button, Flex, Heading, Text, Badge} from 'native-base';
+import {Box, Button, Flex, Heading, Text, useTheme} from 'native-base';
 import {useCallback, useMemo} from 'react';
 import {Site} from 'terraso-client-shared/site/siteSlice';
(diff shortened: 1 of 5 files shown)


dev-client/src/components/home/BottomSheet.tsx
```diff
@@ -83,16 +33,30 @@ const SiteListBottomSheet = ({sites, showSiteOnMap}: Props) => {
     navigation.navigate('CREATE_SITE');
   }, [navigation]);
 
+  const {colors} = useTheme();
+  const style = useMemo(() => ({paddingHorizontal: 16}), []);
```
david-code, line 37: question: what is the use of `useMemo` here?

dev-client/src/theme.ts
```diff
@@ -35,6 +37,24 @@ export const theme = extendTheme({
...
+      },
+    },
+    Badge: {
+      variants: {
+        chip: {
+          borderRadius: '100px',
+          padding: '4px',
+        },
+      },
+    },
```
david-code, line 57: comment: Nice, I love this.

david-code (approved):
Looks good! I really like using the theme to style different variants of component.

dev-client/src/components/home/BottomSheet.tsx
shrouxm: it's (supposed) to avoid re-renders, since if you pass in a raw object on each render, it will pass a new object in and cause a re-render (i think)

PR #270, opened 2023-08-15
feat: Update site list empty display updated

### LINKED ISSUE #251: Update bottom sheet for empty sites list
Update the placeholder text for an empty bottom sheet when there are no sites created

david-code:
## Description

This PR updates the Site bottom sheet to match the current wire frame when the user has not created any sites yet.

I am running into two annoying problems:

1) The link and icon render above where I would like to in the inline text.
2) The bottom menu bar hides the bottom part of the bottom sheet, so I need to add some arbitrary padding to the bottom. I think there is a better solution.

Going to keep looking a bit, but let me know @shrouxm if you have any ideas here.

Fixes #251


shrouxm (approved):
re: 2, it works for me without the manual bottom padding if i remove the BottomSheetScrollView. don't know if that will cause issues on small form factors though, & don't know why it doesn't work with it

re: 1, seems to be an open problem in react native land :/ with web css it's the `vertical-align` property, but the RN equivalent is android only. even in iOS native development it seems like people need to create custom extensions for this. i think it's ok to merge this for now and we can try to figure out a workaround

PR #280, opened 2023-08-15
fix: dont flicker login screen on startup when logged in

### LINKED ISSUE #257: Login requires restarting the app
## Description
Logging into the app returns the user to the login screen. They need to restart the app in order to be able to access the home screen.

This should be easy to fix, but is obviously a high priority bug.

@paulschreiber @ltseng @CourtneyLee333 @Hanal-code FYI for when you try logging in yourselves

## Steps To Reproduce

Try to log in.

shrouxm:
## Description
The app previously would always load into the login screen, and then load the home screen if the user was already logged in. Now the app will directly load the home screen if the user is already logged in, without briefly showing the login screen.

### Checklist

### Related Issues
This fixes the bug in the description, but I am hoping it will also fix #257. I am not able to reproduce #257 on my devices so we will have to wait until we can push some new releases to find out.

### DIFF
--- a/dev-client/src/config/index.ts
+++ b/dev-client/src/config/index.ts
@@ -13,8 +13,8 @@ const apiConfig = setAPIConfig({
   terrasoAPIURL: terrasoAPIURL,
   graphQLEndpoint: terrasoAPIURL + '/graphql',
   tokenStorage: {
-    getToken: async name => {
-      const value = await MMKV.getStringAsync(name);
+    getToken: name => {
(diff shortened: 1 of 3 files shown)


david-code (approved):

shrouxm:
@david-code can you confirm if you still experience #257 with this change merged?

david-code:
Yep, will check it out.

david-code:
@shrouxm Uninstalled the app on my emulator, reinstalled and successfully logged in.

PR #285, opened 2023-08-16
fix: minor polish round 1

shrouxm:
## Description
- Adds some padding to the bottom of the site list when it goes all the way to the bottom of the screen.
- Bottom sheet now displays over top of the search bar when fully open.
- Removed annoying scale bar behind search bar.


david-code:
> Bottom sheet now displays over top of the search bar when fully open.

Thank you! That was on my to-dos

david-code (approved):

PR #286, opened 2023-08-17
feat: bump patch version on android release

shrouxm:
## Description
Bumps the patch version of the android version name when a new release is created so users can see that an update happened.


david-code:
Seems like a good idea to me. Do you want to try releasing from this branch to make sure it doesn't break anything else in the release process?

david-code (approved):

PR #288, opened 2023-08-17
feat: toggle map styles on home screen

### LINKED ISSUE #47: Allow user to switch between map styles

shrouxm:
## Description
Allows toggling the map style between streets and satellite on the home screen. This implementation has an annoying problem where the first time the button is toggled, the Mapbox library raises an error even though everything functions totally fine. So I'm tempted to merge as is and file a backlog issue to figure the error message out, since it won't be visible outside of dev builds. (I think?). Thoughts?

### Checklist

### Related Issues
Fixes #47


david-code:
Agreed, better to have the feature and try and fix the issue later.

david-code (approved):

PR #290, opened 2023-08-18
feat: Update search

### LINKED ISSUE #275: Update Search Component to match mockup
# Description

There is a basic search implementation. It needs to be updated to match what is expected in the wireframe:

> I'm expecting out of the box functionality from Mapbox, with our typography styles applied.
> When user taps a search result from the list they will be taken to that darkened pin on the map without the location card opened. They can choose to tap the pin to open the card.
> Bottom sheet will close when search result is tapped.

david-code:
## Description

This updates the search to match something a bit closer to what we want with the wireframe.

- Items are displayed in an autocomplete dropdown. Each suggestion has two lines, one for the name and one for the address.
- Clicking on a suggestion will transport you to the suggestion location and drop a temporary site pin. However, the temporary site callout will not be shown. The user can open the callout by clicking on the temporary site pin.
- I fixed a bug where you needed to click on the autocomplete item twice in order to be transported to the location.

The trickiest part of this was definitely handling the different temporary site behaviour for the search requirement (not opening the callout on navigation). I solved it by changing the temporary site state to hold a type like `{site: {lat: number, lon: number}, displayCallout: boolen}`, but it makes things a bit messy. Would be nice to come back to this at some point and develop a clearer solution.

### Checklist

### Related Issues
Fixes #275


shrouxm (approved):
LGTM! i don't think what you did with the temp site data model is too bad, i'm also going to need to lift some state up in https://github.com/techmatters/terraso-mobile-client/issues/306 so will need to change it again anyway lol :upside_down_face:

PR #318, opened 2023-08-23
feat: open callout when pressing site card's map button

### LINKED ISSUE #306: Tapping on map buttons on site cards does not open expected site popover
## Description

When using the map icon on the site cards in the sites list view to change my viewport, I expect two things:
- be able to tell which pin is selected/viewed
- open up the site popover

Instead, when I don't open the popover for some random site beforehand, and then use the map icon to zoom to the location of another site, the popover doesn't open. Also, there is not an easy way to determine which pin is selected, which wouldn't normally be an issue since I'm not currently creating sites on top of each other, but right now, the pins are enormous.

## Steps To Reproduce
No popovers open

1. Open app
2. Open sites list view with at least two sites created
3. Tap on pin icon for one site
4. Close bottom sheet to view map (which is also a bug)

One popover open
1. Open app
2. Select a site on the map and open the popover
3. Open the sites list view with at least two sites created
5. Tap on the pin icon for a site that is not the one you have selected
6. Close bottom sheet (which is also a bug)

## Expected Behavior

1. The bottom sheet sheet should close.
2. The map viewport should center on the location.
3. The map marker for that site should be displayed in the "selected" state, which is slightly larger with Elevation 6 style.
4. The site card popover should display with Elevation 18 style.
The result will be the same as if the user had tapped a Site map marker on the map itself, which can be seen in this mockup https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=582%3A36087&mode=design&t=gJYpXXYNt458Ak2R-1

## Additional context
https://github.com/techmatters/terraso-mobile-client/assets/2730009/d65f98d4-2e4a-43e2-a861-ece0c1b6b737

shrouxm:
## Description
**NOTE: THIS PR IS BASED ON https://github.com/techmatters/terraso-mobile-client/pull/290 TO AVOID MERGE CONFLICTS. MAKE SURE THAT PR GETS MERGED AND BASE BRANCH IS UPDATED TO MAIN BEFORE MERGING THIS PR.**

### Checklist

### Related Issues
Fixes #306 

### Verification steps
Clicking on map button on site cards should open site callout as well as panning map to site.

### DIFF
--- a/dev-client/src/model/map/mapSlice.ts
+++ b/dev-client/src/model/map/mapSlice.ts
@@ -1,6 +1,11 @@
 import {PayloadAction, createSlice} from '@reduxjs/toolkit';
 import {Location} from '@rnmapbox/maps';
 
+export type Coords = {
+  latitude: number;
+  longitude: number;
+};
+
 type MapState = {
   userLocation?: Location;
 };
--- a/dev-client/src/screens/HomeScreen.tsx
+++ b/dev-client/src/screens/HomeScreen.tsx
@@ -1,10 +1,10 @@
 import SiteMap from '../components/home/SiteMap';
 import {useCallback, useEffect, useRef, useState} from 'react';
 import Mapbox, {Camera, Location} from '@rnmapbox/maps';
-import {updateLocation} from '../model/map/mapSlice';
+import {Coords, updateLocation} from '../model/map/mapSlice';
 import {useDispatch} from '../model/store';
 import {useSelector} from '../model/store';
-import {fetchSitesForUser} from 'terraso-client-shared/site/siteSlice';
+import {Site, fetchSitesForUser} from 'terraso-client-shared/site/siteSlice';
 import BottomSheet from '../components/home/BottomSheet';
 import {ScreenDefinition} from './AppScaffold';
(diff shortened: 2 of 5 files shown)


dev-client/src/model/map/mapSlice.ts
```diff
@@ -1,6 +1,11 @@
 import {PayloadAction, createSlice} from '@reduxjs/toolkit';
 import {Location} from '@rnmapbox/maps';
 
+export type Coords = {
+  latitude: number;
+  longitude: number;
+};
+
```
david-code, line 8: comment: Yay :)

dev-client/src/screens/HomeScreen.tsx
```diff
@@ -1,35 +1,38 @@
...
+  | {
+      kind: 'site';
+      siteId: string;
+    }
+  | {
+      kind: 'location';
+      coords: Coords;
+      showCallout: boolean;
+    }
+  | {kind: 'none'};
```
david-code, line 27: comment: Yes, good idea.

david-code (approved):
Looks good!

PR #330, opened 2023-08-23
test: attempt to enable version bumps on main

shrouxm:
## Description
Trying to make it so release action will bump the app version on main branch.


shrouxm:
this is out of date, closing to keep things tidy

PR #332, opened 2023-08-23
fix: Navigate to site dashboard after site creation

david-code:
## Description

Like the title says, this makes it so that the user is navigated to the site dashboard instead of the previous screen. It also makes it so that the user is directed to the home screen after exiting the site dashboard.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/dev-client/src/components/sites/CreateSiteView.tsx
+++ b/dev-client/src/components/sites/CreateSiteView.tsx
@@ -84,7 +84,7 @@ export default function CreateSiteView({
 
   const [errors, setErrors] = useState<Error>({});
 
-  const {reset} = useNavigation();
+  const navigation = useNavigation();
 
   /**
    * Checks the form status with the yup library, and posts to backend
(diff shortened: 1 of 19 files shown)


dev-client/src/components/sites/CreateSiteView.tsx
```diff
@@ -107,13 +110,21 @@ export default function CreateSiteView({
...
+      longitude,
+      projectId,
+    });
+    if (createdSite !== undefined) {
+      reset({
+        routes: [
+          {name: 'HOME'},
+          {name: 'LOCATION_DASHBOARD', params: {siteId: createdSite.id}},
+        ],
+      });
```
shrouxm, line 125: question: i didn't know the reset function existed! when i was doing a similar change to navigate to the project page after creating a project, i used the `replace` function for similar reasons. do you know what the exact expected behavior is for navigation? e.g. if you create a site from a project's sites view, you'll now be taken to the site's dashboard. but if you exist the dashboard, should you go back to the project's sites view or back to the home screen?

david-code: Yep, using `replace` seems like a better solution here for sure!

PR #385, opened 2023-08-30
fix: clean up localized text

### LINKED ISSUE #341: Ensure all text is localized
## Description
Ensure all text is localized. Some English strings are hard-coded:
```
            gps: {text: 'Use my current location (GPS)'},
            pin: {
              text: 'Use map pin',
              isDisabled: sitePin === undefined,
            },
            coords: {text: 'Enter coordinates'},
```

paulschreiber:
## Description
Clean up localized text.
* don't use all-caps in strings file
* add site creation strings
* use constants for temporary text

### Related Issues
Fixes #341.

### DIFF
--- a/dev-client/src/translations/en.json
+++ b/dev-client/src/translations/en.json
@@ -32,6 +32,14 @@
         "body": "in the side menu to the left or at <landpks href=\"https://landpotential.org\">landpotential.org</landpks>"
       }
     },
+    "create": {
+        "title": "Create Site",
+        "gps": "Use my current location (GPS)",
+        "pin": "Use map pin",
+        "coords": "Enter coordinates",
+        "location_label": "Site location",
+        "name_placeholder": "Site name"
+    },
     "dashboard": {
       "privacy": "Data Privacy",
       "team_button": "Team",
--- a/dev-client/src/translations/en.json
+++ b/dev-client/src/translations/en.json
@@ -29,7 +29,7 @@
       },
       "learn_more": {
         "lead": "Find more information and guides",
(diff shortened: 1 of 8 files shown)


david-code (approved):

dev-client/src/translations/en.json
```diff
@@ -7,12 +7,11 @@
...
-    "more_info": "MORE INFO",
-    "soil_id_prediction": "SOIL ID PREDICTION",
-    "ecological_site_prediction": "ECOLOGICAL SITE PREDICTION",
-    "annual_precip_avg": "ANNUAL PRECIPITATION AVG",
-    "elevation": "ELEVATION",
+    "more_info": "More info",
+    "soil_id_prediction": "Soil ID prediction",
+    "ecological_site_prediction": "Ecological site prediction",
+    "annual_precip_avg": "Annual precipitation avg",
+    "elevation": "Elevation",
```
shrouxm, line 14: question: is it considered best practice to change the case of i18n'd text programmatically? 

i figured it was best to make the strings uppercase in here since it might vary by language whether it makes sense for them to be upper cased. also if the string is used in multiple places the programmer must remember to uppercase it each time

but i also see the advantages of doing it programmatically. i've never worked on a multilingual UI before

shrouxm (approved):
LGTM

dev-client/src/translations/en.json
paulschreiber: - If you put the text in all-caps in the JSON file, you lose bits of information. By using CSS or `.toUpperCase()`, we preserve the extra data. 
- Keeping the text in mixed case allows us to easy change it systemwide at a later point.

PR #386, opened 2023-08-30
fix: Clear search box when search query empty

david-code:
## Description

Uses the Web API [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/AbortController) to cancel requests that are still pending when a previous request is triggered. Previously this would often happen with short strings, where the search request for a shorter string takes longer than a longer string.

- Request 1 triggered.
- Request 2 triggered.
- Request 2 completes.
- Request 1 completes.

The results of Request 1 would overwrite Request 2.

This was particularly a problem when trying to reset the search results, because setting the react state takes much less time than making a HTTP request. So it would be something like:

- Request triggered for query of length 2
- Suggestions cleared when user continues deleting
- Request completes, and suggestions set with this result.

Which is why this needed to completed to fix this bug.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/dev-client/src/components/home/MapSearch.tsx
+++ b/dev-client/src/components/home/MapSearch.tsx
@@ -42,9 +42,7 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
   const {t} = useTranslation();
   const [query, setQuery] = useState('');
   const [suggestions, setSuggestions] = useState<Suggestion[]>([]);
-  const [_abortController, setAbortController] =
-    useState<AbortController | null>(null);
-  //const [isLoading, setIsLoading] = useState(false);
+  const [, setAbortController] = useState<AbortController | null>(null);
   const [hideResults, setHideResults] = useState(false);
 
   async function makeSuggestionsApiCall(queryText: string) {
--- a/dev-client/src/components/home/MapSearch.tsx
+++ b/dev-client/src/components/home/MapSearch.tsx
@@ -42,14 +42,46 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
   const {t} = useTranslation();
   const [query, setQuery] = useState('');
   const [suggestions, setSuggestions] = useState<Suggestion[]>([]);
+  const [_abortController, setAbortController] =
+    useState<AbortController | null>(null);
   //const [isLoading, setIsLoading] = useState(false);
   const [hideResults, setHideResults] = useState(false);
 
-  async function querySuggestions() {
-    if (query.length > 2 || suggestions.length !== 0) {
-      const {suggestions: newSuggestions} = await getSuggestions(query);
+  async function makeSuggestionsApiCall(queryText: string) {
+    const newAbortController = new AbortController();
+    setAbortController(current => {
+      if (current !== null) {
+        current.abort();
+      }
+      return newAbortController;
+    });
+    const newSuggestions = await getSuggestions(
+      queryText,
+      newAbortController.signal,
+    );
+    setAbortController(null);
+    return newSuggestions;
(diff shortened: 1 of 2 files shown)


dev-client/src/components/home/MapSearch.tsx
```diff
@@ -42,14 +42,46 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
...
 
-  async function querySuggestions() {
-    if (query.length > 2 || suggestions.length !== 0) {
-      const {suggestions: newSuggestions} = await getSuggestions(query);
+  async function makeSuggestionsApiCall(queryText: string) {
+    const newAbortController = new AbortController();
+    setAbortController(current => {
+      if (current !== null) {
+        current.abort();
+      }
```
shrouxm, line 55: nit:
```suggestion
      current?.abort();
```

dev-client/src/components/home/MapSearch.tsx
```diff
@@ -42,14 +42,46 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
...
+      return newAbortController;
+    });
+    const newSuggestions = await getSuggestions(
+      queryText,
+      newAbortController.signal,
+    );
+    setAbortController(null);
+    return newSuggestions;
```
shrouxm, line 63: comment: this took me a second to wrap my head around, i thought there might be a bug where when a request gets cancelled, it sets the next request's abort controller to `null`, because i always forget if you `await` a request that fails the method throws. i think this retry logic should be somehow abstracted and encapsulated somewhere else rather than living in a component's implementation, but that seems tricky, could you file a follow-up tech debt issue for that?

dev-client/src/components/home/MapSearch.tsx
```diff
@@ -42,14 +42,46 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
   const {t} = useTranslation();
   const [query, setQuery] = useState('');
   const [suggestions, setSuggestions] = useState<Suggestion[]>([]);
+  const [_abortController, setAbortController] =
```
shrouxm, line 45: nit: if the variable is entirely unused, i think you can avoid naming it at all like this:
```suggestion
  const [, setAbortController] =
```
i've never actually done this, i just have gotten frustrated in the past about how using `_` in JS actually binds a variable named `_`, but i looked it up and apparently that is a thing: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#array_destructuring

shrouxm (approved):
looks good! couple optional nits

dev-client/src/components/home/MapSearch.tsx
```diff
@@ -42,14 +42,46 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
...
+    useState<AbortController | null>(null);
   //const [isLoading, setIsLoading] = useState(false);
```
paulschreiber, line 47:
```suggestion
```

dev-client/src/components/home/MapSearch.tsx
david-code: Yeah, will do.

dev-client/src/components/home/MapSearch.tsx
david-code: nice!

dev-client/src/components/home/MapSearch.tsx
david-code: https://github.com/techmatters/terraso-mobile-client/issues/388

Issue #367, opened 2023-08-29
Unable to select correct site when two pins are moderately close together.


ltseng:
## Description
When two pins are moderately close together, I'm having trouble selecting the right pin. This was observed on iOS - pins on Android are currently too small, so I'm avoiding this selection issue there.

## Additional context
https://github.com/techmatters/terraso-mobile-client/assets/2730009/3a621b8e-c055-40d9-9526-663e3d75ce32

ltseng:
@shrouxm I filed this separate issue, as requested.

shrouxm:
@ltseng there is a chance #387 may have incidentally solved this, are you able to reproduce on the newest app versions? i'm currently unable to find a pin distance where taps that were clearly closer to one pin than the other still opened the farther pin

ltseng:
Not seeing it any more on Android or iOS!

ltseng:
@david-code I seem to be unable to save any sites in version 37 on iPad now

https://github.com/techmatters/terraso-mobile-client/assets/2730009/7e4a1616-a639-4854-882e-dc203644a9ab

Issue #365, opened 2023-08-29
Temporary site marker is transparent


ltseng:
## Description
The temporary site marker that is added to the map when a search result is selected from the list of search suggestions appears to be transparent. It should instead be fully opaque, with the color #1A202C

![image](https://github.com/techmatters/terraso-mobile-client/assets/2730009/0de35d42-a4c4-457c-9c58-cf992a995c65)

shrouxm:
@CourtneyLee333 in the design the color for "action/active" (which is used for the temp site marker) has a 70% opacity, can i assume based on the existence of this bug report that that's a mistake in the design file?


![Image](https://github.com/techmatters/terraso-mobile-client/assets/3411715/2c2d9b09-41b1-4735-84bf-194e3a6925fb)

ltseng:
I see it fixed on Android, but not yet on iOS

![image](https://github.com/techmatters/terraso-mobile-client/assets/2730009/bc78de76-1de0-4022-aeff-a3f4032368d7)

shrouxm:
i see this also on my latest iOS build. no idea why this would differ by platform, but i will re-assign to paul to look into why it's still not working on iOS specifically

shrouxm:
i see it's marked as low priority, i was only touching it because i happened to be working on the icons anyway for other higher priority issues, so no rush @paulschreiber

paulschreiber:
@ltseng Please check again against build 40 or later. Seems okay now.

ltseng:
Does seem fixed now!

Issue #316, opened 2023-08-22
Map pin icons are different sizes on iOS and Android; blurry on iOS


ltseng:
## Description
The pins are enormous on iOS, but not on Android. The following screenshots have been taken on both iOS and Android at roughly the same scale, aka whatever scale the map is set to upon first opening the app.

Android

![Screenshot_20230822-163031_Terraso LandPKS.jpg](https://github.com/techmatters/terraso-mobile-client/assets/2730009/cf21096f-fc1e-46cc-ba01-23bec19b2402)

iOS

![image](https://github.com/techmatters/terraso-mobile-client/assets/2730009/0c71e9d8-b25e-4b9c-9591-37771c33867c)

shrouxm:
@paulschreiber can you look into this? i think the only relevant values should be in `dev-client/src/components/home/SiteMap.tsx`. we specify the size in these two places:
- when generating the static icon image here: https://github.com/techmatters/terraso-mobile-client/blob/e66af6f882b2d68b0a442459d099c680ef6e52e0/dev-client/src/components/home/SiteMap.tsx#L213-L217
- when specifying the layer styles here: https://github.com/techmatters/terraso-mobile-client/blob/e66af6f882b2d68b0a442459d099c680ef6e52e0/dev-client/src/components/home/SiteMap.tsx#L260-L265

i'm guessing one (hopefully not both 😱) of the values behaves differently per platform, so if you could find a set of values that creates the same size icon on android and iOS that should resolve the issue. See e.g. [here](https://www.figma.com/file/LnZIxeirOsPd5zsUBfJRAf/LandPKS-Design-Prototype-CURRENT-8%2F11%2F23?type=design&node-id=1062-42142&mode=dev) in the design for the expected size. (We aren't displaying the cluster icons or increasing the icon size when it is selected yet).

paulschreiber:
Filed https://github.com/rnmapbox/maps/issues/3009

paulschreiber:
Please retest on build 24 or later.

ltseng:
Map markers on Android are now tiny. On iPad and iPhone, they're a good size, but blurry.

ltseng:
This is leading to a secondary issue of me not being able to select the correct pin when two are moderately close together

https://github.com/techmatters/terraso-mobile-client/assets/2730009/3a621b8e-c055-40d9-9526-663e3d75ce32

shrouxm:
ok soooooo i have tried:
[...]

Issue #192, opened 2023-07-13
bug: icons on home screen have weird hitbox and positioning


shrouxm:
## Description
Icons on the home screen have hitboxes that don't match their visible boundaries, and shift slightly as the map is moved in and out. I suspect the second issue is because the icon component has a padding around the actual icon, and so the bottom of the icon is directly anchored to the coords. The hitboxes we may need to manually specify.

ltseng:
Is this related to #367?

shrouxm:
this was unrelated. this issue is icons reacting to a tap when they have not been tapped, #367 is the app doing a bad job picking which icon to tap when a tap is near multiple icons. this issue seems to have gotten resolved unintentionally in the many iterations of our map icon rendering code so i'm gonna close it!

PR #387, opened 2023-08-31
fix: map icon size and resolution

shrouxm:
## Description
I think what is happening here is we're working around a bug in rnmapbox where they ignore the `scale` attribute of RN `ImageSource`s passed into the `Mapbox.Images` component on Android only. We do this by selectively multiplying the icon size by the pixel ratio on Android only. I'd like paul to confirm it looks good on his iOS dev setup before merging, but i think we still won't know until we push a release with this fix.

### Checklist

### Related Issues
Fixes (hopefully, once QA'd) #192, #367, #316, and #365

### Verification steps
Map icons should be the correct size and not pixelated on all platforms.


shrouxm:
@paulschreiber there are two commits on the branch. the head commit i think should be working properly, but if it isn't try the other commit and make that one the head commit if it works better!

paulschreiber (approved):

Issue #310, opened 2023-08-22
Styling changes for Create Site form


ltseng:
Using this issue to call out differences between what is observed in the app vs. what is present in the Figma designs

- [ ] Form does not show GPS accuracy (Android and ipad). Note the placement of the accuracy field has been updated since the last Figma release. 
- [x] Save button elevation seems too high. Refer to design. 
- [ ] No error message displays when "Enter coordinates" is selected and invalid lat/long coordinates are entered, and the user clicks save (Android and ipad). It should read "Invalid coordinates", but I'm not sure if more detail is needed. Note the design update in the Figma working file that displays helper text for the consolidated lat/long field. 
- [ ] Save button displays on top of bottom sheet project selection (ipad).
- [ ] Error message should be "Site name must be at least 3 characters", both when there is no name entered and when the name is less than 3. Currently, different error messages display for each case, but we should use the same error message.
- [ ] (When adding a site directly to a project from the project dashboard screen) I don't receive the error message for the Site name field when taking this path unless I've entered valid coords (Android and iOS).
- [ ] Let’s get rid of the pencil icons on lat and long fields. 
- [ ] Add counter that shows current character length of site title, and max length of site title

ltseng:
@CourtneyLee333 can you provide screenshots for some of these?

CourtneyLee333:
![Screenshot 2023-08-22 at 5 12 38 PM](https://github.com/techmatters/terraso-mobile-client/assets/8915145/296a791a-e247-4d20-877f-e71dbf9cefe3)
Location accuracy and lat/long. Note that this screenshot is my working file and not included in the current engineer version. Location accuracy was a Jeancarlo task, and hasn't been implemented in the app yet, so might as well put it in the new location.

CourtneyLee333:
![Screenshot_20230822-171621(1)](https://github.com/techmatters/terraso-mobile-client/assets/8915145/808aaa58-6504-409a-a550-43a7a7bd8d28)
Elevation, or drop shadow, on the Save button is not correct. You can find the styles in the LandPKS MUI doc. Engineers, let me know if you need direction on that.

CourtneyLee333:
@ltseng This one can be removed as it will be addressed by another fix:
 Latitude and Longitude input labels should have a different style than the headers "Site Location", "Add to Project" and "Data Privacy". They should be not bold and smaller font size.

CourtneyLee333:
Save button should be underneath bottom sheet:
[...]

Issue #308, opened 2023-08-22
Styling for Create Projects screen


ltseng:
- [ ] Save button is not in the correct location and the elevation is too high (Android and ipad). 
- [ ] Error message should read "Project name must be at least 3 characters", both when there is no name entered and when the name is less than 3.

ltseng:
@CourtneyLee333 at your convenience, can you attach a screenshot or two illustrating what you mean? Unless a screenrecording would be better.

CourtneyLee333:
![Screenshot_20230822-164929](https://github.com/techmatters/terraso-mobile-client/assets/8915145/b466bfc3-58b4-4238-8330-2367b512b2ba)
Save button should be FAB anchored to bottom right of screen

CourtneyLee333:
![Screenshot_20230822-164957](https://github.com/techmatters/terraso-mobile-client/assets/8915145/057f445d-d50b-4646-b353-2cdd229a49aa)
When user taps save before entering a project name, the error message should be consistent with entering too short of a name. "Project name must be at least 3 characters"

CourtneyLee333:
![Screenshot_20230822-165015](https://github.com/techmatters/terraso-mobile-client/assets/8915145/22c58b3b-dd56-4dbd-9df8-49841fd47949)
![Screenshot_20230822-165049](https://github.com/techmatters/terraso-mobile-client/assets/8915145/05c919bc-1383-429f-8b95-f50553c4166a)
Text input field labels should remain visible. Shown in the second image is a common, expected way of handling that.

shrouxm:
it seems that keeping the input's name visible does not come out of the box in our component library, so it will be a much higher lift to implement. thankfully when we do implement it it should automatically apply to all of our inputs at once (unless we only want this feature for some of our inputs?? @CourtneyLee333). i will file a separate issue to track that feature

CourtneyLee333:
@shrouxm If that's a much higher lift I'm open to an alternative, such as a "header" type thing on top of the field. This will come into play a lot in the data inputs, so it's important to figure out holistically.

ltseng:
@CourtneyLee333 can you help me QA this? I don't have your same eye for detail, or familiarity with the designs.

CourtneyLee333:
@ltseng Looks good!

ltseng:
Awesome, closing this one!

PR #389, opened 2023-08-31
fix: project form style issues

shrouxm:
## Description
Fixes some style issues on the project form

### Checklist

### Related Issues
Fixes (once QA'd) #308. Also resolves one item in #310 

### Verification steps
Project form on both project creation and edit views should have FAB in the correct location and elevation, and should always display a min length error message instead of a field required error message even when the name input is empty.


david-code (approved):
This looks good to me. With respect to lifting the `Formik` component up out of `Form`, did you just do that to make manipulating the UI easier for different components, or was there some deeper motivation.

shrouxm:
@david-code i lifted the state because i needed to put the FAB higher up the component tree, so the state needed to get lifted above the new placement of the FAB

Issue #378, opened 2023-08-30
Fix text for empty sites list


ltseng:
## Description
The wrong text is displaying over the empty sites list

## Steps To Reproduce
1. Open the app and sign in with an account that has no sites created
2. Open the bottom sheet sites list.

## Expected behavior

When sites list is empty, the text there should start with "You have not created any sites"...

![image](https://user-images.githubusercontent.com/8915145/264363683-56b77a38-072a-4823-9405-15299983ef54.png)

Then, when a user taps the "i" or info icon on the top right corner of the app bar, the "Soil Science in your hands" chunk of text should display as a popover.

![image](https://user-images.githubusercontent.com/8915145/264364232-da3f5033-137d-41a3-84e3-2de84cab4bf0.png)

## Actual behavior
When the sites list is empty, the "Soil Science in your hands" chunk of text is displaying 
![image](https://user-images.githubusercontent.com/2730009/264157325-122b7bfd-43a0-49d2-84ab-e24ca8ded34f.jpeg)

shrouxm:
should be fixed in version 41 and above!

PR #393, opened 2023-09-01
fix: home screen info bottom sheets

shrouxm:
Fixes (once QA'd) #378, refactors ScreenScaffold to include more flexible AppBar implementation, and uses that implementation for other screens for consistency. Please check out the PR and make sure all the app bars still work as expected before merging, this PR touches all the screens!

### Checklist

### Verification steps
The info button on the home screen should now show the "Soil Science In Your Hands" message, while the info message that appears in the site list bottom sheet when it is empty should match the spec.


david-code:
@shrouxm had to revert this because of a bug on iOS. Part of the App Bar is hidden by the dynamic island, and this cuts of the title text for most screens.

PR #407, opened 2023-09-05
feat: screens refactor with fix for home screen bottom sheet

shrouxm:
## Description
follow-up attempt to #393, tries to fix the iOS bug with a `SafeAreaView`. @paulschreiber please verify if the iOS bug still present before merging!

Fixes (once QA'd) #378, refactors ScreenScaffold to include more flexible AppBar implementation, and uses that implementation for other screens for consistency. Please check out the PR and make sure all the app bars still work as expected before merging, this PR touches all the screens!

### Checklist

### Verification steps
The info button on the home screen should now show the "Soil Science In Your Hands" message, while the info message that appears in the site list bottom sheet when it is empty should match the spec.


paulschreiber:
This is better. There are a few bugs:
* top well should have the same green background as the reset of the screen
* bottom well is unnecessary — we can move the toolbar flush with the bottom of the screen
* the LandPKS title is no longer centered
image

paulschreiber (changes requested):
This is better. There are a few bugs:
* top well should have the same green background as the reset of the screen
* bottom well is unnecessary — we can move the toolbar flush with the bottom of the screen
* the LandPKS title is no longer centered

shrouxm:
@CourtneyLee333 could you chime in on this? some context:
- i am refactoring the way the header and footer get rendered for technical reasons unrelated to their rendering, but the refactor is giving us more control over the rendering and a different default rendering
- previously, on devices with inset elements (like the iPhone sensor bar) the header background would extend behind (as in z-index) those inset elements, and the actual header elements (title, buttons) would sit beneath (as in y-position) them
- in the new version, the background of the inset elements is the device default
- considerations:
  - in the old version, the color contrast between the system text (clock, network info, etc) and our background color was not great
  - in the new version, we lose a bit of screen real estate on the bottom but there's a bit more breathing room in the bottom bar, and the look and feel could be seen as a bit less unified
  - in the new version, the title is left aligned instead of centered which is (now) uniform with android but a bit less iOS-y. did we want to have iOS-y and android-y flavors of the app?

CourtneyLee333:
I agree with Paul's first two statements above, and there should be a way to set the system info at the top to white so the contrast is sufficient. 

Whether the screen title is centered or left aligned, or whether it is consistent across platforms, doesn't matter to me.

shrouxm:
ok i think i have everything working! @paulschreiber can you confirm looks good on iOS?

[bottom-sheets.webm](https://github.com/techmatters/terraso-mobile-client/assets/3411715/ef6ff6b5-1f7b-4b58-82ed-b66e31810aa7)

paulschreiber:
Looks great:
image

paulschreiber (approved):

PR #428, opened 2023-09-14
fix: update lat/lon of create site form as user's location changes

### LINKED ISSUE #256: Add GPS refresh to Create Site form
The Create Site form needs a way to refresh the GPS coords when the user moves around.

shrouxm:
## Description
Moves the logic that updates user's location to the root of the app, and modifies the create site form to keep the lat/lon fields updated as the user moves.

### Checklist

### Related Issues
Fixes #256

### Verification steps
Latitude/longitude field of create site form should update as user moves around.

### DIFF
--- a/dev-client/src/components/sites/CreateSiteView.tsx
+++ b/dev-client/src/components/sites/CreateSiteView.tsx
@@ -1,7 +1,6 @@
-import {Location} from '@rnmapbox/maps';
 import RadioBlock from '../common/RadioBlock';
 import {Fab, FormControl, Input, ScrollView, Text, VStack} from 'native-base';
-import {useCallback, useMemo, useState} from 'react';
+import {useCallback, useMemo, useState, useEffect} from 'react';
 import {
   ProjectPrivacy,
   SiteAddMutationInput,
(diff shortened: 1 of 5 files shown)


dev-client/src/components/sites/CreateSiteView.tsx
```diff
@@ -55,25 +53,24 @@ type LocationInputOptions = 'coords' | 'gps' | 'pin';
 
 export default function CreateSiteView({
   defaultProject,
-  userLocation,
   createSiteCallback,
   sitePin,
 }: Props) {
   const {t} = useTranslation();
 
+  const userLocation = useSelector(state => state.map.userLocation);
```
david-code, line 61: comment: Mostly just curious, is there a reason why you moved the selector down to this component? I prefer to pass the location down as a prop, and have the screen be the one pulling in the global state with a selector.

david-code (approved):
Looks good!

dev-client/src/components/sites/CreateSiteView.tsx
shrouxm: i usually default to pushing hook usages as far down as possible if they can trigger a re-render and to avoid repetitive parameter passing, but it's one of those weird squishy intuitive things i think

PR #465, opened 2023-10-02
feat: Add free form text input

david-code:
This input is intended for the add user to project screen. It will just encapsulate the logic of validating the user emails exists and displaying an error message if invalid. In the future the error message may be more complex, like displaying a modal or something, so I've decided to make this a whole separate component.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/dev-client/src/components/common/FreeformTextInput.tsx
+++ b/dev-client/src/components/common/FreeformTextInput.tsx
@@ -1,0 +1,43 @@
+import {FormControl, Input} from 'native-base';
+import {useCallback, useState} from 'react';
+
+type Props = {
+  validationFunc: (input: string) => Promise<null | string>;
+  helperText?: string;
+  placeholder?: string;
+};
+
+/**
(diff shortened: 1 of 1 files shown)


dev-client/src/components/common/FreeformTextInput.tsx
```diff
@@ -0,0 +1,38 @@
...
+      setHasError(validationResults);
+    } else {
+      setTextValue('');
+      setHasError(null);
+    }
+  }, [validationFunc, textValue, setTextValue]);
+
+  return (
+    <FormControl isInvalid={hasError !== null}>
+      <Input
```
shrouxm, line 30: nit: the linter complains about "Empty components are self-closing" here. i guess we're missing a settings somewhere to make the CI lint check always error on warnings? we should find/update that CI setting

shrouxm (approved):

PR #477, opened 2023-10-03
fix: Don't submit locationSource on create site

david-code:
This allows me to create a site locally, and logically makes sense. However I'm not sure why Typescript does not pick up on locationSource being included in the input to the GraphQL call here. @shrouxm any idea?

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/dev-client/src/components/sites/CreateSiteView.tsx
+++ b/dev-client/src/components/sites/CreateSiteView.tsx
@@ -190,10 +190,8 @@ export const CreateSiteView = ({
   const navigation = useNavigation();
 
   const onSave = useCallback(
-    async (form: FormState) => {
-      let input: Partial<FormState> = {...form};
-      delete input.locationSource;
-      const {coords, ...site} = validationSchema.cast(input);
+    async ({locationSource: _, ...form}: FormState) => {
+      const {coords, ...site} = validationSchema.cast(form);
       const createdSite = await createSiteCallback({
         ...site,
         ...parseCoords(coords),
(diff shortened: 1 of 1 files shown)


shrouxm:
it's a [longstanding TypeScript deficiency](https://github.com/Microsoft/TypeScript/issues/12936). at some point i tried to find a way to format our mutations to catch this kind of bug: https://github.com/techmatters/terraso-client-shared/pull/45, but the fix i tried there wouldn't actually help with this case anyway :tired_face:

dev-client/src/components/sites/CreateSiteView.tsx
```diff
@@ -191,7 +191,9 @@ export const CreateSiteView = ({
 
   const onSave = useCallback(
     async (form: FormState) => {
```
shrouxm, line 193: nit: this would be a bit tidier, but all good either way!
```suggestion
    async ({locationSource: _, ...form}: FormState) => {
```

shrouxm (approved):

david-code:
> it's a [longstanding TypeScript deficiency](https://github.com/Microsoft/TypeScript/issues/12936). at some point i tried to find a way to format our mutations to catch this kind of bug: [techmatters/terraso-client-shared#45](https://github.com/techmatters/terraso-client-shared/pull/45), but the fix i tried there wouldn't actually help with this case anyway 😫

Thanks for the link, that's interesting!

PR #479, opened 2023-10-06
feat: Add hook for when keyboard is open

david-code:
## Description

This hook just returns a reactive value that is true when the keyboard is open, and false when it is closed. I'm using it to hide the submit button on a form when the keyboard is open, because I kept on accidentally submitting while testing - maybe it will be useful elsewhere.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/dev-client/src/hooks/index.tsx
+++ b/dev-client/src/hooks/index.tsx
@@ -1,0 +1,22 @@
+import {useEffect, useState} from 'react';
+import {Keyboard} from 'react-native';
+
+export const useKeyboardOpen = () => {
+  const [keyboardOpen, setKeyboardOpen] = useState<boolean>(
+    Keyboard.isVisible(),
+  );
+  useEffect(() => {
+    const openListener = Keyboard.addListener('keyboardDidShow', () =>
+      setKeyboardOpen(true),
+    );
+    const closeListener = Keyboard.addListener('keyboardDidHide', () =>
+      setKeyboardOpen(false),
+    );
+    return () => {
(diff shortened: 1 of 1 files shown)


dev-client/src/hooks/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+      setKeyboardOpen(true),
+    );
+    const closeListener = Keyboard.addListener('keyboardDidHide', () =>
+      setKeyboardOpen(false),
+    );
+    return () => {
+      if (openListener) openListener.remove();
+      if (closeListener) closeListener.remove();
+    };
+  });
```
shrouxm, line 19: nit: i think without the empty dependency list we will be adding and removing the listeners on each render
```suggestion
  }, []);
```

dev-client/src/hooks/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  useEffect(() => {
+    const openListener = Keyboard.addListener('keyboardDidShow', () =>
+      setKeyboardOpen(true),
+    );
+    const closeListener = Keyboard.addListener('keyboardDidHide', () =>
+      setKeyboardOpen(false),
+    );
+    return () => {
+      if (openListener) openListener.remove();
+      if (closeListener) closeListener.remove();
```
shrouxm, line 17: nit: is `if (...listener)` check necessary? there doesn't seem to be one in the docs

shrouxm (approved):

PR #482, opened 2023-10-06
fix: fix crash on launch on iOS

paulschreiber:
## Description
* Fix crash on launch on iOS by moving <Provider> above <NativeBaseProvider>
* Update Podfile to build with Xcode 15
* bump version number to 44


david-code (approved):
Nice! It kind of makes sense that this could cause problems.

PR #494, opened 2023-10-12
fix: project user list updating

david-code:
## Description

Previously this screen just accepted a list of users as a route param and removed the user when they clicked on the remove button. I've updated this to instead read the list of users/members from the redux store, which makes more sense.

Relies on https://github.com/techmatters/terraso-client-shared/pull/122

Also includes some styling fixes.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


dev-client/src/util.ts
```diff
@@ -0,0 +1,9 @@
+export function formatNames(firstName: string, lastName?: string) {
+  let formatted = '';
+  if (lastName !== undefined) {
+    formatted += lastName;
+    formatted += ', ';
+  }
+  formatted += firstName;
+  return formatted;
+}
```
paulschreiber, line 9: We have similar code in AccountAvatar.js:

    const name = `${user.firstName} ${user.lastName}`;

And AppBar.js

    {user.firstName} {user.lastName}

Should we consolidate name formatting code?

david-code: I think this format is slightly different, here the order is lastName, firstName. But it would make sense to group the other two.

paulschreiber: @david-code I would write this as 
```js
[lastName, firstName].filter(Boolean).join(", ")
```

david-code: Oh yeah, that's nice. If you want to make a PR with that change, I will approve it.

Issue #492, opened 2023-10-11
 Tapping a map marker should center the marker and the popover


paulschreiber:
## Summary
On the home screen,  tapping a map marker should center the marker and the popover.

Currently, the items may appear partially offscreen:
![image](https://github.com/techmatters/terraso-mobile-client/assets/86784/1530170f-900c-4f01-9444-ec35bccf2c2a)

ltseng:
Should be available in version 48

ltseng:
This works well!

Issue #445, opened 2023-09-22
Keep location popover open until dismissed


ltseng:
In a usability study, a user opened the location popover, selected "More Info" to view the location dashboard, and then went back to the map, only to discover that their pin had disappeared, along with the popover. They had to search for that same location again and drop another pin. I think it's reasonable for users to expect that a popover stay around until dismissed when the user temporarily moves away to view detailed information.

Current behavior:
1. User searches for a location and selects a search result, or user selects a location on the map
2. Pin is dropped and location popover is opened.
3. User clicks "More Info" to open up the location dashboard, and then hits the "x" button in the top left to close the dashboard.
4. User sees that the pin has disappeared and the location popover is not open.

Desired behavior.
4. User sees that the pin is still there and the location popover is still open. They have to physically press the close button ("x" in the top right corner) or tap outside of the popover to close it.

Note that if a user taps a site and then clicks on the site's name to open the Site Dashboard, the pin is still dropped and the location popover is still open when the user closes the site dashboard.

ltseng:
Looks good on iPhone and Android!

Issue #444, opened 2023-09-22
Short tap instead of long press to bring up location popover


ltseng:
Coming from the usability study - people were not familiar with the concept of long-pressing on a map to drop a pin or bring up additional information. We'd like to change the current interaction model:

Current:
Long press on map to drop a pin and open the location popover

Desired:
Short tap on map to drop a pin and open the location popover.

ltseng:
This is working well and as expected on all my devices. I'm seeing a bit more evidence of the behavior in #120, but on the home screen map sometimes. I'll upload a recording there.

ltseng:
@CourtneyLee333 this is available to test if you want to see it in action!

CourtneyLee333:
Yay, works great! I'd like to test it by searching and tapping on a location where I already have site, or have some very nearby, but the app seems to have logged me out and I can't log back in.

CourtneyLee333:
Ok, what I'm seeing now is that when I tap my GPS location icon I cannot access the popover. Also, the icon is still moving around a lot, so once I tried to tap it but it moved at that moment so I created a pin at what I thought was my location but the location icon had moved. These issues exist on my iPad and Android phone. 

https://github.com/techmatters/terraso-mobile-client/assets/8915145/0b30d82d-ccc2-4c96-a14e-5746b8afb840

ltseng:
Okay, we haven't addressed the issue of the user's location icon moving around too often. That is logged in #464. 

@ajabeckett do you want to use this issue to fix the bug that Courtney brought up, i.e. that of the GPS icon functionality interfering with the popover?

ajabeckett:
Yes, I'm working on a fix for the user location interfering with taps now.

ltseng:
Fixed!

Issue #436, opened 2023-09-18
Styling changes for the map


ltseng:
- [ ] Map popover
  - [ ] Needs drop shadow
- [ ] Map marker icon
  - [ ] Needs style when selected ([figma link](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=582%3A36087&mode=design&t=hfeNdwjZ8MPjsGCy-1))
- [ ] Search bar needs "x" option to clear the search text in the input field.

ltseng:
Partially done as of 10/17, still need drop shadow and search bar "x"

ltseng:
The search bar x looks good and functions as expected. The drop shadow shows up on iOS for the map popover, but not on Android though?

ltseng:
Android screenshot ![Screenshot_20231026-112051_Terraso LandPKS.jpg](https://github.com/techmatters/terraso-mobile-client/assets/2730009/c86f1e4d-d559-4f10-8a9c-d4de050ee198)

ltseng:
iOS screenshot![image](https://github.com/techmatters/terraso-mobile-client/assets/2730009/843cb434-5c9d-452e-ab53-7e7e1ad4dab6)

ajabeckett:
I'll take a look at this once I fix an issue with the Android emulator on my machine.

ltseng:
Reminder that this one is still pending :)

PR #495, opened 2023-10-12
feat: location popover updates

ajabeckett:
## Description
Several updates to location popovers: 
- Location popover should stay open until dismissed, not close when navigating to the More Info screen
- Switch from long press to short press to open popover
- Added a conditional triangle to the top of the popover cards 
- Tapping a map marker should center the marker (or vertically offset, in the case of the taller TemporarySiteCallout)

The additional camera centering made the SiteMap file a bit unwieldy, so I broke the components and utility functions out into separate files. 

### Related Issues
Refs: #436
Refs: #444
Refs: #445
Refs: #492


dev-client/src/components/common/Card.tsx
```diff
@@ -30,18 +30,48 @@ export const CardCloseButton = (
...
+      height={0}
+      borderBottomWidth={15}
+      borderBottomColor="white"
+      borderLeftWidth={15}
+      borderLeftColor="transparent"
+      borderRightWidth={15}
+      borderRightColor="transparent"
+      alignSelf="center"
+      position="absolute"
+      top={-14}
```
shrouxm, line 46: comment: i still feel a bit affronted that this is one of the most reliable ways to make a triangle in 2023

dev-client/src/components/common/Card.tsx
```diff
@@ -30,18 +30,48 @@ export const CardCloseButton = (
   );
 };
 
+export const CardTriangle = () => {
+  return (
+    <Box
+      width={0}
+      height={0}
+      borderBottomWidth={15}
```
shrouxm, line 38: nit: factor the repeated use of "15" into a constant

dev-client/src/components/home/repositionCamera.ts
```diff
@@ -0,0 +1,31 @@
+import {CameraRef} from '@rnmapbox/maps/lib/typescript/components/Camera';
+
+export const repositionCamera = (
+  feature: GeoJSON.Feature,
+  zoomLevel: number,
+  animationDuration: number,
+  paddingBottom: number,
+  cameraRef?: React.RefObject<CameraRef>,
+) => {
```
shrouxm, line 9: nit: this could probably benefit from named parameters at this point, which would also make it easier to provide defaults for e.g. paddingBottom!

dev-client/src/components/home/repositionCamera.ts
```diff
@@ -0,0 +1,31 @@
+import {CameraRef} from '@rnmapbox/maps/lib/typescript/components/Camera';
+
+export const repositionCamera = (
+  feature: GeoJSON.Feature,
+  zoomLevel: number,
+  animationDuration: number,
+  paddingBottom: number,
+  cameraRef?: React.RefObject<CameraRef>,
```
shrouxm, line 8: question: any reason not to make `cameraRef` a required arg?

shrouxm (approved):
looks good! no blocking comments

Issue #77, opened 2023-06-07
Screen for changing user role


david-code:

ltseng:
Ready by tomorrow

david-code:
Reopening for QA

ltseng:
I'm actually going to use the user story for QA, thanks though!

PR #510, opened 2023-10-16
feat: Dialog to remove user

david-code:
## Description

This PR introduces a confirmation dialog that can be re-used in many different parts of the UI (Like #77 for example). In order to test it out, I added it to the button to remove users.

Screenshot_1697481149

A comment: Originally I wanted to re-use the nice custom `Modal` element that was introduced earlier. However, I realized that some parts of it are not customizable for now. So I made a custom `ConfirmModal` element, which is quite similar. I want to refactor the `Modal` element to make it a bit more versatile, but I want to wait until after our demo on Wednesday, just to make sure we don't mess anything up with the large amount of code that has been added to the Site Input scaffolding. Then I plan to come back and refactor the code. (Made an issue #511 to deal with this.)

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


dev-client/src/translations/en.json
```diff
@@ -129,7 +129,13 @@
       }
     },
     "team": {
-      "add": "Add team members"
+      "add": "Add team members",
+      "leave_project_modal": {
+        "trigger": "LEAVE PROJECT",
```
paulschreiber, line 134: issue: should not be in all caps.

dev-client/src/translations/en.json
```diff
@@ -129,7 +129,13 @@
...
+        "title": "Leave Project?",
+        "body": "This will remove your access to the project, but data you have contributed will remain with the project.",
```
paulschreiber, line 136:
```suggestion
        "body": "This will remove your access to the project. Data you contributed will remain with the project.",
```

paulschreiber (changes requested):

PR #512, opened 2023-10-16
fix: use absolute paths for remaining imports

paulschreiber:
## Description
Use absolute paths for remaining imports


david-code:
@paulschreiber Added a lint rule to warn on relative imports, could you take a look at it too?

ajabeckett (approved):

PR #515, opened 2023-10-17
fix: simplify formatNames function and use in UserList

paulschreiber:
## Description
* simplify the formatNames function
* rename it to formatName
* use in UserList


david-code (approved):
Looks good!

PR #520, opened 2023-10-18
feat: Manage team member screen

david-code:
## Description

This PR implements the screen used to change project member roles and remove users from a project.

Some comments:

- Screen is accessed by clicking on the user row in the Team Project Management tab
- I updated the UserList component to restrict access to the project member roles to only project managers. This is pretty messy here. The feature will be used in lots of different places in the project managements, so I think it would be best to tackle it properly in a separate PR, where I'll remove the related code from the UserList and hopefully have a cleaner solution

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


ajabeckett (approved):
Looks great!

PR #526, opened 2023-10-20
fix: adjust keyboard settings for email address

paulschreiber:
## Description
Adjust keyboard settings for email address:
* disable capitalization
* enable autofill
* use email address-optimized keyboard

Note sure if this works with `FreeformTextInput` as-is or needs to be passed in as XYZProps.


david-code (approved):
Added a commit passing the props down to the Input components, seems to work as expected.

PR #549, opened 2023-10-27
feat: Site Transfer

### LINKED ISSUE #145: Connect site transfer to backend

david-code:
## Description

Lots of things here, it turned out to be more complicated than I expected.

- Updated CheckboxGroup. Adds a select all checkbox, which allows setting the values of all of the checkboxes in the group. Calls a callback with the checkbox group ID and item ID when the value changed.
    - Had to add a dependency for some random Checkbox component. The NativeBase one does not work. Considered using one of the NativeBase replacement frameworks, but they all need some extra config, I think this is the easiest solution.
- Loads the state of checkboxes in the `SiteTransferProjectScreen`. Basically a record of project ids -> site ids -> boolean, storing if a given checkbox has been checked.
- Keeps a list of projects separate from the list of available sites, so that we can display projects without any sites
- Uses the search box to filter the list of sites
- Handles unaffiliates sites differently from projects. This creates some overhead: I'm using Javascript Symbols to create a unique projectID for these sites without a project ID, and needs to be handled seperately in quite a few places. Should definitely be refactored to encapsulate this at a later date! But I have spent too long on this issue anyways, going to leave that for later.

Final fixups:
- [ ] Fix checkbox styling

### Checklist
- [ ] New tests added

### Related Issues
Fixes #145


ajabeckett (approved):
Great work! 🎉

PR #567, opened 2023-11-02
fix: Polyfill browser b64 functions to fix login

david-code:
## Description

- jwt-decode (used in shared-client) no longer carries a polyfill for browser b64 functions which it uses
- react native JS built-ins does not include these functions, so we need to provide implementation
- see https://github.com/facebook/hermes/issues/1178


ajabeckett (approved):
🙌

paulschreiber (approved):

Issue #517, opened 2023-10-17
Create sort options for sites in list view


ltseng:
Users should be able to sort the Sites list view by name, date last modified, and distance from my current position (if GPS is enabled)

ltseng:
@david-code please keep in mind that it should be possible to sort by these three criteria in ascending and descending order!

PR #583, opened 2023-11-10
feat: Site geospatial context

david-code:
## Description

This PR adds a context that can be used to access a list of sites sorted by distance from the user's current location.

I originally wanted to use a spatial index like [geokdbush](https://github.com/mourner/geokdbush) to provide this functionality, but unfortunately I can't find a good candidate. The library linked, which looks great, is unfortunately currently broken. Instead I've just used a simple approach of calculating the distance from the site to the current user using the Haversine formula, and sorting based on that. This has the drawback that the list needs to be recalculated every time the user's location changes. I don't think this will be a performance problem at the moment, but it certainly has the possibility to become one. In the meantime we should think of how to replace this eventually.

### Checklist
- [x] New tests added

### Related Issues
Needed for https://github.com/techmatters/terraso-mobile-client/issues/517


david-code:
superseded by #633

PR #600, opened 2023-11-16
feat: enter lat/long coordinates in map search

### LINKED ISSUE #370: Entering lat/long coordinates into search bar does not initiate valid search
## Description

Typing coordinates into the home screen search bar isn't working the way I would expect, i.e. dropping a temporary site marker right on that location. Seen on Android and iOS.

https://github.com/techmatters/terraso-mobile-client/assets/2730009/e7b3056d-0c38-46aa-950f-80a6b1904f2e

ajabeckett:
## Description
Adds coordinates check to home screen map search.

### Related Issues
Fixes #370

### Verification steps
Enter valid coordinates in the map search field and press enter on the keyboard. A temporary location popover should open at the correct coordinates.

### DIFF
--- a/dev-client/src/components/home/MapSearch.tsx
+++ b/dev-client/src/components/home/MapSearch.tsx
@@ -85,14 +85,7 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
   async function querySuggestions(queryText: string) {
     if (isValidCoordinates(queryText)) {
       const [latitude, longitude] = queryText.split(',').map(Number);
-      if (
-        latitude >= -90 &&
-        latitude <= 90 &&
-        longitude >= -180 &&
-        longitude <= 180
-      ) {
-        zoomTo && zoomTo({latitude, longitude});
-      }
+      zoomTo && zoomTo({latitude, longitude});
     } else {
       if (queryText.length >= 2) {
         try {
--- a/dev-client/src/components/home/MapSearch.tsx
+++ b/dev-client/src/components/home/MapSearch.tsx
@@ -26,6 +26,7 @@ import {
 import {Icon, IconButton} from 'terraso-mobile-client/components/common/Icons';
 import {Keyboard} from 'react-native';
 import {Coords} from 'terraso-mobile-client/model/map/mapSlice';
+import {isValidCoordinates} from 'terraso-mobile-client/util';
 
 const {getSuggestions, retrieveFeature} = initMapSearch();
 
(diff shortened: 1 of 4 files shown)


dev-client/src/components/home/MapSearch.tsx
```diff
@@ -82,24 +82,39 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
...
-      try {
-        const {suggestions: newSuggestions} =
-          await makeSuggestionsApiCall(queryText);
-        setSuggestions(newSuggestions);
-      } catch (e: any) {
-        if (e.name !== 'AbortError') {
-          throw e;
-        }
+    const coordRegex =
+      /^([-+]?[1-8]?\d(?:\.\d+)?),\s*([-+]?180(?:\.0+)?|[-+]?((1[0-7]\d)|([1-9]?\d))(?:\.\d+)?)$/;
```
paulschreiber, line 86: question: what are you trying to match here? looks like a pair of decimal numbers or an integer or a decimal number? it's hard to follow. Is there a coord regex in one of the libraries we have that we can reuse?

dev-client/src/util.ts
```diff
@@ -42,3 +42,17 @@ export const removeKeys = (a: any, b: any) => {
     }
   }
 };
+
+export const isValidCoordinates = (input: string) => {
+  const coordRegex =
+    /^([-+]?[1-8]?\d(?:\.\d+)?),\s*([-+]?180(?:\.0+)?|[-+]?((1[0-7]\d)|([1-9]?\d))(?:\.\d+)?)$/;
+
```
paulschreiber, line 49: issue: this is duplicating code in MapSearch.tsx.

dev-client/src/util.ts
```diff
@@ -42,3 +42,17 @@ export const removeKeys = (a: any, b: any) => {
...
+  if (!coordRegex.test(input)) {
+    return false;
+  }
+
+  const [latitude, longitude] = input.split(',').map(Number);
+  return (
+    latitude >= -90 && latitude <= 90 && longitude >= -180 && longitude <= 180
+  );
```
paulschreiber, line 57: issue: We have `isValidLatitude` and `isValidLongitude` in gisUtils in the web client. Move those to shared client and reuse here.

dev-client/src/components/home/MapSearch.tsx
```diff
@@ -82,24 +82,39 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
...
+
+    if (coordRegex.test(queryText)) {
```
paulschreiber, line 88: issue: use isValidCoordinates() here.

paulschreiber (changes requested):

dev-client/src/components/home/MapSearch.tsx
ajabeckett: Whoops, this was a stale file.

dev-client/src/util.ts
ajabeckett: Will do, thanks!

paulschreiber (changes requested):
Avoid duplicating work in MapSearch — logic should be in isValidCoordinates.

ajabeckett:
Whoops, yes, that was an oversight, thanks for catching!

paulschreiber (approved):

PR #632, opened 2023-11-20
fix: add confirm dialog for site deletion (from site settings screen)

paulschreiber:
## Description
Add confirm dialog for site deletion (from site settings screen).


david-code (approved):
Looks good!

PR #634, opened 2023-11-22
fix: Adjusted SafeAreaView to fit on all devices

RR3b3l0:
## Description
The current implementation of SafeAreaView was not including the bottom insets, so for example an iPhone 15 pro max the tab bar was not being displayed correctly. This PR not only adjust the bottom padding but also adjusts the sites listing initial snap point based on this values.

### DIFF
--- a/dev-client/src/components/home/BottomSheet.tsx
+++ b/dev-client/src/components/home/BottomSheet.tsx
@@ -47,6 +47,7 @@ import {SearchBar} from 'terraso-mobile-client/components/common/search/SearchBa
 import {SiteFilter} from 'terraso-mobile-client/components/sites/filter';
 import {ProjectSelect} from 'terraso-mobile-client/components/projects/ProjectSelect';
 import {USER_ROLES} from 'terraso-mobile-client/constants';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
 
 const EmptySiteMessage = () => {
   const {t} = useTranslation();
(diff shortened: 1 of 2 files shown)


dev-client/src/components/home/BottomSheet.tsx
```diff
@@ -63,6 +64,24 @@ const EmptySiteMessage = () => {
...
+  // We set the maximum bottom insets as 34 which is the currently maximum for a device
+  const weight0 = 1 - bottomInsets / 34; // Weight for 13
+  const weight34 = 1 - weight0; // Weight for 16
+
+  // Use the weights to calculate a weighted average
+  const result = weight0 * 13 + weight34 * 16;
+
+  // Round the result
+  return Math.round(result);
+};
```
david-code, line 83: question: Nice function! Two questions, just for my own knowledge:
1) Where does the value of 34 come from? I tried doing a quick search through iOS docs and couldn't find anything.
2) To make sure I'm understanding properly: if the bottom inset is smaller, we want the site list to open at a lower percentage (13% if there is no bottom inset), and a higher percentage if the bottom inset is higher (16% max). This is just because the bottom inset hides more of the bottom list if the bottom inset is higher, right?

david-code (approved):
Thanks for the PR, looks like a great change. I added a question I have for my own knowledge.

dev-client/src/components/home/BottomSheet.tsx
RR3b3l0: 1. The maximum number is not written anywhere as there can be a device that comes out later which might have an higher value. We set it as 34 because based on our experience (For the top inset is totally different as Androids differ way more)
2. If the bottom inset is near or equal 0 we set it for 13, if the value is near or equal to 34 we set it to 16, the mid values are calculated too, for example an 26 bottom inset which is normal for ios devices will point up to 15. So for an device really small like an iPhone se that has no bottom bar we set 13 as it's the value which looks better.

Feel free to change the minimum and maximum value on the result variable by changing the 13 and 16 to the desired values.

PR #644, opened 2023-11-27
feat: Allow users to clear select option

david-code:
## Description

- Creates a `None` option that users can select to clear a previously selected option
- Should be present on HomeScreen and ProjectListView filters, as well as Create Site Screen project options
- Might be some other places where this should be implemented, but that's all I found with grep. Probably easiest to let people see if they run into more issues.

### Checklist
- [x] New tests added

### Related Issues
Fixes #526


dev-client/__tests__/context/GeospatialContext.test.tsx
```diff
@@ -27,9 +27,11 @@ const TestComponent = ({mock}: Props) => {
...
   useEffect(() => {
     mock(
-      Object.entries(siteDistances || {})
-        .sort(([, distanceA], [, distanceB]) => distanceA - distanceB)
-        .map(([name]) => name),
+      siteDistances === null
+        ? null
+        : Object.entries(siteDistances)
+            .sort(([, distanceA], [, distanceB]) => distanceA - distanceB)
+            .map(([name]) => name),
```
david-code, line 34: This really should have been a different PR. It's just fixing a test I made fail in the previous PR.

Issue #502, opened 2023-10-14
feat: finish depth interval related UI work on site soil tab


shrouxm:
```[tasklist]
### remaining work after #459
- [x] Finish styling the edit interval modal
- [x] Add proper error messages for the interval modal validation
- [x] Correctly display & disable the data input toggle for data inputs that are required at the project level in the site-level depth interval edit modal
- [x] Add a delete button to the edit interval modal for site-level depth intervals with associated modal warning
- [x] Wire up the "apply to all intervals" checkbox to actually do the thing
- [x] Add modal warning when changing a depth interval with existing soil data
- [x] There are probably a few other minor style issues to address besides the edit interval modal
- [x] Make radio switches save asynchronously to correct for lag
- [x] Restrict options based on preset  selected
- [x] Restrict options based on user role
```

```[tasklist]
### remaining work after feb 28
- [x] apply to all button doesn't work
- [x] delete button shouldn't exist for preset or project intervals
- [x] disabled forms are still editable / what to do about disabled forms
- [x] styling for required data input switches
- [x] label can show up for preset depth interval
- [x] add modal warning when deleting custom depths
- [x] remove cm from depth interval table
- [x] add modal warning when changing site depth preset
- [x] Check whether new sites have depth dependent data inputs enabled automatically
```

david-code:
@ltseng For your visibility: I'm almost done with this screen, but there's a few final issues that are causing me some problems. I think it makes more sense to finish #381 , so we can see how it will work when the user selects presets at the project level. There's still some backend work that needs to be done as part of the same issue (#381) that will make finishing this issue more straightforward.

ltseng:
Thanks for the update! This makes sense. Let me know if there's anything I can do to help shed clarity or unblock you otherwise.

david-code:
@shrouxm Trying to think through project/site depth intervals. Writing this up so we can discuss next week. (Remember to also discuss deletion of site dependent data!)

A few points:

- Project depth intervals are not saved to the backend for the LandPKS or NRCS preset selections. 
    - These presets define a fixed set of depth interval bounds which can be generated on the frontend.
    - There is no need to save the required inputs to each interval. These are saved to the project soil settings.
- A given site intervals are made up of the list of project level intervals, as well as any user defined custom intervals
    - LandPKS and NRCS define intervals for the whole possible range, so that is not really an issue for them.
[...]

PR #645, opened 2023-11-28
fix: Display soil data intervals when site not in project

david-code:
## Description

There was a small bug where the soil data intervals were being skipped over when the site was not part of a project. Wanted to submit this as a separate PR to make sure it does what is expected.

### Checklist
- [ ] New tests added

### Related Issues
Related to #502


shrouxm (approved):
nice catch, ty!

PR #648, opened 2023-11-29
fix: Change Form components onChange listener

david-code:
## Description
- The Formik `onChange` listener expects an event. React native events are synthetic and often do not contain all of the properties of a browser DOM event, which Formik is expecting.
- Changed from using the `onChange` listener, to `setValue`, which should have the correct effect.
- I think this was likely working for selects and text inputs, because there seems to be some custom logic in the Formik code about if a string is passed to `onChange` . I didn't really look into it more though.
- Tested it out with Checkbox, Input and TextArea Form elements to make sure it works
- Also changes the Soil data tab to use the checkbox library instead of the NativeBase one

### Checklist
- [ ] New tests added

### Related Issues
Related to #502

### DIFF
--- a/dev-client/src/components/common/Form.tsx
+++ b/dev-client/src/components/common/Form.tsx
@@ -65,19 +65,19 @@ export const useFieldContext = <
         formikContext.setFieldValue(name, value);
       }
     },
-    [formikContext.values, name],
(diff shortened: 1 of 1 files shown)


ajabeckett (approved):
👍

PR #649, opened 2023-11-30
feat: warn on soil interval update

david-code:
## Description

Adds a modal warning that data will be deleted if the interval is updated. I want to make sure that the criteria for showing the warning here is correct.

### Checklist
- [ ] New tests added

### Related Issues
Related to #502


ajabeckett (approved):
I think this makes sense.

PR #669, opened 2023-12-06
fix: path in android/settings.gradle

shrouxm:
## Description
I had to make this change to `android/settings.gradle` to get the app to compile locally, which is required for react native version 0.72+ according to [this thread](https://github.com/facebook/react-native/issues/36643). I was surprised because I assume other things should have been failing without that setting so I'm wondering if I'm missing something? @paulschreiber @david-code


shrouxm:
@david-code did you end up having time to check this? it's definitely still needed for me locally, but doesn't seem to be necessary for remote builds

Issue #127, opened 2023-06-29
Create slope data input screen


shrouxm:
not including slope aspect and landscape position

PR #670, opened 2023-12-06
feat: slope screen

shrouxm:
## Description
Implements the slope screen, with the steepness and shape inputs. To be implemented in future issues: slope aspect/landscape position inputs, slope meter. Note the new components `ImageRadio`, `DataInputSummary`, and `LastModified` to be re-used across the various data input screens. The PR is a bit of a mouthful, but I think I managed to keep the commit history pretty clean & explanatory so I recommend reviewing one commit at a time!

```[tasklist]
### remaining work
- [x] data summary card full styling
- [x] import the actual images
- [x] slope shape page
- [x] slope steepness confirmation modal
- [x] make done fab work
- [x] factor out some common patterns
- [x] implement the behavior where the inputs update each other on the manual slope steepness input
- [x] fix slope shape design (will handle when assets are ready)
- [ ] slope meter (will handle in another issue)
```

### Checklist

### Related Issues
Progresses #127


shrouxm:
@CourtneyLee333, working on the shape screen right now, the design doesn't have any way to differentiate the choices for down slope from cross slope:
image

i threw a couple titles and a divider in there but let me know if you want to mock up something more detailed than this:

image

shrouxm:
another thing i'm bumping into, looping @ltseng in as well: the spec shows the original app as only having 7 options for slope steepness, and there are only 7 graphics in the folder, but the spec gives 9 steepness options. options for proceeding:
1. drop 2 steepness options
2. re-use images for similarly steep options
3. create new images for the remaining options and try to match style
4. create an entirely new image set for all 9 options

2 is lowest effort, 4 is highest effort and most correct, no idea if 1 is viable but it would be more correct than 3 and similar effort. i will implement 2 for now while waiting for resolution so it doesn't hold any other work up

ltseng:
Thanks for the update @shrouxm - you made the right call for now. I'll work with Laura and/or Courtney to get new images drawn

shrouxm:
i was curious if we're there yet lol, not quite:
image

CourtneyLee333:
I went in a different direction for slope shape, [using the image seen here](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=138%3A13384&mode=design&t=n0DkTppczngwlINM-1) and without the two columns. However, since you already have this built we might just go with it, and yes we'd need headings for each of the column to make it clear that there are two choices to be made. Hold on this until @ltseng weighs in.

ltseng:
Sorry, I only saw Garo's first comment about the missing images. I wasn't responding at all to the cross slope vs. down slope discussion.

@CourtneyLee333 your image link above takes me to this:



But I assume you were referring to this?



In which case, yes, I do like what Garo has done, and I will update the image that Laura is writing help text for. Please go ahead with refining the design.

CourtneyLee333:
@ltseng I was referring to this (I guess you can't link to an image in figma) 
Screenshot 2023-12-15 at 11 43 14 AM

ltseng:
Sorry, I'm all turned around. @shrouxm how difficult would it be to redo the slope shape input and return to Courtney's original designs?

shrouxm:
ohhh sorry i misunderstood! i didn't realize the image was actually tied to the design for the input. i definitely see how laying it all out like that could be less confusing, but do we have assets for that? (also what do the C, S, and V stand for in that image? conCave, conVex, Straight?). i could cut squares out of that 3x3 image to make individual images for each option but they'd still have the text labels. assuming we have the assets, the effort would be pretty low

david-code (approved):

PR #698, opened 2023-12-15
feat: reorganize and reimplement the entire navigation logic

igorbej:
### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


dev-client/src/screens/SettingsScreen.tsx
```diff
@@ -0,0 +1,9 @@
+import {View, Text} from 'react-native';
```
paulschreiber, line 1: issue: add copyright notice.

paulschreiber (changes requested):

paulschreiber:
@david-code @shrouxm Please review this PR.

@igorbej This is very helpful. Thank you. Please update the PR description to disucss what you have done and why, and how the navigation system now works.

shrouxm (changes requested):
i'm requesting changes as i've found some non-trivial regressions introduced by this PR and i have some questions/concerns about the overall approach

my overall question is: what are the pros of the react navigation approach to defining screens as opposed to the `ScreenScaffold` and `AppBar` that were removed? from my perspective, the pros are:
- already pretty well implemented and maintained
- simpler for simple cases
while the cons are:
- more complicated for more complicated cases (e.g. dynamic screen titles from redux store, app bar interactions with the screen)
- breaks the react abstraction
- forces you to use their style system

my experience writing the app so far has been that to implement the design i would have to write cumbersome or error-prone workarounds to react navigation to get several more complicated use-cases to work. `ScreenScaffold` and `AppBar` required some effort to get right in the first place but have been much more idiomatic/cleaner from then on

my concern with this PR is that we have already done the work to get `ScreenScaffold` and `AppBar` working, and by switching back to react navigation's approach we're then restricted to the quirks of its implementation as we're unlikely to do the work once again to get `ScreenScaffold` and `AppBar` working, whereas right now if there's something about the way the screens work we'd like to change it's much more straightforward and idiomatic

shrouxm:
thank you so much for the super detailed comment @igorbej, this is very helpful for me having never built a React Native app or used React Navigation before! i don't have time to really dig into this right now, but i'm going to record my initial thoughts since with more explanation i can see a lot of ways the two approaches aren't necessarily in conflict with each other:
- if you have pointers on what kind of abstractions we could use for our apps' context that better integrates with React Navigation that would be appreciated, but i know you're already working overtime on this so no worries either way! (specific goals of the abstraction were to keep configuration of a screen local to the screen's implementation file, and make creating screens more idiomatic to the React component model)
- i definitely agree with you re: it being a bad idea to re-implement navigation logic/components, i would like to move away from this. a potential compromise here is using the internal components React Navigation provides (which won't provide an identical experience, but could be a big improvement)
- i didn't know about the authenticated flows/modal presentation APIs, those both seem like great ideas to implement!
- it is really good to know what is expected/idiomatic for an experienced React Native developer. something we also need to take into account is what it'll be like to introduce a developer _without_ React Native experience to the codebase, since we often don't have the bandwidth to be able to hire people with extensive experience on our specific tech stack. the abstractions feel a lot more idiomatic to me coming from a more general react and UI programming background (i'm sure another flutter programmer would recognize where the `Scaffold` idiom comes from :upside_down_face:)

i think the next step for this would be to spend a bit more time reading over this PR and the React Navigation docs and then make a list of navigation improvements that can be made independently that we can peel off when the issue they resolve has become a bigger pain point. there are much higher prio issues than the navigation experience papercuts we have right now, and i don't think there's time to try to get a big overhaul like this right, so i'll mark the PR as draft for now until we can synthesize the knowledge in it elsewhere

thank you again for taking the time to write all this up Igor, much appreciated!

PR #725, opened 2023-12-27
feat: slope meter

### LINKED ISSUE #127: Create slope data input screen

shrouxm:
## Description
Adds a slope meter. This is stacked on top of #670 and should be merged into main after #670 is merged. Reviewer: please ensure that the iOS app still builds and the feature functions properly before approving.

### Checklist

### Related Issues
Closes #127


dev-client/android/settings.gradle
```diff
@@ -8,3 +8,6 @@ buildCache {
         removeUnusedEntriesAfterDays = 30
     }
 }
+
+apply from: new File(["node", "--print", "require.resolve('expo/package.json')"].execute(null, rootDir).text.trim(), "../scripts/autolinking.gradle")
+useExpoModules()
```
paulschreiber, line 13:
```suggestion
useExpoModules()

```

dev-client/package.json
```diff
@@ -46,7 +46,11 @@
     "react-native-vector-icons": "^10.0.2",
     "terraso-client-shared": "github:techmatters/terraso-client-shared#23fb61c",
     "uuid": "^9.0.1",
-    "yup": "^1.3.2"
+    "yup": "^1.3.2",
+    "expo": "^49.0.0",
+    "expo-screen-orientation": "~6.0.6",
+    "expo-sensors": "~12.3.0",
+    "expo-camera": "~13.4.4"
```
paulschreiber, line 53: question: we use `^` elsewhere. should we do the same here?

paulschreiber:
@shrouxm after updating the podfile and Xcode project, it builds and runs (in the iOS simulator).

dev-client/ios/Podfile
```diff
@@ -26,6 +27,29 @@ if linkage != nil
...
+    begin
+      expo_patch_react_imports!(installer)
+    rescue => e
+      Pod::UI.warn e
+    end
+    begin
+      expo_patch_react_imports!(installer)
+    rescue => e
+      Pod::UI.warn e
+    end
```
paulschreiber, line 51: question: does this need to be run four times?

shrouxm: probably not? the ios and android file changes were autogenerated by the expo install command, but i had to run it several times with minor changes to get it to work so it might just be a sloppy script that adds the command each time it is run instead of checking if it already exists

dev-client/package.json
shrouxm: `~` was also an auto generated by expo, it specifies a narrower range than `^` (`^` is a version range which means "this version and any patch or minor version updates to it", and `~` means "this version and any patch version updates to it"). more details [here](https://docs.npmjs.com/cli/v6/using-npm/semver#tilde-ranges-123-12-1). i'm not really sure without thinking further about it if one or the other is more appropriate for our needs, but i lean towards `^` for consistency absent any specific reason not to. thoughts?

shrouxm:
> @shrouxm after updating the podfile and Xcode project, it builds and runs (in the iOS simulator).

can you run it on an iOS device instead of the simulator to check if the camera works? (unless the camera works in a simulator? [the docs](https://docs.expo.dev/versions/latest/sdk/camera/#usage) say it does not)

dev-client/package.json
paulschreiber: I would prefer `^` as well.

dev-client/ios/Podfile
paulschreiber: Agreed. Please remove the duplication (and perhaps file a bug on the installer script).

paulschreiber (approved):

Issue #120, opened 2023-06-27
bug: app renders unusably slowly


shrouxm:
## Description
This is seen most easily by trying to switch between different screens on the bottom tabs, but also applies when leaving form screens (e.g. Create Site), navigating to nested screens (e.g. navigating through nested data input screens), or causing screen updates (e.g. forms with updating helper text). The behavior is intermittent but has been getting overall worse over time.

ltseng:
Still replicable on version 40 on iOS and Android

shrouxm:
@ltseng could you provide a bit more info? i haven't really been seeing this issue and first reported it when the app was very new, so i'm not sure what form this bug takes anymore. specific questions:
- are there particular screen transitions that are slow, or could it be any screen transition?
- how slow is the screen transition? (less than one second but noticeable lag? more than one second? inconsistent?)
- how often does it happen? (every time? some of the time? rarely?)

ltseng:
I specifically see this on iPhone when switching between the Home and Projects tab. I'm not sure if it's because the touch target is somehow smaller on the iPhone, but when there is a delay, it's consistently less than a second, and sometimes the screen fails to switch at all.

shrouxm:
ok thank you for the extra info! from watching the video, i'm guessing it's because of missing the touch target

we could try to resolve by making the touch targets as big as possible (i.e. extend all the way to the top and bottom of the bar, and halfway to the next button in the bar horizontally). that could potentially introduce a different frustrating interaction where users accidentally change screens when trying to tap elements near the bottom bar, so there might be a trade-off there

it seems like it happens pretty infrequently at this point and the work to do the fix is straightforward but the results might be finnicky so i will update the issue to reflect that

ltseng:
Yeah, I wouldn't consider the bug that I encountered to be a high-priority fix at all.

ltseng:
I think @CourtneyLee333 has encountered an issue with seeing screens sometimes taking up to 5 or 10 seconds to respond though, so I'll try to get more details or a screenrecording from her whenever she is able to consistently replicate it.

shrouxm:
that sounds like what i used to experience that originally prompted me to file this issue, but i stopped experiencing it at some point, so would be good to know if it's still happening on other devices!

CourtneyLee333:
It happened on my ipad today, but my app was out of date. I'll keep an eye out for it.

ltseng:
[...]

PR #787, opened 2024-01-19
feat: performance improvements

igorbej:
This PR relates to the performance issues discussed in https://github.com/techmatters/terraso-mobile-client/issues/120

## Summary
- I haven’t found anything that would affect iOS _specifically_, as brought up in the above issue
- That being said, I saw multiple performance problems that make the app perform very poorly in general (shouldn’t be heavily skewed towards one platform though, but more on that later)
- Since my time was limited and we’re working from different time zones, I decided my best bet was to try and resolve (or at least point out to you) as many issues as I could, hoping either:
    - whatever perf gains I could achieve would diminish the significance of the unidentified iOS-only problem, or
    - fortuitously, I resolve/ameliorate the problem, because it’s not actually restricted to iOS, but rather was only showing up on iOS in your tests due to some other quirks (like the specifics of your setup) we currently don’t know about
- Following this strategy, I improved what was impacting the performance the most according to my tests and I managed to make some nice performance gains even with only some of the issues I observed resolved. More results below, but tl;dr is:
    - the RAM usage (on iOS, release build) is stable at around 200-300MB (previously I could easily get it to 700MB+, depending on the amount of time spent in the app)
    - there are significantly fewer instances of sudden frame drops throughout the app, and the ones that do persist are less severe
    - the performance of FlatLists is slightly better (but could be improved even further with more work)
    - CreateProjectScreen is much more responsive to user typing compared to other input-based screens (say, CreateSiteScreen, which I did not update). With more time, analogous improvements could be made for the remaining text inputs/screens that use them throughout the app and you can use CreateProjectScreen as a reference here.

## Details

To reiterate and to be extra explicit: I haven’t been able to replicate an identical iOS lag/freeze as the one in https://github.com/techmatters/terraso-mobile-client/issues/120, but judging by how that looks in the recording, but also the app’s behaviour, its patterns of using system resources, the load it puts on the main and the JS threads, my current theory is that it’s not a singular issue causing the lags, but rather multiple unrelated issues compounding to slowly but surely smother the app (especially the JS thread). And that results in lags and unresponsiveness like the one you saw in your tests, where the app’s business logic part (i.e. JavaScript) is so overwhelmed it doesn’t appropriately process the touch events (i.e. button presses) coming to it from the native world.

### The issues

Of the issues I found: some I resolved immediately; some, specifically ones that required more time than I had available or some discussions/decisions on your end, I solved partially and/or marked in code with comments.

Here's what I've noticed:
- Creating a new site redirects the user back to the HomeScreen but it’s not the original HomeScreen, a new one is created each time a new site is created, which quickly gobbles up system resources
- There are at least two very non-performant FlatLists in the app:
    - the one on the HomeScreen, in the BottomSheet, which displays a list of sites
    - the one on the ProjectListScreen, which displays a list of projects
- At any given moment the app renders **a lot** of elements that are not displayed in the viewport, e.g. because they belong to a different screen, etc. (HomeScreen, ProjectViewScreen or LocationDashboardScreen are good examples of this; you can investigate it with React Native's iOS performance monitor.)
- The process of memoization is a two-step one and in many places it is only implemented halfway through, meaning the memoization doesn't have an effect. (WIP)
- Formik forms use input fields that:
    - make the whole (heavy) Form component rerender with every keystroke, rather than only rerendering the input field
    - are controlled (i.e. receive a `value` prop) which is mostly fine on the web, but can be problematic for React Native due to its asynchronous nature and the fact the JS <> Native communication is happening over the bridge (see [here](https://github.com/facebook/react-native/issues/20119) and [here](https://www.youtube.com/watch?v=83ffAY-CmL4&t=1366s) for more context). And since the app is struggling already, it is indeed problematic here, resulting in input fields being extremely laggy, sometimes to the point of updating only after a multiple-second delay
- NativeBase components are used extensively which is especially painful in the case of FlatList items, which should be as lean as possible
    

### iOS vs Android

I would argue that Android suffers from the same performance problems as iOS as they are mostly JS-specific, and during my tests I could, for example, easily crash the Android emulator (debug build, Google Pixel 3) due to insufficient RAM after creating just a handful (~4-5) of new sites. Frame drops and some lags were also present.

That being said, the particular interaction of switching back and forth between the HomeScreen and ProjectListScreen tabs, which in your implementation uses the Stack navigator (rather than the Bottom Tabs navigator appropriate for this use case, which we've talked about in https://github.com/techmatters/terraso-mobile-client/pull/698), is different on Android and iOS (a sort of fade-in on Android vs a slide from the edge of the screen on iOS), as the underlying native primitives also differ. And these differences might be the reason why the overall poor performance of the app affects this interaction unevenly on the two platforms. Although I have to say I'm speculating here, this isn't a conventional use case and I'd encourage you again to consider using [Bottom Tabs](https://reactnavigation.org/docs/bottom-tab-navigator) for your bottom navigation (or [Material Bottom Tabs](https://reactnavigation.org/docs/material-bottom-tab-navigator) if animations/transition effects are what you're looking for).

### iOS

- Memory usage, release build

| Before | After |
|---|---|
| ios memory before | ios memory after |

- Number of views rendered, debug build

| Before | After |
|---|---|
| ios views rendered before | ios views rendered after |

- Other stuff

| frame drops | 3gb ram usage |
|---|---|
|  Serious JS thread frame drops | Extreme RAM usage on the latest `main` (debug build, but still) |

- Text input recordings
(WIP)

### Android

As noted earlier, I had no trouble getting Android to 700MB+ memory usage too, at which point it was crashing:
android

Providing it with more memory resulted in bigger consumption:
Screenshot 2024-01-25 at 13 58 50

Plus the overall experience was laggy, just like the iOS one, therefore I'm sceptical all of these issues are relevant for iOS exclusively.

## Other remarks

- The bottom navigation could benefit from some visual feedback when the tabs are pressed
- I agree with @CourtneyLee333 that the map preview for SiteCard items isn't very beneficial for the users, and considering that FlatList items should be kept as lightweight as possible I recommend you remove it. Plus it's very glitchy visually during rendering on iOS in my experience.
- As another avenue to explore in terms of performance you might consider looking at your [bridge](https://www.callstack.com/blog/react-native-how-to-check-what-passes-through-your-bridge). (But I'm not an expert here and I'd leave that for later and take care of the more obvious issues first.)

### Checklist
- [ ] New tests added

### Related Issues
Relates to #120 

(^Added by @shrouxm, I sincerely hope the app can clear this bar with these changes, but it is possible that more work will be required to achieve that for every scenario/interaction.)

### DIFF
--- a/dev-client/src/screens/CreateProjectScreen/components/Form.tsx
+++ b/dev-client/src/screens/CreateProjectScreen/components/Form.tsx
@@ -24,7 +24,7 @@ import {
   TextArea,
   VStack,
 } from 'native-base';
-import {Formik, useFormikContext} from 'formik';
+import {Formik} from 'formik';
 import {RadioBlock} from 'terraso-mobile-client/components/RadioBlock';
 import {IconButton} from 'terraso-mobile-client/components/Icons';
 import {useTranslation} from 'react-i18next';
@@ -130,6 +130,7 @@ export const EditForm = ({
   submitProps,
 }: Omit<FormProps, 'privacy'>) => {
   const {t} = useTranslation();
+
   return (
     <Formik<Omit<ProjectUpdateMutationInput, 'id'>>
       validationSchema={editProjectValidationSchema(t)}
@@ -163,12 +164,17 @@ export const EditForm = ({
   );
 };
 
-export default function Form({editForm = false, onInfoPress}: Props) {
+// TODO(performance): Adjust types
+export default function Form({
+  editForm = false,
+  onInfoPress,
+  handleChange,
+  handleBlur,
+  privacy,
+}: Props) {
   const {t} = useTranslation();
-  const {handleChange, handleBlur, values} =
-    useFormikContext<ProjectFormValues>();
+
   const inputParams = (field: keyof ProjectFormValues) => ({
-    value: values[field],
     onChangeText: handleChange(field),
     onBlur: handleBlur(field),
   });
@@ -183,7 +189,13 @@ export default function Form({editForm = false, onInfoPress}: Props) {
     <VStack space={3}>
       {EditHeader}
       <FormLabel>{t('projects.create.name_label')}</FormLabel>
-      <Input placeholder={t('projects.add.name')} {...inputParams('name')} />
+      <Input
+        placeholder={t('projects.add.name')}
+        defaultValue=""
+        autoCorrect={false}
+        scrollEnabled={false}
+        {...inputParams('name')}
+      />
       <ErrorMessage fieldName="name" />
 
       <FormLabel>{t('projects.create.description_label')}</FormLabel>
@@ -192,6 +204,7 @@ export default function Form({editForm = false, onInfoPress}: Props) {
         numberOfLines={3}
         fontSize={16}
         autoCompleteType="off"
+        autoCorrect={false}
         {...inputParams('description')}
       />
       <ErrorMessage fieldName="description" />
@@ -211,7 +224,7 @@ export default function Form({editForm = false, onInfoPress}: Props) {
           PRIVATE: {text: t('projects.create.private')},
         }}
         groupProps={{
-          value: values.privacy,
+          value: privacy, // TODO(performance): Investigate whether this input _really_ has to be controlled
           variant: 'oneLine',
           onChange: handleChange('privacy'),
           name: 'data-privacy',
--- a/dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
+++ b/dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
@@ -88,6 +88,14 @@ export const CreateSiteForm = ({
         <VStack p="16px" pt="30px" space="18px">
           <FormField name="name">
             <FormLabel>{t('site.create.name_label')}</FormLabel>
+
+            {/* TODO(performance):
+            1. All text inputs should be adjusted into uncontrolled components
+               - https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components
+               - https://github.com/facebook/react-native/issues/20119
+            2. The code should be rewritten/reorganized/memoized so that a change to a text input's value doesn't
+            trigger a re-render of the whole Form/Component (chokes the JS thread and results in terrible frame drops)
+            */}
             <FormInput placeholder={t('site.create.name_placeholder')} />
           </FormField>
           <FormField name="coords">
--- a/dev-client/src/screens/ProjectListScreen/components/ProjectList.tsx
+++ b/dev-client/src/screens/ProjectListScreen/components/ProjectList.tsx
@@ -16,20 +16,49 @@
  */
 
 import {Box, FlatList, Text} from 'native-base';
+import {useCallback} from 'react';
 import {useTranslation} from 'react-i18next';
 
 import {Project} from 'terraso-client-shared/project/projectSlice';
 import {useListFilter} from 'terraso-mobile-client/components/ListFilter';
 import {ProjectPreviewCard} from 'terraso-mobile-client/screens/ProjectListScreen/components/ProjectPreviewCard';
 
+// TODO(performance):
+// Some relevant reading: https://reactnative.dev/docs/optimizing-flatlist-configuration#windowsize
+const WINDOW_SIZE = 6; // Default is 21, bringing it down as the items in this FlatList are very costly to render
+const MAX_TO_RENDER_PER_BATCH = 5; // Similar as above, default is 10
+// const ITEM_HEIGHT = ??
+const SEPARATOR_HEIGHT = 8;
+
+type RenderItemProps = {
+  item: Project;
+};
+
 export const ProjectList = () => {
   const {t} = useTranslation();
   const {filteredItems} = useListFilter<Project>();
+
+  const renderItem = useCallback(
+    ({item}: RenderItemProps) => <ProjectPreviewCard project={item} />,
+    [],
+  );
+
+  // TODO(performance): Consider adjusting ProjectPreviewCard with the design team
+  // to always have a fixed height so you can leverage getItemLayout (example below):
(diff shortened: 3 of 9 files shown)


dev-client/src/screens/CreateProjectScreen/components/Form.tsx
```diff
@@ -163,12 +164,17 @@ export const EditForm = ({
   );
 };
 
-export default function Form({editForm = false, onInfoPress}: Props) {
+// TODO(performance): Adjust types
```
paulschreiber, line 167: What sort of adjustments are needed?

dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
```diff
@@ -88,6 +88,14 @@ export const CreateSiteForm = ({
...
           <FormField name="name">
             <FormLabel>{t('site.create.name_label')}</FormLabel>
+
+            {/* TODO(performance):
+            1. All text inputs should be adjusted into uncontrolled components
+               - https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components
+               - https://github.com/facebook/react-native/issues/20119
+            2. The code should be rewritten/reorganized/memoized so that a change to a text input's value doesn't
+            trigger a re-render of the whole Form/Component (chokes the JS thread and results in terrible frame drops)
+            */}
```
shrouxm, line 98: could you expand on how this would work? the form components aren't controlled by props when they're in a form, and are memoized. or are you saying that the entire architecture of Formik is causing the forms to re-render on each keystroke and we need to move away from the library?

dev-client/src/screens/ProjectListScreen/components/ProjectList.tsx
```diff
@@ -16,20 +16,49 @@
...
+    [],
+  );
+
+  // TODO(performance): Consider adjusting ProjectPreviewCard with the design team
+  // to always have a fixed height so you can leverage getItemLayout (example below):
+  // const getItemLayout = useCallback((_: any, index: number) => {
+  //   const length = ITEM_HEIGHT + SEPARATOR_HEIGHT;
+  //   const offset = length * index;
+  //   return {length, offset, index};
+  // }, []);
```
shrouxm, line 52: this is kind of a general UI question, but if you're specifying fixed heights in pixels in advance, how do you handle something like a user who has set a large font size preference for accessibility reasons?

shrouxm (commented):
thanks igor! the suggested changes all make sense, i left a couple questions about the details of implementing some of them if you have time to respond. if you don't have time to keep working on the PR i'll take it over

were you able to benchmark the app and identify any performance bottlenecks any particular performance bottlenecks? and did the changes you made here make a significant difference in iOS performance or is there still more work to do to get things usable on that platform?

dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
igorbej: Sure, I'll try to elaborate on this further next week

dev-client/src/screens/ProjectListScreen/components/ProjectList.tsx
igorbej: Need to look into that, but off the top of my head, I think it doesn't have to be fixed-fixed but rather deterministic and repeated for all items, great question though

dev-client/src/screens/CreateProjectScreen/components/Form.tsx
igorbej: Nothing major, just regular props type adjustments I omitted here to save a bit of time

shrouxm (approved):
thank you @igorbej! i've started working on my own PR #823  based on this one with further work, including replacing some of our usages of NativeBase components. i rebased this PR onto current main, but just saw you said you might do some more work this week so hopefully that won't cause any conflicts! (the rebase was conflict free so i think it'll be ok). you can check out my PR to make sure we aren't doing any overlappping work, besides what's pushed there i'm now just working on replacing some more NativeBase components because that seemed to be the most impactful change in terms of overall performance improvements (as opposed to fixing bugs with specific interactions). if you don't have time, i'm happy to take over this PR and merge it this week!

shrouxm:
@igorbej one thing i was curious your opinion on, i enabled react navigation's [react-freeze](https://github.com/software-mansion/react-freeze) support in my PR. it made a huge difference with a few problem screens, but it's marked experimental and i was curious if you had any thoughts on pros/cons of the feature for our app!

dev-client/src/screens/ProjectListScreen/components/ProjectList.tsx
shrouxm: i see, so it'd work but you'd need to find a way to compute that dynamically. i've run into this problem with other UI frameworks before, never understood why there can't be an option to say "i promise they'll all be the _same_ dimension" instead of "i promise they'll all be _this_ dimension" so the framework can just measure one item's layout and use that value for the rest of the list which seems much less error-prone

shrouxm:
@igorbej ok sounds good! some updates from today:
- i built a new iOS release based on my branch which is stacked on your branch
- things are looking wayyy better on my device, so i released it to teammates to validate it's also working on the devices that originally reported the worst issues
- if it looks good there too, then i'm just going to focus on getting these PRs mergeable and validate that there aren't any regressions
- i've made a lot of progress on my branch in the last couple days so i'd say if you do want to make more code changes, base them on my branch rather than continue to update this one
- but i think it'll make sense to wait and see if we do actually need to put more effort into this right now, because if perf is looking ok then it'll definitely be higher prio to avoid regressions and finish building out the features for MVP than try to eke out some more frames

shrouxm:
@igorbej ok team confirmed this is resolved on all (our) devices so we're switching focus back to feature work & no need to keep working on it! i'm just going to spend today getting these PRs merged and fixing any style regressions

thank you so much for your insight on all of this!

shrouxm:
closing as these changes were merged into main with #823

PR #850, opened 2024-02-09
feat: surface cracking screen

### LINKED ISSUE #473: feat: vertical cracking option

shrouxm:
## Description
Implements the surface cracking screen.

### Checklist

### Related Issues
Fixes #473 

### Verification steps
User should be able to select vertical cracking options from soil screen. Data completion/requirements should display correctly.


dev-client/src/screens/SoilScreen/components/SoilSurfaceScreen.tsx
```diff
@@ -15,14 +15,76 @@
...
+              value={crack}
+              label={t(`soil.verticalCracking.${crack}`)}
+            />
+          ))}
+        </Select>
+        <Box height="lg" />
+        <Paragraph>{t('soil.verticalCracking_description')}</Paragraph>
+        <Box width="100%" alignItems="center">
+          <Image
+            source={require('terraso-mobile-client/assets/surface/verticalCracking.png')}
```
paulschreiber, line 78:
```suggestion
            source={require('terraso-mobile-client/assets/surface/vertical-cracking.png')}
```

dev-client/src/translations/en.json
```diff
@@ -434,6 +434,12 @@
       "photos": "Photos",
       "notes": "Notes"
     },
+    "verticalCracking_description": "Vertical cracks may appear on the soil surface when some soils experience changes in moisture or temperature. The cracks may be permanent, or they may disappear after some time.\n\nVertical cracks are at least 5 mm (3/16”) wide and 25 cm (10”) deep. You should be able to place a pencil inside the cracks.",
```
paulschreiber, line 437: issue: use either cm or in depending on what units the user has selected. Add additional keys to the JSON and pass them in as parameters.

paulschreiber (changes requested):
note: I compressed the image with TinyPNG. Remember to do that. (Weirdly, this one kept shrinking on subsequent runs. I gave up after a few.)

shrouxm:
@paulschreiber i did actually run this one through tinypng, didn't think to put it through again as i assumed it'd be idempotent

dev-client/src/translations/en.json
shrouxm: @CourtneyLee333 do you want text in tooltips/info sheets to respond to the user's preferred units?

CourtneyLee333: Would that mean calculations need to get made, since inch and cm, yard and meter, etc aren't exactly the same? @ltseng do you have a thought about this?

shrouxm: for info sheets we would just be choosing different text depending on the units, so we can choose when it's appropriate to write e.g. 3/4" vs 0.7567234571248" etc. in this case rather than the text:
> Vertical cracks are at least 5 mm (3/16”) wide and 25 cm (10”) deep.

The user would see only one of the following depending on their preferred units:
> Vertical cracks are at least 5 mm wide and 25 cm deep.
> Vertical cracks are at least 3/16” wide and 10” deep.

CourtneyLee333: I think that sounds ideal!

paulschreiber (approved):

Issue #129, opened 2023-06-29
Create soil color data input screen


shrouxm:
```[tasklist]
### things remaining to be implemented
- [x] translating munsell values to rgb values
- [x] properly handling remembering camera preferences (may require storing an extra flag on the backend for each depth interval)
- [x] full height sheet tooltip component
- [x] confirmation modals/clearing completed values
- [x] picking image
- [x] cropping image
- [x] computing munsell value from image + reference
- [x] do the assets properly
- [ ] better cropping UX
- [x] keep app responsive while color algorithm is running (not sure if there's a good way to do this)
- [ ] handle color algorithm failure more gracefully
- [x] implement modal to let user choose whether to select an existing image or take a new one with camera
- [x] display selected image better on color analysis screen
- [x] fix hue ordering
- [ ] display N hue value
```

ltseng:
Please note an updated requirement in the user story:

In the US app, the Munsell color value should be the default/first choice. In the International app, the photo selection process should be the default/first choice.

shrouxm:
moved remaining work to a new issue so i can close this one

PR #882, opened 2024-02-22
feat: implement soil color algorithm

shrouxm:
## Description
Ports the soil color algorithm code to our codebase. Implements the image picker/cropping workflow to create the inputs to the soil color algorithm, and saves the output. Also finishes filling out the assets for the color guide, and properly displays munsell colors.

Note: most of the added lines are from a giant precomputed table of munsell color values, so the PR is a lot smaller than it looks.

### Checklist

### Related Issues
Next step for #129. See that issue for a list of remaining work for the soil color screen.

Depends on https://github.com/techmatters/terraso-client-shared/pull/289.

### Verification steps
User should be able to use the color algorithm to crop a photo from their library and get an estimated munsell color. The accuracy of the estimate can vary wildly depending on the lighting conditions in the photo. See here for some example images you can use to test: https://drive.google.com/drive/folders/1gIYC6yxpHXJ962QkFPx2TuqFG9Z20GWU


shrouxm:
@paulschreiber in `munsellTable.ts` there is an unedited comment from the original app describing attribution for the values in that file, not sure what to make of it re: licensing. I'm also not sure why a table is needed to compute the RGB values when the process to compute munsell values from RGB seems to mathematically be invertible, but i was going for "just get it working" for this PR. If the table is necessary, it can definitely be stripped of most of its entries.

dev-client/package.json
```diff
@@ -24,15 +25,18 @@
...
-    "expo-av": "^13.10.5",
-    "expo-camera": "^14.0.4",
-    "expo-screen-orientation": "^6.4.1",
-    "expo-sensors": "^12.9.1",
+    "expo": "~50.0.6",
+    "expo-av": "~13.10.5",
+    "expo-camera": "~14.0.4",
+    "expo-image-picker": "~14.7.1",
+    "expo-screen-orientation": "~6.4.1",
+    "expo-sensors": "~12.9.1",
```
paulschreiber, line 33: question: why do we want to use `~` instead of `^`?

dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts
```diff
@@ -0,0 +1,209 @@
+import quantize from 'quantize';
```
paulschreiber, line 1:
```suggestion
/*
 * Copyright © 2024 Technology Matters
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Affero General Public License as published
 * by the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU Affero General Public License for more details.
 *
 * You should have received a copy of the GNU Affero General Public License
 * along with this program. If not, see https://www.gnu.org/licenses/.
 */

import quantize from 'quantize';
```

dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts
```diff
@@ -0,0 +1,209 @@
...
+        }
+      }
+      return [];
+    });
+    //Quantize.js performs median cut algorithm, and returns a palette of the "top 16" colors in the picture
+    var cmap = quantize(pixelArray, COLOR_COUNT);
+    return cmap ? cmap.palette() : null;
+  };
+
+  //Get the color palettes of both soil and card
```
paulschreiber, line 157:
```suggestion
  // Get the color palettes of both soil and card
```

dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts
```diff
@@ -0,0 +1,209 @@
...
+    const pixelArray = pixels.flatMap(([r, g, b, a]) => {
+      // If pixel is mostly opaque and not white
+      if (a >= 125) {
+        if (!(r > 250 && g > 250 && b > 250)) {
+          return [[r, g, b] as quantize.RgbPixel];
+        }
+      }
+      return [];
+    });
+    //Quantize.js performs median cut algorithm, and returns a palette of the "top 16" colors in the picture
```
paulschreiber, line 152:
```suggestion
    // Quantize.js performs median cut algorithm, and returns a palette of the "top 16" colors in the picture
```

dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts
```diff
@@ -0,0 +1,209 @@
...
+  return xyy ? rgb_delinearize(xyy_to_rgb_linear(xyy)) : undefined;
+};
+
+export const getColor = (
+  pixelCard: RGBA[],
+  pixelSoil: RGBA[],
+  referenceRGB: RGB,
+) => {
+  const getPalette = (pixels: RGBA[]) => {
+    //Transform pixel info from canvas so quantize can use it
```
paulschreiber, line 142:
```suggestion
    // Transform pixel info from canvas so quantize can use it
```

paulschreiber (changes requested):
question: did you run the images through TinyPNG?

paulschreiber:
I compressed the images with TinyPNG.

dev-client/package.json
shrouxm: because expo makes breaking changes on minor version increments. (this is a bit reductive but) `~` means "this version and any patches to it", and `^` means "this version and any minor upgrades or patches to it". expo recommends `~` and uses it by default when you `expo install` a new package because they don't try to avoid breaking changes in minor upgrades. this was the source of e.g. this problem that snagged you & david for a while at the beginning of the month: https://tech-matters.slack.com/archives/C02K02L3Q48/p1706818388930309

shrouxm:
@paulschreiber i did compress with tinypng, i ran them through a few times until they were only getting around `1%` savings on each pass

dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts
```diff
@@ -0,0 +1,209 @@
...
+  else if (value <= 3.5) colorValue = 'VALUE_3';
+  else if (value <= 4.5) colorValue = 'VALUE_4';
+  else if (value <= 5.5) colorValue = 'VALUE_5';
+  else if (value <= 6.5) colorValue = 'VALUE_6';
+  else if (value <= 7.5) colorValue = 'VALUE_7';
+  else if (value <= 8.25) colorValue = 'VALUE_8';
+  else if (value <= 8.75) colorValue = 'VALUE_8_5';
+  else if (value <= 9.25) colorValue = 'VALUE_9';
+  else colorValue = 'VALUE_9_5';
+  /* eslint-enable curly */
```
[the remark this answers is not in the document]
shrouxm: yes, but it would be harder to read without any safety benefits

dev-client/src/screens/SoilScreen/ColorScreen/ColorAnalysisScreen.tsx
```diff
@@ -64,17 +91,20 @@ export const ColorAnalysisScreen = (props: ColorAnalysisProps) => {
...
-              updateDepthDependentSoilData({
-                siteId: pitProps.siteId,
-                depthInterval: pitProps.depthInterval.depthInterval,
-                ...analyzeImage({photo, reference, soil}),
-              }),
-            );
-            navigation.pop();
+            const analysisResult = analyzeImage({photo, reference, soil});
+            if (analysisResult !== undefined) {
+              console.log(analysisResult);
```
paulschreiber, line 96:
```suggestion
```

dev-client/src/screens/SoilScreen/ColorScreen/utils/hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+    const response = await launchImageLibraryAsync({
+      mediaTypes: MediaTypeOptions.Images,
+      base64: true,
+    });
+    if (response.canceled) {
+      return;
+    }
+    const image = response.assets[0];
+    if (!image.base64) {
+      throw new Error('missing base64 from image picker!');
```
paulschreiber, line 33: question: what would case this to happen?

paulschreiber (approved):

dev-client/src/screens/SoilScreen/ColorScreen/utils/hooks.ts
shrouxm: assuming no bugs in the dependency, it would be because we fudged a config setting elsewhere and didn't instruct the image picker to return the base64 encoded data. i.e. the type of `image.base64` is `string | undefined`, with the actual value being one or the other depending on the arguments to `launchImagePickerAsync`. so here we're just making sure to fail fast if we do make a coding error and it's not coming back as `string` like we expected

PR #898, opened 2024-02-26
fix: Update README and environment sample

paulschreiber:
## Description
* Clean up ZSH vs Bash instructions for Ruby
* Make Mapbox instruction consistent across platforms
* Add Android SDK instructions
* fix: add SENTRY_PROJECT to env sample


tm-ruxandra (approved):
LGTM, the changes all track with regard to what we went over together.

PR #899, opened 2024-02-26
fix: don't enable Sentry if DSN is not defined

paulschreiber:
## Description
Don't enable Sentry if DSN is not defined. Matches backend behaviour.


dev-client/src/App.tsx
```diff
@@ -46,12 +46,16 @@ import * as Sentry from '@sentry/react-native';
 import {enableFreeze} from 'react-native-screens';
 import {HomeScreenContextProvider} from 'terraso-mobile-client/screens/HomeScreen/HomeScreen';
 
+const sentryEnabled = APP_CONFIG.sentryDsn;
```
shrouxm, line 49: nit:

```suggestion
const sentryEnabled = Boolean(APP_CONFIG.sentryDsn);
```

shrouxm (commented):
questions: would it be better to just require sentry dsn to be defined, do we ever want to not have sentry enabled? i've found it useful even when doing local dev. and if we do want to disable it, could we have a separate required config flag `sentryEnabled`? with the change as is, if we make a build configuration error and sentryDsn is not set we will silently not have sentry enabled in any new builds. that seems like a more likely build configuration error to make than accidentally explicitly setting `sentryEnabled` to `false`

paulschreiber:
@garo If someone who is not Tech Matters runs Terraso, they may not have Sentry installed/configured. I want the app to be able to launch successfully for them.

shrouxm:
@paulschreiber that makes sense, in that case i have the same question

> could we have a separate required boolean config flag sentryEnabled? with the change as is, if we make a build configuration error and sentryDsn is not set we will silently not have sentry enabled in any new builds. that seems like a more likely build configuration error to make than accidentally explicitly setting sentryEnabled to false

dev-client/.env.sample
```diff
@@ -7,3 +7,4 @@ PUBLIC_MAPBOX_TOKEN=""
 SENTRY_DSN=https://xyz@o4506299756642304.ingest.sentry.io/abcd
 SENTRY_DISABLE_AUTO_UPLOAD="true"
 SENTRY_PROJECT="terraso-lpks"
+SENTRY_ENABLED="false"
```
shrouxm, line 10: question: why are some of these quoted and some not?

dev-client/src/App.tsx
```diff
@@ -46,12 +46,16 @@ import * as Sentry from '@sentry/react-native';
 import {enableFreeze} from 'react-native-screens';
 import {HomeScreenContextProvider} from 'terraso-mobile-client/screens/HomeScreen/HomeScreen';
 
+const sentryEnabled = Boolean(APP_CONFIG.sentryEnabled);
```
shrouxm, line 49: nit: the `Boolean` call is redundant

dev-client/src/config/index.ts
```diff
@@ -99,6 +100,10 @@ if (Config.SENTRY_DSN === undefined) {
   throw new Error('Config setting SENTRY_DSN not set');
 }
 
+if (Config.SENTRY_ENABLED === undefined) {
+  throw new Error('Config setting SENTRY_ENABLED not set');
+}
```
shrouxm, line 105: nit:
```suggestion
if (Config.SENTRY_ENABLED !== 'true' && Config.SENTRY_ENABLED !== 'false' ) {
  throw new Error(`Config setting SENTRY_ENABLED set to invalid value: ${Config.SENTRY_ENABLED}`);
}
```

shrouxm (approved):
couple non-blocking nits/questions

dev-client/.env.sample
```diff
@@ -7,3 +7,4 @@ PUBLIC_MAPBOX_TOKEN=""
 SENTRY_DSN=https://xyz@o4506299756642304.ingest.sentry.io/abcd
```
paulschreiber, line 7:
```suggestion
SENTRY_DSN="https://xyz@o4506299756642304.ingest.sentry.io/abcd"
```

dev-client/.env.sample
```diff
@@ -4,6 +4,7 @@ GOOGLE_OAUTH_IOS_CLIENT_ID="abcdefghijklmnopqrstuvwxyz.apps.googleusercontent.co
 TERRASO_BACKEND=https://api.staging.terraso.net
```
paulschreiber, line 4:
```suggestion
TERRASO_BACKEND="https://api.staging.terraso.net"
```

dev-client/.env.sample
```diff
@@ -1,9 +1,10 @@
 ENV=development
```
paulschreiber, line 1:
```suggestion
ENV="development"
```

dev-client/.env.sample
paulschreiber: no idea. i quoted all of them for consistency.

Issue #910, opened 2024-02-27
tech-debt: migrate to expo prebuild


shrouxm:
we currently have a bare react native app, and integrate with expo's libraries. if we used expo's prebuild feature, we could:
- delete the android/ios directories
- have easier upgrade paths between RN releases
- do all of the app configuration in one place
- potentially do releases without the awkward commit workflow we currently use

shrouxm:
progress update: the CI pipeline is working, and successfully uploading app bundles to the app store. there is an issue that only happens in the CI pipeline with the way environment variables get passed into the app, will try to resolve tomorrow.

PR #911, opened 2024-02-27
feat: expo prebuild migration

shrouxm:
## Description
migrates our build process to use expo prebuild, so we don't have to maintain android and ios directories, have a smoother upgrade process, and can configure the app builds in one place. would also enable future workflow improvements like triggering new releases from git tags rather than having to make new commits.

### Checklist

### Related Issues
Resolves https://github.com/techmatters/terraso-mobile-client/issues/910

### Verification steps
App builds, releases, and works as before.


dev-client/app.config.ts
```diff
@@ -0,0 +1,71 @@
+import {ExpoConfig, ConfigContext} from 'expo/config';
```
paulschreiber, line 1: issue: add copyright notice.

dev-client/app.config.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const appVersion = 82;
```
paulschreiber, line 3: suggestion: can we read this from a text or JSON file?

shrouxm: what i would like to implement but haven't yet implemented is to read this from an environment variable. then our github workflows can provide it based on commit tags, so rather than creating and merging a commit updating the value in a config file, the codebase itself has no idea what version it is and we can push a tag `v86` to a branch to tell the CI pipeline to build that commit as build number 86. this will also be a necessary step for e.g. building multiple different apps from one codebase

dev-client/app.config.ts
shrouxm: i don't understand which files need to have copyright notices and which don't. e.g. why don't we have one in `metro.config.js`? or in our yaml files?

paulschreiber: `metro.config.js` should have one too. That was an oversight. I hadn't thought about the YAML files before, just source code. We can add notices there too.

paulschreiber (changes requested):
This looks great! Two concerns:

* linter error in `app.config.ts`
* need to be able to bump the build number separately from the version number

shrouxm:
@paulschreiber fixing the lint, copying my comment from slack re: bumping build/version numbers:

yes, it works that way right now because internal testers on android can only see the version string and not the build number, so we actually need to increment the patch version each time to be able to refer to specific builds when speaking with testers

here i just made iOS work the same way because i don't think it's quite possible to get the exact same workflow we had before with expo's setup, but it makes it more consistent (since now we can just say build "1.0.101" instead of "1.0.101 on android, 1.0 (101) on iOS")

in the future, we could either:
- have the user-visible version string hard coded in the repo, and push tags which just increment the build number
- push a tag format like v1.0.2-101, and parse that into display 1.0.2 to users on each platform and setting the internal build number to 101

paulschreiber (approved):

PR #929, opened 2024-03-07
style: clean up Podfile with Rubocop

paulschreiber:
## Description
Clean up Podfile with Rubocop.


shrouxm (commented):
i'm not sure it matters too much as this file will be auto-generated with expo prebuild, but i'm wondering whether style changes like this would make it more difficult to apply diffs when upgrading RN versions using https://react-native-community.github.io/upgrade-helper ?

shrouxm (approved):
whoops meant to approve as i don't have strong feelings either way

paulschreiber:
This will go away soon anyway, but it bugs me to have non-linted source kicking around.

PR #936, opened 2024-03-12
fix: slope-class-names

ltseng:
## Description
Changing 1-15% to 2-5%, as pointed out on a demo 

### Related Issues
No formal issue filed, it was pointed out in the weekly BLM meeting just now


ltseng:
@shrouxm can you help me make sure that the underlying data type for this value is actually 2-5%, not 1-15%?

paulschreiber (approved):

PR #938, opened 2024-03-13
feat: many color algorithm improvements

shrouxm:
## Description
Many improvements to the color algorithm workflow:
- algorithm should now run in under a second
- modal warnings when clearing completed color values
- displays full size image on color analysis screen
- app remembers which workflow user used for a particular depth interval
- allow user to take a new photo with camera and save it to their library

### Checklist

### Related Issues
Aaaaaalmost there for #129 
Needs https://github.com/techmatters/terraso-client-shared/pull/336


dev-client/src/translations/en.json
```diff
@@ -683,5 +650,9 @@
         "LINEAR": "Linear"
       }
     }
+  },
+  "image": {
+    "use_camera": "Use Camera",
+    "choose_from_gallery": "Choose From Gallery"
```
paulschreiber, line 656:
```suggestion
    "use_camera": "Take Photo",
    "choose_from_gallery": "Photo Library"
```

dev-client/src/screens/SoilScreen/utils/renderValues.ts
```diff
@@ -30,13 +30,6 @@ export const renderDepthInterval = ({
   return `${label ? `${label}: ` : ''}${start}-${end} cm`;
```
paulschreiber, line 30: issue: This should use i18n.t instead of concatenating strings in code. Also: unit should not be hard-coded.

paulschreiber (approved):

dev-client/src/translations/en.json
shrouxm: feedback for copy best lives in the figma file

PR #941, opened 2024-03-13
fix: improve depth interval and camera / photo library strings

paulschreiber:
## Description
• Use i18n strings for renderDepthInterval
• Use en dashes for depth intervals
• Use standard language for camera/photo library picker

### DIFF
--- a/dev-client/src/translations/en.json
+++ b/dev-client/src/translations/en.json
@@ -482,6 +482,7 @@
         "overlaps": "Overlaps existing",
         "ordering": "Interval bounds error"
       },
+      "bounds_labelled": "{{label}}: {{start}}–{{end}}{{units}}",
       "bounds_unitless": "{{start}}–{{end}}",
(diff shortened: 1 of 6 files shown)


dev-client/src/translations/en.json
```diff
@@ -652,7 +653,7 @@
     }
   },
   "image": {
-    "use_camera": "Use Camera",
-    "choose_from_gallery": "Choose From Gallery"
+    "use_camera": "Take Photo",
+    "choose_from_gallery": "Photo Library"
```
shrouxm, line 657: @CourtneyLee333 do you have thoughts on changing this copy?

shrouxm (commented):
code changes look good, i will defer to courtney about the copy

dev-client/src/translations/en.json
CourtneyLee333: I'm seeing a lot of different terminology across platforms and apps. I'm fine to start with the suggested text shown here, and see how it works out.

Issue #915, opened 2024-03-01
Update select form style


CourtneyLee333:
Select list form fields need to be styled correctly throughout the app. Form styles and expected behaviors are outlined in Notes for Engineers:

[Notes for Engineers: Form Styles](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=2647%3A10445&mode=design&t=ZSMVryxmVDtcwJvi-1)

Edit: outstanding issues:

```[tasklist]
### Tasks
- [ ] LandPKS Data Privacy sheet should have square, rather than rounded corners
```

CourtneyLee333:
@ltseng Do we need to list all the locations in the app where this needs to be fixed?

ltseng:
I really hope not 😂

ltseng:
Work is mostly done, ended up touching a couple other things, so there's more tech debt refactoring work that will be included here.

Bringing in React Native Paper library to handle the appropriate text field label/placeholder behavior.

ltseng:
Looks good on Android, waiting for iOS build to be fixed before I can finish testing it.

DerekCaelin:
We are closing this issue and will file new bugs as needed.

PR #959, opened 2024-03-20
feat: refactor Select, modals/cards, and some code reorganization/consistency-ifying

shrouxm:
## Description
A bunch of refactoring/style updating work. This PR:
- brings in the react-native-paper library to help us a bit with styling and do some utility functions NativeBase was doing for us
- reorganizes/renames a bunch of things for consistency
- replaces the NativeBase modal component, and also updates the API of the BottomSheetModal and Card component to be consistent and allow for fixing some outstanding bugs
- replaces the NativeBase select component

I went back through and made a cleaner git commit history, so this is best reviewed one commit at a time to view logically related changes next to each other, and also because some commits could definitely be a decent sized PR on their own.

### Checklist

### Related Issues
Resolves, once QA'd, #915, and probably several other open bugs that i will try to track down another time.

### Verification steps
Styles on affected screens should look closer to / match the Figma.


dev-client/tsconfig.json
```diff
@@ -9,5 +9,10 @@
     }
   },
   "include": ["src", "__tests__", "jest"],
-  "exclude": ["dist"]
+  "exclude": ["dist"],
+  "ts-node": {
+    "compilerOptions": {
+      "module": "commonjs",
```
paulschreiber, line 15: question: can we use ESM here? we will need to use some ESM dependencies.

dev-client/src/translations/en.json
```diff
@@ -294,9 +291,9 @@
     "example_email": "name@domain.com",
     "profile_image_alt": "Profile image",
     "role": {
-      "viewer": "Viewer",
-      "contributor": "Contributor",
-      "manager": "Manager"
+      "MANAGER": "Manager",
+      "CONTRIBUTOR": "Contributor",
+      "VIEWER": "Viewer"
```
paulschreiber, line 296: question: why are these all-caps? why did the order change?

dev-client/src/screens/SoilScreen/components/SoilSurfaceScreen.tsx
```diff
@@ -59,16 +65,12 @@ export const SoilSurfaceScreen = ({siteId}: Props) => {
...
-          {surfaceCracks.map(crack => (
-            <Select.Item
-              key={crack}
-              value={crack}
-              label={t(`soil.vertical_cracking.value.${crack}`)}
-            />
-          ))}
-        </Select>
+          nullable
+          value={cracking ?? null}
```
paulschreiber, line 69: question: do you need the `?? null`? Do we need to have `null` instead of `undefined`?

dev-client/src/screens/SoilScreen/components/DepthDependentSelect.tsx
```diff
@@ -17,32 +17,32 @@
...
   siteId,
   depthInterval,
   ...props
 }: Props<K>) => {
-  const currentValue = useSelector(
-    selectDepthDependentData({siteId, depthInterval}),
-  )?.[input];
+  const currentValue =
+    useSelector(selectDepthDependentData({siteId, depthInterval}))?.[input] ??
+    null;
```
paulschreiber, line 45: question: is this just to convert undefined to null?

dev-client/src/screens/SlopeScreen/SlopeShapeScreen.tsx
```diff
@@ -149,7 +149,7 @@ export const SlopeShapeScreen = ({siteId}: Props) => {
       <Fab
         onPress={() => navigation.pop()}
         leftIcon={<Icon name="check" />}
-        label={t('general.done_fab')}
+        label={t('general.done')}
```
paulschreiber, line 152: 👍

dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
```diff
@@ -106,7 +106,10 @@ export const SiteSettingsScreen = ({siteId}: Props) => {
               icon={<Icon name="info" />}
             />
           </FormControl.Label>
-          <Select dropdownIcon={<Icon name="arrow-drop-down" />} />
+          <ProjectSelect
+            projectId={site.projectId ?? null}
+            setProjectId={_ => {}}
```
paulschreiber, line 111: question: what does `_ => {}` do?

dev-client/src/screens/ScreenScaffold.tsx
```diff
@@ -48,26 +47,23 @@ export const ScreenScaffold = ({
...
-    </BottomSheetModalProvider>
+    <SafeAreaView
+      style={[
+        styles.safeAreaContainer,
+        {backgroundColor: colors.primary.dark},
+      ]}>
+      <StatusBar
+        barStyle="light-content"
+        translucent
+        backgroundColor={'#00000000'}
```
paulschreiber, line 58: question: can this be #000,  #000000 or a name color?

dev-client/src/components/util/transformNativeBase.ts
```diff
@@ -15,62 +15,107 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
```
paulschreiber, line 17: question: what does this file do? is it temporary?

dev-client/src/components/inputs/Select.tsx
```diff
@@ -0,0 +1,162 @@
...
+      ref={ref}
+      scrollable={false}
+      CloseButton={null}
+      trigger={onOpen => (
+        <Pressable
+          accessibilityState={{disabled}}
+          onPress={disabled ? null : onOpen}>
+          <TextInput
+            editable={false}
+            value={value === null ? '' : renderValue(value)}
```
paulschreiber, line 134: is this equivalent?
```suggestion
            value={value ? renderValue(value) : null}
```

dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
shrouxm: oh good catch! that was a placeholder function during an earlier pass, it's a NOOP function but should definitely be an op!

dev-client/src/screens/ScreenScaffold.tsx
shrouxm: #000 and #000000 mean different things (the extra 0s make it fully transparent) but it should be named yes, not actually part of the PR but i will update it

dev-client/src/components/util/transformNativeBase.ts
shrouxm: it is a dev-time script to do structured code analysis/refactoring, sort of temporary. it's how i've been tracking how many native base components are left and automatically rewriting import statements. there's no real harm to having it in the `src` tree so it didn't seem worth it to figure out how to have it somewhere else. i will either delete it or move it out of `src` when the native base migration is done

dev-client/src/components/inputs/Select.tsx
shrouxm: they aren't equivalent no, the type of value is generic so if its type was number and the value was 0 (which is falsy) then `value ? renderValue(value) : null` would return `null` when we want it to return `renderValue(0)`

dev-client/src/screens/SoilScreen/components/SoilSurfaceScreen.tsx
shrouxm: well we designed the API for `Select` so we don't _need_ to have it, but `null` felt more semantically appropriate to me and allowing either felt a bit disorganized

dev-client/src/translations/en.json
shrouxm: because they are constant enum values (i.e. they are actually values that get passed around in the app, not just i18n keys), so they're delineated by being caps lock. in practice, it means you can't change this key by just changing it and `CTRL+F`ing for `MANAGER` because it might be getting passed in from a variable, so the caps are saying "be careful when changing this!"

dev-client/tsconfig.json
shrouxm: i don't know, this is a dev only config thing for the `transformNativeBase` script i explained below so i just did whatever made it work and then moved on

dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
shrouxm: ok i went back in, this was already a NOOP before, and implementing it properly now would require a confirmation modal and generally non-trivial changes so i don't want to also pull that into this PR

paulschreiber (approved):

PR #970, opened 2024-03-26
fix: Update GitHub CI build caching

### LINKED ISSUE #969: ci: update build caching logic for expo
i think the expo update broke the pod/gradle caching logic. i also think the entire `android` and `ios` directories now only depend on the `package-lock.json` and `app.config.ts` files and can be cached in their entirety by keying on those files. but not sure exactly what the best thing to do here is.

paulschreiber:
## Description
* Get sentry project and org from environment
* cache `android` and `ios` folders

### Related Issues
Fixes #969.

### DIFF
--- a/.github/workflows/build-shared-android.yml
+++ b/.github/workflows/build-shared-android.yml
@@ -17,6 +17,14 @@ on:
         required: false
         type: string
         default: 'v1'
+      SENTRY_ORG:
+        required: false
+        type: string
+        default: 'techmatters'
+      SENTRY_PROJECT:
+        required: false
+        type: string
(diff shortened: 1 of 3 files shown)


shrouxm (approved):
changes look good, i will trigger some no-op builds with it when i'm done with the other release and check it's all working

shrouxm (changes requested):
i looked at the second run and it looks like it didn't work, i think because the `prebuild` step does a 'clean' prebuild. i will try with the 'clean' rebuild disabled

shrouxm (approved):

Issue #916, opened 2024-03-01
Fix help and info icon inconsistency


CourtneyLee333:
I noticed on the Create Site form that the icons for help and info buttons are different sizes and colors. They should be consistent per [the design](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=2910%3A16054&mode=design&t=vJjFzKbLvlL4v8uR-1).

Please look out for other places where this might be the case, and I will too.

PR #1061, opened 2024-04-01
fix: use same color for info and help icons on create site form

paulschreiber:
## Description
Use the same color for info and help icons on create site form.

image

### Related Issues
This will address #916.


tm-ruxandra (approved):
lgtm

CourtneyLee333:
Please note they are also currently different sizes but should be the same size and color as defined in Figma.

Issue #974, opened 2024-03-27
feat: finish soil color


shrouxm:
soil color is almost done! the algorithm runs fast, and returns reasonably accurate results under ideal conditions. the following work remains to close out this feature:

```[tasklist]
### Remaining work
- [x] The cropping UX is not ideal. Currently, if you exit out of a crop screen, you will lose the progress of both crops. And if you re-enter a crop screen, you lose progress on the crop. I think it would also be nice to show a dimmed outline around the crop area instead of the crop area going directly to the edge of the frame, but that's not so crucial.
- [x] The color algorithm currently fails silently if a color is returned outside of what we consider a valid munsell soil color. We should instead show a modal displaying the computed color and informing the user what has happened, and offer them the choice to either proceed with the algorithm's value anyway, or offer them the closest valid munsell soil color to the color the algorithm returned.
- [x] We currently take a very broad view of what is considered a valid munsell soil color. We'd like to consider exactly the set of colors in the munsell soil color book to be valid. There's a mostly complete table here: https://docs.google.com/spreadsheets/d/1mW6sCFKyu3FrorivCB-PfQcLxs3uq_s-/edit#gid=1464184837, so we just need to cross-reference that with the book to implement this feature.
- [x] We always display a hue, even for very low chroma values. It's standard practice to abbreviate munsell colors with low chroma values with the `N` hue, so we should do this when displaying color strings
- [ ] https://github.com/techmatters/terraso-mobile-client/issues/1063
```

CourtneyLee333:
@shrouxm Is there any design work you want me to do for the first item? If so, let's discuss because I don't quite follow. Otherwise, I trust you to build what you're thinking and I can review it. 

I think the only other piece in the list for me is the second item, for which I have created ticket #593.

shrouxm:
@CourtneyLee333 re first item: nope! I just hadn't implemented the natural assumptions for how navigating between the screens would behave in the versions so far.

oh i think i see the confusion, i should've written "The cropping UX _as currently implemented_ is not ideal.", the designs don't specifically say anything about the way it works right now because no one would assume it should be implemented the way it works right now :upside_down_face:

DerekCaelin:
@shrouxm I did some testing of the photo processing tool, results are [here.](https://docs.google.com/spreadsheets/d/1XuqQspff4TptXOuG5UHs-i6RkRuqhrcrMqCltyqYLGo/edit#gid=2004954072)

[...]

PR #1064, opened 2024-04-03
feat: invalid color confirmation modal

shrouxm:
## Description
Displays the following modal when the color algorithm returns an invalid color. Note that there are forthcoming improvements to figuring out whether the color is valid and returning a suggestion, just keeping the PRs small so this is just the modal and some very rough heuristics. There's some noisy refactor commits first, so might be easier to review one commit at a time.

### Checklist

### Related Issues
Progress on #974 

### Verification steps
Should see the following modal when using the color algorithm and inputting nonsensical values
image


dev-client/src/components/modals/ActionsModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export type ActionsModalProps = Pick<ModalProps, 'trigger'> & {
+  title: string;
+  actions: React.ReactNode;
+};
+/**
+ * Modal presented to a user when asked to confirm a decision
+ */
+export const ActionsModal = forwardRef<
```
paulschreiber, line 92: question: should this be `ActionModal`? Other uses are signular.

paulschreiber: suggestion: add file-level comment explaining how ActionModal, ConfirmModal and Modal differ.

paulschreiber (approved):

dev-client/src/theme.ts
```diff
@@ -190,16 +190,6 @@ export const theme = extendTheme({
           width: 'full',
           justifyContent: 'flex-start',
         },
-        confirmModal: {
```
tm-ruxandra, line 193: Question: was this dead code? I don't see it replicated anywhere else

dev-client/src/screens/SoilScreen/ColorScreen/components/SwitchWorkflowButton.tsx
```diff
@@ -82,6 +82,7 @@ export const SwitchWorkflowButton = ({
       title={t('soil.color.confirm_delete.title')}
       body={t('soil.color.confirm_delete.body')}
       actionName={t('soil.color.confirm_delete.action_name')}
+      isConfirmError
```
tm-ruxandra, line 85: question: to confirm my understanding of react syntax here -- will this default `isConfirmError` to true? or is it being passed from somewhere that i've missed?

paulschreiber: I believe this is equivalent to `isConfirmError={isConfirmError}`.

shrouxm: this making me think i should maybe not use this syntax. it's equivalent to `isConfirmError={true}`. in an object, e.g.
```
{ 
  actionName: t('...'),
  isConfirmError
}
```
is the same as `isConfirmError: isConfirmError`, but in JSX syntax it's different. presumably because JSX tries to look like html, where e.g. `<input type="checkbox" checked/>` means "this is checked"

dev-client/src/theme.ts
shrouxm: this was one of the things i was curious to ask your opinion about re: refactoring the components to be more re-usable/separate styling/content concerns

when a specific component wants to have its own bespoke styling, i was thinking it makes more sense to define those styles next to the component rather than in the global theme file. the global theme file would be for style values that need to be re-used across many components

shrouxm: but open to feedback here!

dev-client/src/components/modals/ActionsModal.tsx
shrouxm: re: `Action` vs `Actions`, i chose `ActionsModal` as it's a modal which can contain many actions, whereas `ActionButton` is a button to describe a particular action. i updated the comments as you asked to make it clearer

PR #1095, opened 2024-04-11
docs: better docs on running the app locally

shrouxm:
better docs

they make the docs better

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,15 +83,20 @@ EOF
 $ sudo apt install node watchman openjdk-17-jdk
 ```
 
-## Android
+# Running the app locally
 
-1. Run `npm run start` to start the dev server.
-2. Run `npm run android` to load the app on an emulator or connnected physical device.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,15 +83,20 @@ EOF
 $ sudo apt install node watchman openjdk-17-jdk
 ```
 
-## Android
+# Running the app locally
+
+Do **not** use the `a` or `i` subcommands in `npm run start`, they don't work with our workflow.
```
knipec, line 88: Ohh I thought it was specifically the `s` subcommand that does cloud stuff and doesn't work with our workflow, was that incorrect?

knipec (approved):
Yay thanks!

README.md
shrouxm: `s` is kind of a no-op, it switches between expo go and dev builds, neither of which we use, so `a` or `i` don't work reliably in either mode

PR #1097, opened 2024-04-11
feat: Add Sign In With Apple

paulschreiber:
## Description
Add Sign In With Apple (via authentication).

## Related issues
Addresses https://github.com/techmatters/terraso-product/issues/680
Depends on https://github.com/techmatters/terraso-backend/pull/1301

### DIFF
--- a/dev-client/src/auth/index.ts
+++ b/dev-client/src/auth/index.ts
@@ -26,6 +26,7 @@ import {
   IssuerOrDiscovery,
   resolveDiscoveryAsync,
 } from 'expo-auth-session';
+import * as AppleAuthentication from 'expo-apple-authentication';
 
 type AuthConfig = AuthRequestConfig & {issuer: IssuerOrDiscovery};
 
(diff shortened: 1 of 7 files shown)


dev-client/src/auth/index.ts
```diff
@@ -78,7 +82,11 @@ async function exchangeToken(
 ) {
   const payload = await request<AuthTokens>({
     path: '/auth/token-exchange',
-    body: {provider, jwt: identityJwt},
+    body: {
+      provider,
+      client_id: Constants.expoConfig!.ios?.bundleIdentifier,
```
knipec, line 87: Suggestion/question: Could this use the `APP_CONFIG.appleClientId` from config/index.ts? Right now it seems like the logic of "use the bundleIdentifier as the client id" is happening in two places and am wondering if there's a way to have one place that's the source of truth for the client id

knipec (approved):

PR #1101, opened 2024-04-12
chore: update app icons to fix border/padding

paulschreiber:
## Description
Update app icons to fix border/padding. Use adaptiveIcon for Android.


dev-client/app.config.ts
```diff
@@ -86,7 +86,10 @@ export default ({config}: ConfigContext): ExpoConfig => ({
   android: {
     package: 'org.terraso.landpks',
     versionCode: appVersion,
-    icon: 'src/assets/landpks-round-1024.png',
+    icon: 'src/assets/landpks-ios-1024.png',
```
tm-ruxandra, line 89: question: Is this correct? It looks like you're linking the ios icon under an android config section

dev-client/app.config.ts
```diff
@@ -101,7 +104,7 @@ export default ({config}: ConfigContext): ExpoConfig => ({
   ios: {
     bundleIdentifier: 'org.terraso.test.Terraso-LandPKS',
     buildNumber: '0',
-    icon: 'src/assets/landpks-square-1024.png',
```
knipec, line 104: question: Why are we no longer using a square image? Aren't iOS icons supposed to be square with rounded edges as opposed to circular?

paulschreiber: This is per @CourtneyLee333's request and was approved/decided on.

dev-client/app.config.ts
paulschreiber: Will rename for clarity.

tm-ruxandra (approved):

PR #1137, opened 2024-04-19
fix: Soil Color Guide matches design + move general info modals to same directory

### LINKED ISSUE #1115: Update Soil Color Guide to match the design + tangential refactoring
From [this Slack thread](https://tech-matters.slack.com/archives/C0376RCCXD2/p1713491698883739):
1. Separate out metric vs imperial strings
2. Add in the correct Step 5 and move the existing Step 5 to be Step 6
3. Check with Courtney about: 
- 2mm sieve: ok to be 1/16" when in imperial?
- Do we care that we don't have the registered trademark symbol by "Post-It"?
- Is it ok if Step 5 is its own white box? Or is it important to be in the same white box as Step 4?

Also some refactoring that I didn't want to clutter PR #1100 so will do as a separate item:
1. PR comment to move color out of SlopeMeterScreen into themes.ts --> Garo just did this in their PR!
2. Organize directory structure for info modal files

knipec:
## Description
In Soil Color Guide:
- Separate out metric vs. imperial strings
- Add the correct Step 5 and move the existing Step 5 to be Step 6
Also: 
- Organize directory structure for info modal files

### Checklist

### Related Issues
Fixes #1115 

### Verification steps
1. See Soil Color Guide and confirm that step "2. Prepare soil sample" says "2mm sieve" and "anything larger than 2mm". There's no way to see imperial units in the app currently, but I promise it would look like this: 
Color-Propsed-Imperial
2. Also in Soil Color Guide, confirm that step "5. Compose photo." exists and matches the [Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=6168-21844&mode=design&t=oMk90TZVExCQdqPL-0)


paulschreiber (approved):

paulschreiber:
@knipec this looks good. Please add a prefix to the PR title so the conventional commits check will pass.

PR #1175, opened 2024-04-26
feat: snapshot integration testing!

### LINKED ISSUE #1167: testing: create integration tests covering basic styling of core app screens
having gotten a "my first test" to run using detox on android and validated the ability to mock external dependencies, this issue covers the effort to:
- implement some screenshot-based golden tests to ensure style consistency for as many of the following screens as will fit into an effort 3 issue:
  - home screen
  - create site
  - create project
  - site dashboard
  - project dashboard
- ensure that everything is set up in a way that teammates can also add tests and run the suite

notably, to keep scope small this issue does _not_ cover:
- running detox on iOS
- running detox in CI

shrouxm:
## Description
Adds a number of snapshot tests of screens, and sets them up to run in CI. I'd appreciate feedback from people if this is mergeable as is, and then can be iterated on in further PRs, or if we'd like to see more progress before merging this.

Basically what's happened here is I've set up a render hook which can accept an initial state for the redux store and a route name, and found a way to get the house of cards of jest mocks for our dependencies to produce reasonable snapshot results. I then fudged around with the implementation files to confirm the snapshots also fudge in the same way. There's many many dimensions for future improvement (testing more screens, testing screens for different sets of data loaded, interacting with screens, coverage statistics, etc) but I wanted to check in if this is making sense for people.

This is a bit of a pivot from my original plan for integration testing, which was to use detox to set up screenshot testing. I ended up getting stuck on that again, and decided to go with this method. It provides less comprehensive/more granular coverage and won't be able to catch e.g. platform bugs, but it is simpler and more robust and most importantly actually works!

I rebased my first native base migration PR on this and it gave me some useful feedback and didn't totally crap out so yay!

One screen I haven't been able to get working with this is the home screen :/ I think the offending component is the `BottomSheetFlatList`, which is pretty core so I don't want to stub it out. I'm also not sure how mocking the map will work, so I'm punting on that screen for now even though it is literally the home screen.

### Checklist

### Related Issues
Fixes #1167

### DIFF
--- a/dev-client/__tests__/integration/LoginScreen-test.tsx
+++ b/dev-client/__tests__/integration/LoginScreen-test.tsx
@@ -1,0 +1,34 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {LoginScreen} from 'terraso-mobile-client/screens/LoginScreen';
+import {render} from '@testing/utils';
+
+test('renders correctly', () => {
+  expect(
+    render(<LoginScreen />, {
+      initialState: {
+        account: {
+          currentUser: {
+            data: null,
+            fetching: false,
+          },
+        } as any,
+      },
+    }).toJSON(),
+  ).toMatchSnapshot();
+});
--- a/dev-client/__tests__/integration/LoginScreen-test.tsx
+++ b/dev-client/__tests__/integration/LoginScreen-test.tsx
@@ -19,16 +19,16 @@ import {LoginScreen} from 'terraso-mobile-client/screens/LoginScreen';
 import {render} from '@testing/utils';
 
 test('renders correctly', () => {
-  expect(
-    render(<LoginScreen />, {
-      initialState: {
-        account: {
-          currentUser: {
-            data: null,
-            fetching: false,
-          },
-        } as any,
-      },
-    }).toJSON(),
-  ).toMatchSnapshot();
+  const screen = render(<LoginScreen />, {
+    initialState: {
+      account: {
+        currentUser: {
+          data: null,
+          fetching: false,
+        },
+      } as any,
+    },
+  }).toJSON();
+
+  expect(screen).toMatchSnapshot();
 });
--- a/dev-client/babel.config.js
+++ b/dev-client/babel.config.js
@@ -2,7 +2,7 @@ module.exports = api => {
   api.cache(true);
 
   return {
-    presets: ['module:@react-native/babel-preset'],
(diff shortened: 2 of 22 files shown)


shrouxm:
comment: need to check whether tests will fail if there's a new test that doesn't have a corresponding snapshot or if there's an obsolete snapshot that hasn't been deleted

shrouxm:
todo: update README

dev-client/__tests__/integration/LoginScreen-test.tsx
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {LoginScreen} from 'terraso-mobile-client/screens/LoginScreen';
+import {render} from '@testing/utils';
+
+test('renders correctly', () => {
+  expect(
```
knipec, line 22: Suggestion: use the same pattern as the other files, getting the screen variable and doing `expect(screen).toMatchSnapshot();` at the end -- unless there's some reason for the different pattern here?

dev-client/babel.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = api => {
   api.cache(true);
 
   return {
-    presets: ['module:@react-native/babel-preset'],
+    presets: ['babel-preset-expo', '@babel/preset-typescript'],
```
knipec, line 5: Question: Can you explain what this is doing?

dev-client/jest/utils.tsx
```diff
@@ -25,14 +34,48 @@ const nativeBaseInset = {
...
+  initialState,
+  route,
+  children,
+}: React.PropsWithChildren<WrapperProps>) => {
+  const store = useRef(createStore(initialState));
   return (
-    <NativeBaseProvider theme={theme} initialWindowMetrics={nativeBaseInset}>
-      <Portal.Host>{children}</Portal.Host>
-    </NativeBaseProvider>
+    <Provider store={store.current}>
```
knipec, line 49: Question: Out of curiosity, did you just manually go up from a Screen and see what provider-like things there were as parent components?

knipec (approved):
I'm not really following the setup.ts file, but seems reasonable enough!

Could you give an example of what the output looks like when a test fails?

Also it does seem like as we're doing a lot of styling changes we may have to do a lot of updating of the snapshots, which may decrease the general usefulness of the tests outside of the nativebase migration, but I'm down to try it and see how it goes!

dev-client/jest/utils.tsx
shrouxm: a combination of this and running the test and seeing what error messages emerged!

dev-client/babel.config.js
shrouxm: babel is a javascript transpiler (transforms javascript to more compatible/faster javascript). it is used as a pre-processing step before jest tests the code. here we're changing the preset configuration babel uses to presets that better support our codebase. i was confused why the project was working without having the expo/typescript preset, i think it might be because babel is _only_ used when running jest, but i'm not sure

dev-client/__tests__/integration/LoginScreen-test.tsx
shrouxm: nope! that was just the first one i wrote and i forgot to go back and update it

shrouxm:
@knipec here is an example of a test failure:
image

i agree that this style of testing will not be of much use when making intentional big style changes and/or writing new screens. i think this provides value beyond the NB migration in the following ways:
- when _not_ intending to make a big style change, provides security you haven't actually done so
- if you do make a big style change, once it's implemented correctly you can update the snapshots and then if there's any refactoring/code cleanup to do you can be more confident you didn't accidentally break your new change do think it provides value beyond validating the migration
- all of the config changes/CI set up etc is necessary regardless to set up unit tests, so we can also start to think about writing those now!

Issue #1153, opened 2024-04-22
Update Location Details Styling


DerekCaelin:
Update the location dashboard page (site and non-site variants) with information and content needed for Capri launch. Please see parent task for more details.

#### Temporary Location Dashboard
[Design](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-15842&mode=design&t=YHeqcffux9COXQxs-0)

#### Design as of 4/23/2024 - please see figma for the most up-to-date version
![Image](https://github.com/techmatters/terraso-product/assets/10799815/3b555213-ceae-4535-8d61-211b6fbf72ae)

#### Site Location Dashboard 
[Design](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-13434&mode=design&t=rgquQ2rmdqG3pD0r-0)

#### Design as of 4/25/2024 - please see figma for the most up-to-date version
<img width="183" alt="image" src="https://github.com/techmatters/terraso-mobile-client/assets/10799815/a557556f-8566-444c-9bec-caa23abc6e16">

## Acceptance Criteria
- [x] As a user, when I select a location on the map, I want to see different UI elements depending on whether the location Tempory, or a Site, and is viewed by a Manager, Contributor, Viewer, or any logged in user.  [View this spreadsheet](https://docs.google.com/spreadsheets/d/10gAV_Ja-x7mVaZa5hzLnq_yien-qfTnwhzXJiek4aww/edit#gid=0)
    - [x] Temporary location
    - [x] Site location
        - [x] Manager
        - [x] Contributor
        - [x] Viewer
- [x] As a user, I want to see updated styling that matches the design
- [x] Remove "last modified by" field

PR #1212, opened 2024-04-30
fix: update location details screen

paulschreiber:
## Description
Update location details screen to match design:
* remove plus button from app bar
* round GPS accuracy to whole numbers
* change "more info" to "learn more"
* add create site button label and description
* use actual count for people badges

### Related Issues
Addresses #1153.


dev-client/src/screens/LocationDashboardScreen.tsx
```diff
@@ -51,18 +51,9 @@ export const LocationDashboardScreen = ({siteId, coords}: Props) => {
...
-          onPress={() => navigation.navigate('CREATE_SITE', {coords})}
-        />
-      );
-    }
-
-    // display nothing if user does not own the site / is not manager
-    if (userRole === null || !isSiteManager(userRole)) {
+    // display nothing if no site associated with location or
+    // user does not own the site / is not manager
+    if (!siteId || userRole === null || !isSiteManager(userRole)) {
```
knipec, line 56: (Aside) Question: I've mostly seen "site owner" in our code so far. 
@tm-ruxandra Do you know if we have standardized terminology of "site owner" or "site manager"?

dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
```diff
@@ -91,7 +91,7 @@ export const SiteSettingsScreen = ({siteId}: Props) => {
             size="container"
             alignItems="center"
             justifyContent="space-around">
-            <Icon name="people" />
+            <Icon name="people-alt" />
```
knipec, line 94: Issue: Why this change? As far as I can tell we use the "people" icon for the badges. But also this is commented out.

paulschreiber: I checked the Figma file, and it looked closest to `people-alt`. Prior to this PR, we had been using a combination of both.

dev-client/src/translations/en.json
```diff
@@ -258,7 +256,7 @@
       "heading": "Transfer existing sites",
       "description": "Choose existing LandPKS sites you manage to transfer to this project. Note that only those who have access to the project will be able to access the sites in the project.",
       "unaffiliated": "Unaffiliated",
-      "tooltip": "Sites you transfer to this project will inherit this project’s data settings, input requirements, and team members. Learn more..."
+      "tooltip": "Sites you transfer to this project will inherit this project’s data settings, input requirements, and team members. Learn more…"
```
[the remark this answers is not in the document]
paulschreiber: That is weird! I will remove it.

dev-client/src/screens/SiteScreen/components/ProjectInstructionsButton.tsx
```diff
@@ -49,12 +49,12 @@ export const ProjectInstructionsButton = ({project}: Props) => {
         mt={2}
         pl={4}
         pr={4}
-        size="lg"
+        size="md"
         backgroundColor="primary.dark"
```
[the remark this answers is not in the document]
paulschreiber: Ah! That's why it looked funny. I thought maybe it was my screen.

dev-client/src/screens/LocationDashboardScreen.tsx
tm-ruxandra: Unfortunately there's a terminology collision happening here.

For unaffiliated sites, the user responsible for the site is referred to as "site owner" in (most) code and "site manager" in product documents. However, neither of those is consistent with the usage in the helper method here, which defines that either "site owner" or "project manager" is a "site manager". This is something I hope we can standardize on the frontend in the future to avoid precisely this sort of confusion :)

dev-client/src/components/PeopleBadge.tsx
knipec, line 1: Yay! Love this refactor

knipec (approved):

dev-client/src/screens/LocationDashboardScreen.tsx
knipec: Gotcha -- so sounds like today we mostly use 
- "site owner" for unaffiliated sites
- "project manager" for affiliated sites, because they adopt the user roles of their project
- "site manager" for any site, regardless of if it's affiliated or not

I think my ideal would be something like 
- "unaffiliated site owner"
- "affiliated site owner" / "project manager" (?) 
- "site owner"

tm-ruxandra: Maybe we can have a dedicated chat to go over this - I think there's still some misunderstanding going on about terminology. "Project manager" is a project-level role, not a site-level one; affiliated sites definitionally cannot have owners.

PR #1232, opened 2024-05-02
feat: Design updates to form for Create Project, Edit Project, and Create Site

### LINKED ISSUE #534: Styling feedback for project & site creation forms
Update Project Creation form with feedback from designer.

## Designs
See [Capri design](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-12823&mode=design&t=3ueKX3VtS8XyKx1e-0)

## Acceptance criteria
**Project**:
Changes:
- [x] Project Name field needs `required` indicator.
- [x] Set maximum character length for Project Name to 50 characters
- [x] Set maximum character length for Project Description to 72 characters
- [x] Save button should be disabled until required field (Project Name) is completed.
Where to see it:
1. This is mostly about the project creation form (from Projects > Create Project)
2. But behavior should be the same on the Project Settings form (from Projects > an existing project > Settings tab)

**Site**:
Changes:
- [x] Site Name field needs `required` indicator.
- [x] Set maximum character length for Site Name to 50 characters
- [x] Save button should be disabled until required field (Site Name) is completed.
Where to see it:
1. When creating a site (from Home > click map > Create Site)
2. Maybe when editing a site?? (from Home > an existing site > Settings icon in upper right) --> Nope, this is now [#1234](https://github.com/techmatters/terraso-mobile-client/issues/1234)

## Additional Context
[Feedback from Courtney](https://github.com/techmatters/terraso-product/issues/22#issuecomment-1764735754) on initial implementation

knipec:
## Description
Make updates to the form for Create Project, Edit Project, and Create Site

**NOTE:** (Re: Edit Site) I will create an Issue for SiteSettingsScreen to use Formik to make these same updates

### Checklist
- [x] New tests added -- Ohh I'm gonna have to update some test snapshots probably

### Related Issues
Fixes #534 

### Verification steps
1. Click on the Project tab and create a new project.
2. Confirm there is a red "Required" indicator on the name field when it is empty
3. Confirm there is a different red indicator on the name when it has 1 or 2 characters
4. Confirm the name cannot be more than 50 characters
5. Confirm you can have an empty description
6. Confirm the description cannot be more than 72 characters
7. Confirm the Save button is grayed and unpressable when anything is invalid
8. In an existing project, go to Settings. Confirm the same things.
9. Click a location on the homepage map to create a new site. Confirm the same things (minus description ones).


dev-client/src/constants.ts
```diff
@@ -25,10 +25,10 @@ export const LATITUDE_MAX = 90;
 export const LONGITUDE_MIN = -180;
 export const LONGITUDE_MAX = 180;
 export const SITE_NAME_MIN = 3;
-export const SITE_NAME_MAX = 120;
-export const PROJECT_NAME_MAX_LENGTH = 120;
+export const SITE_NAME_MAX = 50;
+export const PROJECT_NAME_MAX_LENGTH = 50;
 export const PROJECT_NAME_MIN_LENGTH = 3;
-export const PROJECT_DESCRIPTION_MAX_LENGTH = 240;
+export const PROJECT_DESCRIPTION_MAX_LENGTH = 72;
```
knipec, line 31: This seems rather short, I commented on #534 asking Courtney why we want 72, but lmk if you have an opinion

dev-client/src/screens/CreateProjectScreen/components/CreateProjectForm.tsx
```diff
@@ -98,7 +111,7 @@ const FormContainer = React.memo(
         <Box position="absolute" bottom={8} right={3} p={3}>
           <Button
             onPress={handleSubmit}
-            disabled={isSubmitting}
+            isDisabled={isSubmitting || !isValid} //TODO-cknipe: Comment in PR: is this what was intended with isSubmitting?
```
knipec, line 114: I'm assuming the new behavior is what was intended originally but I don't think it worked, because when I put the isValid check under `disabled`, it didn't actually appear to disable the button when invalid.

dev-client/src/constants.ts
paulschreiber: Agree this seems too short.

knipec: Courtney still says 72 ([link](https://github.com/techmatters/terraso-mobile-client/issues/534#issuecomment-2090411784)). I figure we can change it easily later if users want longer.

dev-client/src/screens/CreateProjectScreen/components/ProjectForm.tsx
```diff
@@ -149,22 +150,25 @@ export const EditForm = ({
...
+            <Button
+              onPress={handleSubmit}
+              isDisabled={isSubmitting || !isValid}
+              shadow={5}
+              size={'lg'}
+              display={userRole === 'MANAGER' ? 'flex' : 'none'}
+              _text={{textTransform: 'uppercase'}}>
+              {t('general.save_fab')}
+            </Button>
+          </Box>
```
paulschreiber, line 164:
```suggestion
          {userRole === 'MANAGER'  &&
          <Box position="absolute" bottom={0} right={0}>
            <Button
              onPress={handleSubmit}
              isDisabled={isSubmitting || !isValid}
              shadow={5}
              size={'lg'}
              display={'flex'}
              _text={{textTransform: 'uppercase'}}>
              {t('general.save_fab')}
            </Button>
          </Box>
          }
```
suggestion: omit the item instead of hiding it with CSS. We can omit the button if omitting its box breaks the layout.

dev-client/src/constants.ts
paulschreiber: @CourtneyLee333 where did 72 come from? Can we do a round number like 75?

CourtneyLee333:
@paulschreiber I'm pretty sure Carissa is already implementing this with 72, but I'm not attached to the exact number.

dev-client/src/screens/CreateProjectScreen/components/ProjectForm.tsx
knipec: 👍  
Aside: I think the userRole check is maybe not even needed, because I think the entire Project Settings tab is unavailable for a non-manager to navigate to. I'm kind of hesitant to remove it since we don't have tests so I'm not confident there isn't some way though.

paulschreiber (approved):

dev-client/src/screens/CreateProjectScreen/components/ProjectForm.tsx
paulschreiber: That's a good point!

PR #1262, opened 2024-05-06
feat: add elevation to temporary site callout

paulschreiber:
## Description
Add elevation to temporary site callout. 
- Show spinner when it is loading.
- Show unknown on failure.
- Uses nationalmap.gov API

### DIFF
--- a/dev-client/src/screens/HomeScreen/components/TemporarySiteCallout.tsx
+++ b/dev-client/src/screens/HomeScreen/components/TemporarySiteCallout.tsx
@@ -15,14 +15,15 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import {useCallback} from 'react';
-import {Button, Divider} from 'native-base';
+import {useCallback, useMemo, useState} from 'react';
+import {Button, Divider, Spinner} from 'native-base';
 import {Coords} from 'terraso-mobile-client/model/map/mapSlice';
 import {useNavigation} from 'terraso-mobile-client/navigation/hooks/useNavigation';
 import {useTranslation} from 'react-i18next';
 import {Card} from 'terraso-mobile-client/components/Card';
 import {CardCloseButton} from 'terraso-mobile-client/components/CardCloseButton';
 import {CalloutDetail} from 'terraso-mobile-client/screens/HomeScreen/components/CalloutDetail';
+import {getElevation} from 'terraso-mobile-client/services';
 import {
   Column,
   Row,
@@ -31,7 +32,6 @@ import {
 
 const TEMP_SOIL_ID_VALUE = 'Clifton';
 const TEMP_ECO_SITE_PREDICTION = 'Loamy Upland';
-const TEMP_ELEVATION = '2800 feet';
 
 type Props = {
   coords: Coords;
@@ -41,6 +41,15 @@ type Props = {
 export const TemporarySiteCallout = ({coords, closeCallout}: Props) => {
   const {t} = useTranslation();
   const navigation = useNavigation();
+  const [siteElevation, setSiteElevation] = useState('');
+
+  useMemo(async () => {
+    const elevation = await getElevation(
+      parseFloat(coords.latitude.toFixed(5)),
+      parseFloat(coords.longitude.toFixed(5)),
+    );
+    setSiteElevation(t('site.elevation_value', {value: elevation, units: 'm'}));
+  }, [coords, t]);
 
   const onCreate = useCallback(() => {
     navigation.navigate('CREATE_SITE', {coords});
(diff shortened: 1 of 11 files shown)


dev-client/src/screens/HomeScreen/components/TemporarySiteCallout.tsx
```diff
@@ -68,10 +77,14 @@ export const TemporarySiteCallout = ({coords, closeCallout}: Props) => {
           value={TEMP_ECO_SITE_PREDICTION.toUpperCase()}
         />
         <Divider />
-        <CalloutDetail
-          label={t('site.elevation').toUpperCase()}
-          value={TEMP_ELEVATION.toUpperCase()}
-        />
+        {siteElevation ? (
```
tm-ruxandra, line 80: Wouldn't this leave the spinner in place for sites that happen to be at zero sea level? I recognize that this may be very unlikely in practice but it still seems like an incorrect implementation.

paulschreiber: Since this is a string and not a float, `"0"` an `"0.0"` are truthy, so the spinner would go away. I could use `siteElevation !== ''` if you'd prefer.

paulschreiber:
@shrouxm How can I round coords.latitude and coord.longitude to 5 decimal places and then memo-ify that? It'll save an API call for two taps right near each other.

dev-client/src/screens/HomeScreen/components/TemporarySiteCallout.tsx
tm-ruxandra: That might be more clear - wasn't obvious to me that this was a string.

dev-client/src/screens/HomeScreen/components/TemporarySiteCallout.tsx
```diff
@@ -41,6 +41,15 @@ type Props = {
...
   const navigation = useNavigation();
+  const [siteElevation, setSiteElevation] = useState('');
+
+  useMemo(async () => {
+    const elevation = await getElevation(
+      parseFloat(coords.latitude.toFixed(5)),
+      parseFloat(coords.longitude.toFixed(5)),
+    );
+    setSiteElevation(t('site.elevation_value', {value: elevation, units: 'm'}));
+  }, [coords, t]);
```
shrouxm, line 52: suggestions:
  - could move the lat/lon rounding into the `getElevation` call
  - it's more idiomatic in our codebase to store the model value (i.e. the actual numeric elevation returned) in state, and convert it to a string while rendering. but if you want to keep it stored as a string, i agree with ruxandra that i expect it to be a number, so maybe you could change the variable name to `elevationString` or something similar?

dev-client/src/services.ts
```diff
@@ -0,0 +1,39 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+export const getElevation = async (lat: number, lng: number) => {
+  // TypeScript complains if the values passed to URLSearchParams are floats instead of strings.
```
shrouxm, line 19: nit: the reason TypeScript complains is because `URLSearchParams` requires its arguments to be strings, so it'd be simpler/more accurate to just say "URLSearchParams requires its arguments to be strings"

dev-client/src/services.ts
```diff
@@ -0,0 +1,39 @@
...
+  // This API uses X for longitude and Y for latitude. That's not a typo.
+  const params = {
+    x: lng.toString(),
+    y: lat.toString(),
+    units: 'Meters', // TODO: switch based on user preference
+  };
```
shrouxm, line 25: nit: it might be clearer to make the call to `new URLSearchParams` here. then any associated type errors will happen local to the definition, and it'd be more obvious what the params are for. obviously not a big deal for a function this small, but a good general style note i think

dev-client/src/screens/HomeScreen/components/TemporarySiteCallout.tsx
```diff
@@ -41,6 +41,15 @@ type Props = {
...
   const {t} = useTranslation();
   const navigation = useNavigation();
+  const [siteElevation, setSiteElevation] = useState('');
+
+  useMemo(async () => {
+    const elevation = await getElevation(
+      parseFloat(coords.latitude.toFixed(5)),
+      parseFloat(coords.longitude.toFixed(5)),
+    );
+    setSiteElevation(t('site.elevation_value', {value: elevation, units: 'm'}));
```
shrouxm, line 51: issue: if i'm tracing the code right, i think if the API call fails this will render the string "unknown m" instead of "unknown"?

shrouxm:
@paulschreiber could you expand what you mean by `memo-ify` that?

dev-client/src/navigation/navigators/BottomNavigator.tsx
```diff
@@ -35,7 +34,10 @@ export const BottomNavigator = memo(
       state => state.account.currentUser.data !== null,
     );
 
-    const onHome = useCallback(() => navigation.navigate('HOME'), [navigation]);
+    const onSites = useCallback(
```
tm-ruxandra, line 37: These changes are already in main and I don't understand why they're showing up in this PR - worth investigating to confirm.

paulschreiber: I rebased from main. Should be better now.

tm-ruxandra (approved):

Issue #1196, opened 2024-04-29
Soil ID - "Explore the data" button opens new Soil ID sheet


DerekCaelin:
Clicking “Explore the data” button from Soil ID tile takes me to Location Soil ID sheet. This task covers creating the screen, the static content and universal elements. See parent task for context and related tasks. 

### Soil ID Tile 
**Tapping the button in the tile...**
[Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-15842&mode=design&t=bn4bR0g90ayypnJZ-0) - Temporary 
[Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-13455&mode=design&t=bn4bR0g90ayypnJZ-0) - Site

### Soil ID screen
**... leads me to a Soil ID Screen...**
[Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-15872&mode=design&t=bn4bR0g90ayypnJZ-0) - Temporary
[Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-14751&mode=design&t=bn4bR0g90ayypnJZ-0) - Site

**...With these basic elements:**
## Acceptance Criteria:
- [x] If user presses "Explore the data" from a Temporary Location OR a Site view, they see this page (Eng note: we'll want separate components for Temporary Location and Site, but they should share much of their content)
- [x] Screen has correct static text content (Site Name, Soil ID header+text, Top Soil Matches header+"i" button)
- [x] Pressing the "i" icon opens the informational overlay sheet
- [x] Pressing the back arrow returns to the Location Dashboard
- [x] (Temporary Location) Screen has a button to create site, that navigates to the “Create Site” page 

**Non-goals**
- [ ] Sheet does not yet have soil match tiles
- [ ] Top Soil Matches section doesn't yet have content under the header
- [ ] Site Data section does not yet exist

## FYI
Most other stories of "parent" story https://github.com/techmatters/terraso-product/issues/666 depend on this

knipec:
This should be in the next build (v124) to verify.

DerekCaelin:
[...]

PR #1265, opened 2024-05-08
Feat: Soil ID sheet exists for site or temporary location

knipec:
## Description
- Create the Soil ID Sheet and implement most of its static content.
- Also includes some minor refactoring, like renaming SiteScreen to LocationDashboardContent and extracting CreateSiteButton as a component

### Checklist
- [ ] New tests added -- Did not do this, don't think snapshot tests are valuable at this point

### Related Issues
#1196

### Verification steps
1. Press the "Explore the data >" button from a site 
Confirm the components that don't require Soil ID or site data are there
2. Press the "Explore the data >" button from a temporary location
Confirm the components that don't require Soil ID or site data are there


dev-client/src/screens/LocationScreens/components/TopSoilMatchesInfoContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {useTranslation} from 'react-i18next';
+import {Text} from 'terraso-mobile-client/components/NativeBaseAdapters';
+
+export const TopSoilMatchesInfoContent_Site = () => {
```
paulschreiber, line 21:
```suggestion
export const TopSoilMatchesInfoSiteDescription = () => {
```
question: will this have more than the description? that will help name it. If we aren't showing anything else, then perhaps a single component with a siteId or isSite prop would be simpler.

Suggest avoiding underscores for internal consistency.

dev-client/src/translations/en.json
```diff
@@ -76,13 +76,40 @@
         "action_name": "Delete"
       },
       "location_accuracy": "Location Accuracy",
-      "default_title": "Location Information",
+      "default_title": "Temporary Pin",
```
paulschreiber, line 79: issue: this sounds like an internal name, not something user-visible.

dev-client/src/screens/LocationScreens/LocationSoilIdScreen.tsx
```diff
@@ -0,0 +1,52 @@
...
+type Props = {
+  siteId?: string;
+  coords: Coords;
+};
+
+export const LocationSoilIdScreen = ({siteId, coords}: Props) => {
+  const {t} = useTranslation();
+  const site = useSelector(state =>
+    siteId === undefined ? undefined : state.site.sites[siteId],
+  );
```
shrouxm, line 40: suggestion: use `selectSite` from `terraso-client-shared/selectors`. (this isn't done consistently throughout the codebase but i'm hoping to at least be good about this with new code)

dev-client/src/translations/en.json
shrouxm: (comment: i continue to feel that the right place to give feedback on copy is in the Figma, not in PRs, since it's unrelated to implementation)

dev-client/src/screens/LocationScreens/components/CreateSiteButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const CreateSiteButton = ({coords}: Props) => {
+  const {t} = useTranslation();
+  const navigation = useNavigation();
+
+  const onCreate = useCallback(() => {
+    navigation.navigate('CREATE_SITE', {coords});
+  }, [navigation, coords]);
+
+  return (
+    <Box paddingVertical={'20px'}>
```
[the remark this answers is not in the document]
knipec: I did it this way because it maintained the spacing that existed before (which was `mt={5}` for this Box and for the element after it) and there was no string defined under `theme.space` that corresponded to 20px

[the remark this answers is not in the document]
paulschreiber: Do you want to add an `ml` to themes? Or as Courtney if we can use 16 or 24px here?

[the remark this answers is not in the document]
knipec: I'll actually take the padding out of the button itself and let the parent figure out the padding. I'll put it at "md" (16px) for now, I'm guessing the designs for that screen aren't pixel perfect as currently it's 21px.

dev-client/src/translations/en.json
knipec: I left a comment in the Figma to this effect, but will move forward with this PR unless I hear otherwise

dev-client/src/components/content/ScreenContentSection.tsx
```diff
@@ -0,0 +1,48 @@
...
+  Box,
+} from 'terraso-mobile-client/components/NativeBaseAdapters';
+
+type Props = {
+  title?: string;
+  backgroundColor?: string;
+  children: ReactNode;
+};
+
+export const ScreenContentSection = ({
```
tm-ruxandra, line 30: very happy to see this :)

knipec: Still don't know how to fully align with the hierarchy in Courtney's mental model. These are all I've got in mind, lmk if you have a preference or a better idea:

**Option 1**: (what exists currently)
- ScreenContentSection with title="Soil ID"
- ScreenContentSection for Soil Matches with grey background
- ScreenContentSection with title="Site Data"

**Option 2**: 
- One-off SoilIdSection component that exists only for conceptual reasons
    - ScreenContentSection with title="Soil ID"
    - ScreenContentSection for Soil Matches with grey background
- ScreenContentSection with title="Site Data"

dev-client/src/screens/LocationScreens/LocationSoilIdScreen.tsx
knipec: Hmm if I do 
```const site = siteId === undefined ? undefined : useSelector(selectSite(siteId));``` 
the linter says
```error  React Hook "useSelector" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks```

Was that what were recommending? If so, do you have thoughts on what to do here?

dev-client/src/components/content/ScreenContentSection.tsx
knipec: Ok chatted with Garo and we concluded: Just do Option 1 for now, but I created tech-debt item #1288 to track this

tm-ruxandra: Option 3 sounds ideal to me but it's fine with me to do that as a followup task.

tm-ruxandra (approved):

Issue #1275, opened 2024-05-09
Site location overlay - change size, remove element


DerekCaelin:
Feedback based on courtney's review of [select location on map](https://docs.google.com/spreadsheets/d/1GbEEPST69IHwD_ALwFQssZZRZMJjc9V8SX4g44p2pxs/edit#gid=2004954072)

- [x]  "Last Modified" should get removed. 
- [x] Project Name should be Body 1 style with Primary Text color.

DerekCaelin:
seems done in 127

PR #1285, opened 2024-05-09
fix: update site overlay card to match design

paulschreiber:
## Description
update site overlay card to match design
- remove last modified
- fix font size and spacing

### Related Issues
Addresses #1275

### DIFF
--- a/dev-client/src/components/SiteCard.tsx
+++ b/dev-client/src/components/SiteCard.tsx
@@ -24,7 +24,6 @@ import {IconButton} from 'terraso-mobile-client/components/icons/IconButton';
 import {StaticMapView} from 'terraso-mobile-client/components/StaticMapView';
 import {Card} from 'terraso-mobile-client/components/Card';
 import {
-  Box,
   Row,
   Heading,
   Text,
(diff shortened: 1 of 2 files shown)


dev-client/src/components/SiteCard.tsx
```diff
@@ -71,15 +66,8 @@ export const SiteCard = ({
       onPress={onCardPress}
       buttons={buttons}
       isPopover={isPopover}>
-      {project && <Heading size="md">{project.name}</Heading>}
-      <Text variant="subtitle2" color="text.secondary">
-        {t('site.last_updated', {
-          date: TEMP_MODIFIED_DATE,
-          name: TEMP_MODIFIED_NAME,
```
knipec, line 78: Issue: I noticed when trying out this change, site cards on the map are now choosing their width based on the length of the title: 
image

I imagine this was a problem before, but because we had the extra date and name text it wouldn't be as noticeable. Not immediately sure what to recommend though.

paulschreiber: Fixed:
image

knipec (commented):
Issue: Looks like this change also moved the position of the "location-on" IconButton:
Screenshot 2024-05-13 at 10 54 06 AM

knipec:
Issue: Looks like now the PeopleBadge is too far to the right 
image

knipec (changes requested):
Issue: Looks like now the PeopleBadge is too far to the right 
image

paulschreiber:
@knipec The badge should be fixed now.

knipec (approved):

PR #1315, opened 2024-05-13
fix: update site creation screen

paulschreiber:
## Description
Update site creation screen:
- separate out altitude/longitude
- hide project menu if no projects present
- update field order to match designs

### Related Issues
Fixes #....

### DIFF
--- a/dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
+++ b/dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
@@ -40,24 +40,13 @@ import {HelpTooltipButton} from 'terraso-mobile-client/components/tooltips/HelpT
 import {siteValidationSchema} from 'terraso-mobile-client/schemas/siteValidationSchema';
 import {useSelector} from 'terraso-mobile-client/store';
 
-export type FormState = InferType<ReturnType<typeof siteValidationSchema>> & {
-  coords: string;
-};
-
-// export type FormState = Omit<
-//   InferType<ReturnType<typeof siteValidationSchema>>,
-//   'coords'
-// > & {
-//   coords: string;
-// };
+export type FormState = InferType<ReturnType<typeof siteValidationSchema>>;
 
 export const CreateSiteForm = ({
   isSubmitting,
   handleSubmit,
-  // handleChange,
   setValues,
   values,
-  // sitePin,
   onInfoPress,
   isValid,
 }: FormikProps<FormState> & {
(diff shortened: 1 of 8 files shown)


dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
```diff
@@ -90,40 +72,48 @@ export const CreateSiteForm = ({
...
-              setProjectId={projectId =>
-                setValues(current => ({
-                  ...current,
-                  projectId: projectId ?? undefined,
-                }))
-              }
+          <FormField name="longitude">
+            <FormInput
+              value={values.longitude.toString()}
+              keyboardType="decimal-pad"
```
knipec, line 93: Issue: Can we keep some label text saying which is latitude vs. longitude?

dev-client/src/translations/en.json
```diff
@@ -56,15 +56,10 @@
...
-      "coords_parse_error": {
-        "COORDS_PARSE": "Coordinates must be comma-separated decimal numbers representing a latitude, longitude pair",
-        "LATITUDE_PARSE": "Latitude must be a decimal number",
-        "LATITUDE_MIN": "Latitude must be at least {{LATITUDE_MIN}}",
-        "LATITUDE_MAX": "Latitude must be at most {{LATITUDE_MAX}}",
-        "LONGITUDE_PARSE": "Longitude must be a decimal number",
-        "LONGITUDE_MIN": "Longitude must be at least {{LONGITUDE_MIN}}",
-        "LONGITUDE_MAX": "Longitude must be at most {{LONGITUDE_MAX}}"
-      }
+      "min_latitude_error": "Latitude must be at least {{LATITUDE_MIN}}",
```
knipec, line 59: Issue: You need to either change `LATITUDE_MIN` to `min` here to match the field specified in siteValidationSchema.ts, or vice versa. 

Same with max, and with longitude min/max.

Especially since we don't have automated tests right now, please make sure to manually confirm all added functionality is working as expected as much as possible.

dev-client/src/screens/CreateSiteScreen/components/CreateSiteForm.tsx
paulschreiber: Yes — that will get addressed with the new text field component.

knipec: Are we guaranteed to implement the new text field component within Capri? Just want to be conscious that we're not regressing functionality in expectation of something that we won't end up shipping.

paulschreiber: @knipec Yes. It's in review now.

knipec (approved):

Issue #1230, opened 2024-05-01
Soil ID (Site) - Soil Properties Module - table version - no data


DerekCaelin:
Provide a blank, table-based version of the soil properties module. 

## References
MVP-izes this isse: https://github.com/techmatters/terraso-mobile-client/issues/1201

## Design
[Figma Playground](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=6856%3A33767&mode=design&t=wi4I6AH7WinjE5Hn-1)

## Acceptance Criteria
- [x] Soil Properties header exists
- [x] Button to “Add soil data” navigates to the site’s “Soil” tab
- [x] Button does not appear for Viewer role
- [x] I see empty rows for 4 columns: Depth (cm), Texture, Color, % Rock Fragments, for each expected depth
- [x] Table can scroll horizontally

knipec:
QA note: May be worth playing around with the site navigation behavior now that the "Add soil data" button is introduced, since it pushes a new sheet onto the navigation stack. For example:
- Site tab > press Explore (goes to Soil ID) > press Button (goes to Soil tab) > go to Site tab > press Explore (screen gets pushed from the left, indicating we're going back to a previous screen on the navigation stack)

I think this is reasonable, and wouldn't suggest changes. Just wanted to note it.

DerekCaelin:
This passed my functional review. @CourtneyLee333 over to you for QA.

Recommend looking at a site called "Airbnb headquarters" to see an empty site in a project with depth requirements as a viewer.

CourtneyLee333:
This passes my functional review as well. My two design concerns:

- The table cells are not wide enough. This shows in the "no data" context as the 100-200 depth wrapping to a second line. 
- When "None Specified" is chosen for a project, should the empty table default to at least one row until a depth is created? Currently, if there is no data, the Soil Properties table shows as just a one pixel line. This could be confusing.

CourtneyLee333:
I have one nitpick here: the table cells need to be wider. In the blank

knipec:
1. Width: I believe this was done by Ruxandra's [PR 1475](https://github.com/techmatters/terraso-mobile-client/pull/1475/files) -- though the emulator I'm using didn't show it wrapping, and I'm not sure how to guarantee that it will work on every possible device :\
2. "None Specified": Done!
[...]

PR #1324, opened 2024-05-14
feat: Add "Soil Properties" content with table using bogus data + unfinished "Add soil data" button

knipec:
## Description
Add soil properties table (with bogus data) 
Add "Add soil data >" button (but does not yet navigate to the correct tab)

### Checklist
- [ ] Future PR: Make button navigate to the correct tab
- [ ] Future PR: Check in with Courtney about the exact styling

### Related Issues
Partially implements #1230 
Would like reviews on this part now, to unblock Ruxandra's work that uses the table

### Verification steps
Look at the bottom of the Soil ID sheet for a site

Currently looks like this: 
(alas, I was trying so hard to get the table narrow enough to not need horizontal scrolling, but still fit the strings nicely *shakes fist at Android*)
Screenshot 2024-05-14 at 4 20 55 PM

### DIFF
--- a/dev-client/src/components/SoilPropertiesDataTable.tsx
+++ b/dev-client/src/components/SoilPropertiesDataTable.tsx
@@ -30,7 +30,7 @@ type DataTableHeaderProps = {
 const DataTableHeader = ({width, text}: DataTableHeaderProps) => {
   return (
     <Text
-      variant="table-header"
+      variant="body1-medium"
       width={width}
       paddingHorizontal="sm"
       pb="sm"
@@ -41,10 +41,10 @@ const DataTableHeader = ({width, text}: DataTableHeaderProps) => {
 };
 
 type DataTableCellProps = {
-  text: string;
   width: NBDimensionValue;
+  text: string;
 };
-const DataTableCell = ({text, width}: DataTableCellProps) => {
+const DataTableCell = ({width, text}: DataTableCellProps) => {
   return (
     <Box
       flex={1}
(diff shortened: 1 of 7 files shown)


dev-client/src/components/SoilPropertiesDataTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+      paddingHorizontal={'8px'}
+      paddingVertical={'4px'}>
+      <Text>{text}</Text>
+    </Box>
+  );
+};
+
+type Props = {
+  rows: [string, string, string, string][];
+  // rows: string[][4];
```
knipec, line 64: Oops I will remove this line

dev-client/src/components/SoilPropertiesDataTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+        />
+        <DataTableHeader
+          width={columnWidthRockFragment}
+          text={t('site.soil_id.site_data.soil_properties.rock_fragment')}
+        />
+      </Row>
+
+      <Box borderTopWidth={'1px'} borderLeftWidth={'1px'}>
+        {rows.map((row: (typeof rows)[number], i: number) => (
+          <Row justifyContent="flex-start" key={i}>
```
paulschreiber, line 98: question: do we need the key? I haven't seen it elsewhere in react native code.

paulschreiber (approved):

dev-client/src/components/SoilPropertiesDataTable.tsx
knipec: Oo as a result of this comment, I just learned that while you should use a key on mapped components, it is considered an anti-pattern to use the index alone as the key.
(https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318)

paulschreiber: Yeah, we do that a bunch in regular React. I hadn't seen it elsewhere in the bo app, which is why I asked. Do we get the warning without it?

knipec: Yeah, I tried removing the key and the warning does show up.

PR #1325, opened 2024-05-15
fix: add react/jsx-curly-brace-presence rule to .eslintrc, run linter to enforce style

paulschreiber:
## Description
Add react/jsx-curly-brace-presence rule to .eslintrc, run linter to enforce style.

We were previously inconsistent in our use.


knipec (approved):
Yay!

Issue #1158, opened 2024-04-19
Build Location Soil Info page - Integrate Soil Properties component


DerekCaelin:
As a Manager or a Contributor, I want see the soil properties of a given soil, so I know more about the characteristics of a soil that may be at this location. 

Build a module for SoilID Info page that adapts the selected location soil ID's soil data properties using the standard soil properties chart/visualization components, so that a user can learn more about a potential soil for a location. See parent task for more details.

## How to find the screen
1. From home screen, select a location on a map
2. In the resulting popover, select "Learn more" 
3. Select "Explore the data" on the soilID Tile
4. Select a soil from the ranked soil list

## References
[SoilID Help Text](https://docs.google.com/document/d/1qRvfXfjKkoKAXl3jGeDwlRMXw3rYbvJc5ciY06Ehbqw/edit)

## Design
- [Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-15906&mode=design&t=6V7vK4lZXKKsrz3b-0)

### Dependencies
- https://github.com/techmatters/terraso-product/issues/653

## Acceptance Criteria
- [x] As a Manager or a Contributor, I want see the soil properties of a given soil, so I know more about the characteristics of a soil that may be at this location. 
  - [x] Texture
  - [ ] Color
  - [ ] Rock Fragmentation

## Additional Context
- At launch, display blank values for the site's pH and Electrical Conductivity, since we won't be tracking that.

DerekCaelin:
@tm-ruxandra this is in QA, but I assume I should hold off on reviewing until https://github.com/techmatters/terraso-product/issues/653 is implemented, is that right?

tm-ruxandra:
@DerekCaelin Technically you can review it now / whenever the next build goes out, but it will only be displaying dummy data.

DerekCaelin:
This passes "functional" review in that the page exists and has dummy content displayed correctly, and that different elements appear depending on whether the location is temporary or not. I expect we'll look again once we hook it up to soil ID data with [653](https://github.com/techmatters/terraso-product/issues/653). Over to @CourtneyLee333 for her review

CourtneyLee333:
Woohoo! Looking good! 

Some design observations:
- LocationScoreInfoSheet and PropertiesScoreInfoSheet cover the app bar. They should, instead, open to the proper height. 
[...]

PR #1365, opened 2024-05-17
feat: integrate soil id info sheets with data table

tm-ruxandra:
## Description

Add soil data table to temp location and site soil info sheets. Refactor some associated components and add basic data plumbing to make future integration work on soil ID easier.

### Checklist

### Related Issues
Fixes [#....](https://github.com/techmatters/terraso-mobile-client/issues/1158)

### Verification steps

Open temp location pin and navigate to info sheet. Verify that table displays as expected. Open a site and navigate to the info sheet. Verify that the table displays as expected there as well.

### DIFF
--- a/dev-client/src/components/SoilPropertiesDataTable.tsx
+++ b/dev-client/src/components/SoilPropertiesDataTable.tsx
@@ -15,6 +15,7 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
+import {ScrollView} from 'native-base';
 import {useTranslation} from 'react-i18next';
 import {
   Box,
@@ -64,7 +65,7 @@ type Props = {
   rows: SoilPropertiesDataTableRow[];
 } & React.ComponentProps<typeof Box>;
 
-export const SoilPropertiesDataTable = ({rows, ...containerProps}: Props) => {
+export const SoilPropertiesDataTable = ({rows}: Props) => {
   const {t} = useTranslation();
 
   const columnWidthDepth: NBDimensionValue = '85px';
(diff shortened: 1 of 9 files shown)


dev-client/src/components/SoilPropertiesDataTable.tsx
```diff
@@ -84,38 +85,40 @@ export const SoilPropertiesDataTable = ({rows, ...containerProps}: Props) => {
...
-        <DataTableHeader
-          width={columnWidthColor}
-          text={t('site.soil_id.site_data.soil_properties.color')}
-        />
-        <DataTableHeader
-          width={columnWidthRockFragment}
-          text={t('site.soil_id.site_data.soil_properties.rock_fragment')}
-        />
-      </Row>
+    <ScrollView horizontal={true}>
```
tm-ruxandra, line 88: After discussing with @knipec we agreed to pull the `ScrollView` into the table component, since that was its only use-case in practice. We also agreed to factor out the `containerProps` property because it was not used anywhere.

dev-client/src/model/soilId/soilIdPlaceholders.ts
```diff
@@ -97,3 +105,5 @@ export const DATA_BASED_SOIL_MATCH: DataBasedSoilMatch = {
   dataMatch: SOIL_MATCH_INFO,
   combinedMatch: SOIL_MATCH_INFO,
 };
+
+export const SOIL_DATA: any = {};
```
tm-ruxandra, line 109: This is a placeholder to be replaced with actual API integration in [#653](https://github.com/techmatters/terraso-product/issues/653).

dev-client/src/screens/LocationScreens/components/SoilIdSections.tsx
```diff
@@ -76,6 +78,7 @@ export const SoilIdMatchesSection = ({siteId, coords}: SoilIdSectionProps) => {
           <SiteScoreInfoContent
             locationMatch={locationMatch}
             dataMatch={dataMatch}
+            soilData={soilData}
```
paulschreiber, line 81: question: since each of these constants are only used once, does it make sense to pass the imported values directly?

tm-ruxandra: This is placeholder code; the actual data will be passed in this way since it needs to be loaded fromt he server.

paulschreiber (approved):

Issue #1200, opened 2024-04-29
Soil ID (Site) - Implement "Land Slope" module


DerekCaelin:
Implement the "Land Slope" module on the Soil ID screen for Site locations. See parent issue for context and related issues. 

## Design
[Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-14752&mode=design&t=bn4bR0g90ayypnJZ-0)
<img width="311" alt="image" src="https://github.com/techmatters/terraso-mobile-client/assets/10799815/0b5488a2-b664-43b6-aa7c-34c5006b5115">
<img width="311" alt="image" src="https://github.com/techmatters/terraso-mobile-client/assets/10799815/ddf22166-586a-4b11-8386-052ab8e9e887">

## Acceptance Criteria
- [x] Land Slope text correctly reflects my site data
    - [x] If the user selected an image for Slope Steepness, show the image in the box as well as its accompanying text.
    - [x] If the user did manual entry or slope meter, keep the box empty and just display the text.
    - [x] The text should display the entry, whether a picture is chosen or not
    - [x] Image is an empty square if there is no data
- [x] Button to “Add slope” navigates to the site’s “Slope” input tab. Button is only visible to Managers and Contributors.

DerekCaelin:
This passes my functional review! @CourtneyLee333 please take a look. I reviewed:

## Acceptance Criteria
- [ ] Land Slope text correctly reflects my site data
    - [ ] If the user selected an image for Slope Steepness, show the image in the box as well as its accompanying text.
    - [ ] If the user did manual entry or slope meter, keep the box empty and just display the text.
    - [ ] The text should display the entry, whether a picture is chosen or not
    - [ ] Image is an empty square if there is no data
- [ ] Button to “Add slope” navigates to the site’s “Slope” input tab. Button is only visible to Managers and Contributors.

CourtneyLee333:
Woohoo! Fast and smoth. Some design comments:

- It is weird to have the box empty and display the degree beside it. Would it be a heavy lift to display the number inside the box if an image has not been chosen? Or to hid the box and put the number in its place?
- I think the button should lead to the Slope Steepness screen. And perhaps this means the heading on the Site Dashboard should be Slope Steepness as well, since it is steepness we are specifically referencing and what the algorithm needs.

CourtneyLee333:
[...]

PR #1389, opened 2024-05-21
Feat: Add slope data section to Soil ID sheet

knipec:
## Description
- On the site-specific soil id sheet: 
    - Display slope steepness image and text based on site data
    - Display "Add slope" button that navigates to the "Slope" tab for the site (button is hidden for viewers)
- Refactor: Extracted and moved some files -- FYI the diff tool isn't clear about this, so lmk if you want a walkthrough

### Checklist

### Related Issues
Implements #1200 

### Verification steps
Go to Soil ID page and view the "Slope" section


dev-client/src/screens/SlopeScreen/utils/renderValues.ts
```diff
@@ -23,13 +23,38 @@ export const renderSteepness = (
...
+) =>
+  slopeSteepnessSelect
+    ? t(`slope.steepness.select_labels.${slopeSteepnessSelect}`) +
+      '\n' +
+      t(`slope.steepness.select_labels.${slopeSteepnessSelect}_PERCENT`)
+    : typeof slopeSteepnessPercent === 'number'
+      ? `${slopeSteepnessPercent.toFixed(0)}%`
+      : typeof slopeSteepnessDegree === 'number'
+        ? `${slopeSteepnessDegree}°`
+        : undefined;
```
paulschreiber, line 45: question: why are we using a bare newline?
issue: the percentage and degree strings should be localized.
issue: multi-level nested ternaries are too hard to read/follow. please pull this code into a function.

dev-client/src/screens/SlopeScreen/utils/renderValues.ts
```diff
@@ -23,13 +23,38 @@ export const renderSteepness = (
...
+
+export const renderSlopeSteepnessSelectInline = (
+  t: TFunction,
+  slopeSteepnessSelect: string,
+) => {
+  return (
+    t(`slope.steepness.select_labels.${slopeSteepnessSelect}`) +
+    ' ' +
+    t(`slope.steepness.select_labels.${slopeSteepnessSelect}_PERCENT`)
+  );
```
paulschreiber, line 55: question: what does this look like with real data? Seem likely it should be localized (`{s1} {s2}`).

paulschreiber (changes requested):

dev-client/src/screens/SlopeScreen/utils/renderValues.ts
knipec: 1. There are multiple places that we display the slope data. Most places display it with a space between the name and percentage range (1st picture). But in the new screen we want to display it with the name on one line and the percentage range on a second line (2nd picture).
Is there a preferred way to do that?
    1st picture: Screenshot 2024-05-21 at 4 02 33 PM
    2nd picture: Screenshot 2024-05-21 at 4 02 49 PM

2. Ok, I will plan to add `"percentage": "{{number}}%"` and `"degree": "{{number}}°"` to `en.json`

3. Not sure which code you are suggesting extracting, but I'll plan to just use if / else if / else here for readability.

dev-client/src/screens/SlopeScreen/utils/renderValues.ts
knipec: It will look like how we display it on the slope steepness screen already: "Moderately Steep 50-60%"
    Screenshot 2024-05-21 at 4 02 33 PM

See the changes in `en.json` and lmk if it's still not clear.

paulschreiber: Please add something to en.json along similar to "{steepness_description} {steepness_amount}" with the space in it. We want to avoid string concatenation like this in the code. (It will break for RTL languages.)

dev-client/src/screens/SlopeScreen/utils/renderValues.ts
paulschreiber: Here are two ways to avoid the newline in the code:
• control the wrapping via the container/contain width
• use separate `<Text>` components for the steepness description and steepness percent

tm-ruxandra: FYI, we already have a standard means of doing locale-friendly percentage formatting in `util.ts:formatPercent`.

knipec: @paulschreiber Would it be fine to add `"{{steepness_name}}\n{{steepness_percent}}"` to en.json (since there are many other strings with newlines in there)?

paulschreiber: It'd be better not to do that. I believe the other places where we have newlines are for static text with paragraph breaks. These are two distinct entities.

tm-ruxandra (approved):

paulschreiber (approved):

PR #1391, opened 2024-05-21
feat: data transformations for soil properties tables

tm-ruxandra:
## Description

Intermediate work expanding the `SoilPropertiesDataTable` to allow feeding it input from both Soil ID and collected site soil data. Formalizes the row data type to use standard enum values where possible and refactors the table itself to use translation table for the display of rows. Additionally refactors table components to share a common `components` subfolder.


knipec (approved):
Your code is so readable, thank you 🤩

Issue #1231, opened 2024-05-01
Soil ID (Site) - Soil properties module - table version - color, rock framentation, texture


DerekCaelin:
Update soil properties table that was implemented in https://github.com/techmatters/terraso-mobile-client/issues/1230 with data. In the Site context, show user-collected data. 

## Designs
TBD

## Acceptance Criteria
- [x] (Site context) If I have collected data, I see real data reflected in the Texture visualization
- [x] (Site context) If I have collected data, I see real data reflected in the Color visualization
- [x] (Site context) If I have collected data, I see real data reflected in the Rock Fragmentation visualization

## Notes
- For color: there is [a function for that](https://github.com/techmatters/terraso-mobile-client/blob/ada733f533d008d1b02738aaa26cba949556be63/dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts#L201-L215) and [a longer explanation](https://github.com/techmatters/terraso-mobile-client/blob/ada733f533d008d1b02738aaa26cba949556be63/dev-client/src/screens/SoilScreen/ColorScreen/utils/munsellConversions.ts#L18-L42) in the file munsellConversions.ts

DerekCaelin:
This passes my review. @CourtneyLee333 you can use site `PMC 1` as an example of what a fully filled out data table looks like

CourtneyLee333:
This passes my review, with the comments already made about the cells of the table needing to be wider to accommodate without wrapping:
- 100-200 depth label
- Sandy Loam, Loamy Sand, etc.. texture (a 3-word texture is fine to wrap)
- Full Munsell value

DerekCaelin:
wrapping (and other qualities @CourtneyLee333 highlighted) are addressed so I'm going to call this done. Thank you!

PR #1404, opened 2024-05-24
feat: Use actual site data for the soil properties table

knipec:
## Description
- Create table rows with real soilData collected for the site
- Use `SoilData.depthIntervals` (which has one per depth interval, regardless of if there are data collected for that interval) to make table rows, instead of `SoilData.depthDependentData` (which excludes depth intervals for which there are no data). 

### Related Issues
Implements #1231

### DIFF
--- a/dev-client/src/components/tables/soilProperties/SoilPropertiesData.tsx
+++ b/dev-client/src/components/tables/soilProperties/SoilPropertiesData.tsx
@@ -22,6 +22,7 @@ import {
   SoilIdDepthDependentSoilDataTextureChoices,
   SoilIdSoilData,
 } from 'terraso-client-shared/graphqlSchema/graphql';
+import {AggregatedInterval} from 'terraso-client-shared/selectors';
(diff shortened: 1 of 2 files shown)


dev-client/src/components/tables/soilProperties/SoilPropertiesData.tsx
```diff
@@ -43,24 +42,26 @@ export type SoilPropertiesDataTableRow = {
...
+    });
   return {
-    depth: {
-      start: data.depthInterval.start,
-      end: data.depthInterval.end,
-    },
-    texture: data.texture ?? undefined,
-    rockFragment: data.rockFragmentVolume ?? undefined,
+    depth: interval.depthInterval,
+    texture: dataForInterval?.texture ?? undefined,
```
paulschreiber, line 63: question: if you don't have the `?? undefined`, do you get some other default value?

paulschreiber (approved):

Issue #861, opened 2024-02-12
Deleting a site takes user to blank screen


ltseng:
Just like the issue title states. Deleting a site from the site settings screen takes me to a blank white screen, and pressing the Android device's back button takes me all the way back to my device's home page. Observed in Build 79.

Ideally, deleting a site would take the user to the home screen map. 

https://github.com/techmatters/terraso-mobile-client/assets/2730009/0ffcd6fc-cdbd-477d-ab69-cfb8aff31dab

ltseng:
Laura has run into this bug during their testing of Build 70 as well.

ltseng:
This also occurs when deleting a site that's part of a project.

DerekCaelin:
Reviewed bugs. Still an issue as of 1.10.11.

paulschreiber:
This is a crash (`TypeError: Cannot read property 'name' of undefined`)
Sentry details: https://techmatters.sentry.io/issues/5420620139/

Backtrace:
```
This error is located at:
    in SiteSettingsScreen (created by SceneView)
    in StaticContainer
    in EnsureSingleNavigator (created by SceneView)
    in SceneView (created by SceneView)
```

Problem code:
```js
  const [name, setName] = useState(site.name);
```

DerekCaelin:
this is resolved in 135!

PR #1431, opened 2024-05-28
fix: don't crash after deleting site

paulschreiber:
## Description
Don't crash when deleting a site
- don't navigate to site dashboard or project screen via pop()
- avoid null deference on site settings screen

### Related Issues
Addresses #861.

### DIFF
--- a/dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
+++ b/dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
@@ -46,6 +46,7 @@ export const SiteSettingsScreen = ({siteId}: Props) => {
   const {t} = useTranslation();
   const navigation = useNavigation();
   const site = useSelector(state => state.site.sites[siteId]);
+  const [name, setName] = useState(site?.name);
 
   const onSave = useCallback(
     () => dispatch(updateSite({id: site.id, name})),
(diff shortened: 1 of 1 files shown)


dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
```diff
@@ -61,53 +55,17 @@ export const SiteSettingsScreen = ({siteId}: Props) => {
 
   const onDelete = useCallback(async () => {
     await dispatch(deleteSite(site));
-    navigation.pop();
+    navigation.navigate('BOTTOM_TABS');
   }, [dispatch, navigation, site]);
 
+  if (!site) {
+    return null;
```
knipec, line 62: Question: Since I'll be making a thing in the backlog to follow up on this, could you confirm if this understanding is correct?
1. When this happens, what will we render? --> Nothing for the SiteSettingsScreen, but that's fine because we will then immediately navigate to 'BOTTOM_TABS'
2. When does this happen? --> Every time we delete a site, because for not-yet-understood reasons it tries to re-render SiteSettingsScreen before navigating to 'BOTTOM_TABS'

knipec (approved):

dev-client/src/screens/SiteSettingsScreen/SiteSettingsScreen.tsx
paulschreiber: Thats my understanding.

PR #1432, opened 2024-05-28
fix: update string to match implementation (allow single email address)

paulschreiber:
## Description
On the projects > add team member screen, update the string to match implementation (allow single email address, not multiple).

### Related Issues
Request to support multiple email addresses: #731

### DIFF
--- a/dev-client/src/translations/en.json
+++ b/dev-client/src/translations/en.json
@@ -308,7 +308,7 @@
     },
     "add_user": {
       "heading": "Add Team Members",
-      "help_text": "Enter single or multiple Terraso email addresses in the field below.",
+      "help_text": "Enter a Terraso email address",
       "user_does_not_exist": "User {{email}} is not a Terraso user, unable to add to project.",
       "user_in_project": "User {{email}} is already a member of this project.",
       "already_added": "User {{email}} already added to user list.",


knipec (approved):
The change looks good, ship it! 

Question: But the #731 linked in the description seems like the wrong thing. Is there a different issue it should point to?

(It is not clear to me that the user has to press enter after putting an email address into the text field, and only after doing that does the Save button become enabled. I wanted to see if the associated issue mentions that or not.)

paulschreiber:
@knipec We don't have a new issue for the "fix the text" problem. We also don't have anything tracking the UX being confusing. @CourtneyLee333 please make a note of this.

knipec:
@paulschreiber Is there an old issue for whatever other problem it was? The item linked under "Related Issues" in the description is a dependabot pull request.

CourtneyLee333:
@knipec What do you expect to happen upon entering an email address into the field, if not to hit the enter button?

knipec:
@CourtneyLee333 I don't know best practices for mobile, so happy to be wrong on this, but maybe an "Add" button, or at least something that tells me I need to press enter? I think I was a little surprised that the "Save" button didn't become enabled when I entered an email in, and I didn't realize until trying the Enter button that there was the 2nd step of "choose permissions for this email" (though now I'm like, oh of course it makes sense that it's a 2-step process)

PR #1435, opened 2024-05-29
fix: don’t sent client_id for google/ms SSO

paulschreiber:
## Description
Don’t sent client_id for google/ms SSO. Re-enables login for non-Apple providers.

### DIFF
--- a/dev-client/src/auth/index.ts
+++ b/dev-client/src/auth/index.ts
@@ -84,7 +84,7 @@ async function exchangeToken(
     provider,
     jwt: identityJwt,
   };
-  if (provider === 'apple') {
+  if (provider === 'apple' && Platform.OS === 'ios') {
     body.client_id = Constants.expoConfig!.ios?.bundleIdentifier;
   }
   const payload = await request<AuthTokens>({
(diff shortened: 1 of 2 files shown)


dev-client/src/auth/index.ts
```diff
@@ -80,13 +80,16 @@ async function exchangeToken(
   identityJwt: string,
   provider: Omit<AuthProvider, 'google'> | `google-${'ios' | 'android'}`,
 ) {
+  let body: any = {
+    provider,
+    jwt: identityJwt,
+  };
+  if (provider === 'apple') {
```
knipec, line 87: Question: Do you also need to check if the platform is ios?

```suggestion
  if (provider === 'apple' && Platform.OS === 'ios') {
```

dev-client/src/auth/index.ts
```diff
@@ -80,13 +80,16 @@ async function exchangeToken(
...
+    body.client_id = Constants.expoConfig!.ios?.bundleIdentifier;
```
knipec, line 88: Suggestion: Could this use `apple.clientId` from line 46 above?
```suggestion
    body.client_id = configs.apple.clientId;
```

paulschreiber: @knipec what about `body.client_id = APP_CONFIG.appleClientId;`?

knipec (approved):

Issue #1313, opened 2024-05-13
(iOS) Only the down arrow part of "add to project" is pressable, when the full dropdown should be 


DerekCaelin:
### App Version

123

### Platforms

- [X] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

When creating a site and adding it to a project, the entire field for Add to Project should be pressable. Currently only the down arrow part is.

### Steps to Reproduce

1. Go to "Sites" screen
2. tap a point on the map
3. select "create site"
4. see issue

### Expected behavior

"Add to project" dropdown will expand when any part of the dropdown is pressed

### Actual behavior

Only the down arrow will expand the dropdown.

DerekCaelin:
This is addressed in 142

PR #1443, opened 2024-05-29
fix: make entire select box clickable

paulschreiber:
## Description
Make entire select box clickable on IOS.

### Related Issues
Addresses #1313.

### DIFF
--- a/dev-client/__tests__/integration/__snapshots__/CreateSiteScreen-test.tsx.snap
+++ b/dev-client/__tests__/integration/__snapshots__/CreateSiteScreen-test.tsx.snap
@@ -725,231 +725,37 @@ exports[`renders correctly 1`] = `
                                     style={
                                       [
                                         {
-                                          "backgroundColor": "#0000000F",
-                                          "borderTopLeftRadius": 4,
-                                          "borderTopRightRadius": 4,
+                                          "flexDirection": "row",
                                         },
-                                        {},
+                                        [
+                                          {},
+                                          {
+                                            "backgroundColor": "#0000000F",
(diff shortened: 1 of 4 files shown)


shrouxm (changes requested):
i believe the intention is for the label to still be visible when the dropdown is selected, which this code doesn't seem to do?

knipec:
Re: Garo's comment 
> i believe the intention is for the label to still be visible when the dropdown is selected, which this code doesn't seem to do?

Could be worth a quick checkin with Courtney if it's desired to move the label, since there's already a title "Add to Project" right above it, so it seems kinda redundant.

dev-client/__tests__/integration/__snapshots__/CreateSiteScreen-test.tsx.snap
knipec, line 1: Question: Do you understand why there are a few levels of `<View>` that are no longer in the snapshot? Is it because `<TextInput>` is implemented with nested `<View>`s whereas `<Row>` is not?

knipec (approved):
Approving assuming the answer to my question is "yup", and Courtney is good with the label stuff :)

paulschreiber:
@knipec The label and placeholder placement was discussed with/designed by Courtney.

dev-client/__tests__/integration/__snapshots__/CreateSiteScreen-test.tsx.snap
paulschreiber: No idea how these components are implemented behind the scenes.

PR #1446, opened 2024-05-30
feat: Site dashboard & Soil info sheet displays info resulting from Soil ID API

knipec:
## Description
(Creating from Ruxandra's work)
Use Soil ID API to get soil id matches on the site location screen, and access them within a given site via redux to display them on soil id screen and soil id soil info screen

### Related Issues
Implements [#653](https://github.com/techmatters/terraso-product/issues/653)

### Verification steps
Look at the soil ID info screen for a given soil (need the corresponding client-shared work)


dev-client/package.json
```diff
@@ -61,7 +61,7 @@
     "react-native-screens": "^3.31.1",
     "react-native-svg": "^15.3.0",
     "react-native-tab-view": "^3.5.2",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#7841647",
+    "terraso-client-shared": "github:techmatters/terraso-client-shared#f2b9538c634cc77527c4a385716fe5bb17c513ce",
```
knipec, line 64: TODO: Will need to update this before merging!

paulschreiber (commented):
LGTM, but others should review/approve.

dev-client/src/screens/LocationScreens/LocationDashboardContent.tsx
```diff
@@ -140,6 +139,9 @@ export const LocationDashboardContent = ({
       : state.project.projects[site.projectId],
   );
 
+  const soilIdData = useSelector(selectSoilIdData());
```
knipec, line 142: Question: I believe this should be fine as-is, but do you think we should get in the habit of preferring `useSoilIdData` instead of some places using the selector directly? Just feels like doing this requires more thought about when we'd need to request potentially new data?

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -19,45 +19,28 @@ import {useTranslation} from 'react-i18next';
...
+  siteId,
+  coords,
+}: SoilIdMatchesSectionProps) => {
   const {t} = useTranslation();
   const isSite = !!siteId;
-
-  const locationMatch = LOCATION_BASED_SOIL_MATCH;
-  const soilData = SOIL_DATA;
-  const dataMatch = DATA_BASED_SOIL_MATCH;
+  const soilIdData = useSelector(selectSoilIdData());
```
knipec, line 43: Same question as in `LocationDashboardContent` about preferring `useSoilIdData`

knipec (commented):
Lol finally reviewing the thing I technically put out for PR

dev-client/src/model/soilId/soilIdRanking.ts
```diff
@@ -0,0 +1,45 @@
...
+  results: SoilIdResults,
+): LocationBasedSoilMatch | DataBasedSoilMatch | undefined => {
+  const locationBased = results.locationBasedMatches;
+  const dataBased = results.dataBasedMatches;
+
+  if (dataBased.length > 0) {
+    return dataBased.reduce(
+      (prev, current) =>
+        current.combinedMatch.rank > prev.combinedMatch.rank ? current : prev,
+      dataBased[0],
```
knipec, line 34: Issue: This is backwards. The item with the smallest rank should be returned (it might always be the item with rank 0 but I don't know for usre)

dev-client/src/screens/LocationScreens/LocationDashboardContent.tsx
```diff
@@ -228,12 +230,16 @@ export const LocationDashboardContent = ({
...
-          soilName={TEMP_SOIL_ID_VALUE}
-          ecologicalSiteName={TEMP_ECO_SITE_PREDICTION}
-          onExploreDataPress={onExploreDataPress}
-        />
+        {topSoilMatch && (
+          <LocationPrediction
+            label={t('soil.soil_id')}
+            soilName={topSoilMatch.soilInfo.soilSeries.name}
+            ecologicalSiteName={
+              topSoilMatch.soilInfo.ecologicalSite?.name ?? ''
```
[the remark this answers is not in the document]
paulschreiber: @CourtneyLee333 What's the fallback string here?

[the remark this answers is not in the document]
knipec: Also, @paulschreiber or @shrouxm I'm assuming we never expect a time when there is a top soil match but it is missing the name attribute, right?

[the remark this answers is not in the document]
paulschreiber: No, but we should still display something sensible to the user in that case.

dev-client/src/screens/LocationScreens/LocationDashboardContent.tsx
tm-ruxandra: I can give this some more thought - I agree with your point that this potentially requires more reasoning about when data gets loaded.

dev-client/src/model/soilId/soilIdRanking.ts
tm-ruxandra: Good catch!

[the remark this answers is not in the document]
tm-ruxandra: You're right - I was on brain-autopilot and forgot that JS' reduce method does this automatically.

dev-client/src/screens/LocationScreens/LocationDashboardContent.tsx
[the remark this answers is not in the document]
tm-ruxandra: Will discuss with design to get clarification on copy text to show here in the empty case.

dev-client/src/screens/LocationScreens/LocationDashboardContent.tsx
tm-ruxandra: Will refactor to use this.

dev-client/src/screens/LocationScreens/LocationDashboardContent.tsx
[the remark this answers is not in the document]
tm-ruxandra: From Courtney: "None available" is the required copy text here.

dev-client/src/hooks/soilIdHooks.ts
```diff
@@ -0,0 +1,86 @@
...
+   * (We have to keep the useSelector call to be sure that hook usage is
+   * consistent across renders.)
+   */
+  const soilDataSelector = siteId ? selectSoilData(siteId) : () => undefined;
+  const soilData = useSelector(soilDataSelector);
+
+  useEffect(() => {
+    if (
+      siteId !== soilIdInput.siteId ||
+      !isEquivalentCoords(coords, soilIdInput.coords)
```
paulschreiber, line 69: question: should this be `&&` instead of `||`?

paulschreiber (approved):

dev-client/src/hooks/soilIdHooks.ts
tm-ruxandra: No. If either parameter is different from what was most recently loaded, Redux should fetch new matches.

dev-client/__tests__/integration/SlopeScreen-test.tsx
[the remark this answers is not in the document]
tm-ruxandra: It's because the state is part of the SoilID Redux slice, which this snapshot uses because it's touching soil data (which is part of the slice).

PR #1476, opened 2024-06-03
chore: sort imports

paulschreiber:
## Description
Sort imports in ts/tsx files. This matches what we do in the backend and web client.


dev-client/.prettierrc
```diff
@@ -1,7 +1,25 @@
 {
   "arrowParens": "avoid",
+  "singleQuote": true,
+  "semi": true,
```
knipec, line 4: Questions: This means "anyplace that could have a semicolon should have a semicolon", right?

dev-client/.depcheckrc
knipec, line 1: Question: Were the spacing changes here just done manually, or is there a prettier rule that did this?

dev-client/.prettierrc
paulschreiber: Yes. We're already doing that. I added it to the file so the prettier configuration better matches the web client.

dev-client/.depcheckrc
paulschreiber: I did manually sort the array.

knipec (approved):
I'd just check with @tm-ruxandra before submitting that she won't mind the thrash of this change, since she didn't say either way on the slack thread

PR #1477, opened 2024-06-03
fix: save elevation on site creation

paulschreiber:
## Description
Save elevation to database on site creation.

### Related Issues
Addresses https://github.com/techmatters/terraso-product/issues/774
Depends on https://github.com/techmatters/terraso-mobile-client/pull/1315

### DIFF
--- a/dev-client/src/screens/CreateSiteScreen/components/CreateSiteView.tsx
+++ b/dev-client/src/screens/CreateSiteScreen/components/CreateSiteView.tsx
@@ -36,6 +36,7 @@ import {useSelector} from 'terraso-mobile-client/store';
 type Props = {
   defaultProjectId?: string;
   sitePin?: Coords;
+  elevation?: number;
   createSiteCallback: (
     input: SiteAddMutationInput,
   ) => Promise<Site | undefined>;
@@ -46,6 +47,7 @@ export const CreateSiteView = ({
   defaultProjectId,
   createSiteCallback,
   sitePin,
+  elevation,
   onInfoPress,
 }: Props) => {
   const {t} = useTranslation();
@@ -63,13 +65,13 @@ export const CreateSiteView = ({
   const onSave = useCallback(
     async ({...form}: FormState) => {
       const {...site} = validationSchema.cast(form);
-      const createdSite = await createSiteCallback(site);
+      const createdSite = await createSiteCallback({...site, elevation});
       if (createdSite !== undefined) {
         homeScreen?.showSiteOnMap(createdSite);
         navigation.pop();
       }
     },
-    [createSiteCallback, navigation, validationSchema, homeScreen],
+    [createSiteCallback, navigation, validationSchema, homeScreen, elevation],
   );
 
   return (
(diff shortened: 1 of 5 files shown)


dev-client/src/screens/HomeScreen/components/TemporaryLocationCallout.tsx
```diff
@@ -52,16 +52,23 @@ export const TemporaryLocationCallout = ({
   const {t} = useTranslation();
   const navigation = useNavigation();
   const [siteElevationString, setSiteElevationString] = useState('');
+  const [siteElevationValue, setSiteElevationValue] = useState(0);
```
knipec, line 55: Suggestion: Can we remove the siteElevationString state stuff now? Like, just do `useState` for the elevation value, and then derive the elevation string from the value when we need the string? It seems less clean to save both as separate state variables when one is easily derived from the other.

dev-client/src/screens/LocationScreens/components/CreateSiteButton.tsx
```diff
@@ -25,15 +25,15 @@ import {Coords} from 'terraso-client-shared/types';
 import {Icon} from 'terraso-mobile-client/components/icons/Icon';
 import {useNavigation} from 'terraso-mobile-client/navigation/hooks/useNavigation';
 
-type Props = {coords: Coords};
+type Props = {coords: Coords; elevation?: number};
```
knipec, line 28: Question: Why is `elevation` optional -- is there ever a case where we wouldn't have a value for `elevation` here?

knipec (approved):
Looks good to me after addressing the two comments!

dev-client/src/screens/CreateSiteScreen/components/CreateSiteView.tsx
```diff
@@ -66,17 +64,14 @@ export const CreateSiteView = ({
 
   const onSave = useCallback(
     async ({...form}: FormState) => {
-      const {coords, ...site} = validationSchema.cast(form);
-      const createdSite = await createSiteCallback({
-        ...site,
-        ...parseCoords(coords),
-      });
+      const {...site} = validationSchema.cast(form);
+      const createdSite = await createSiteCallback({...site, elevation});
```
knipec, line 68: You said you plan for a future PR to handle the case where the user has changed the lat/long in the form, right?

dev-client/src/screens/LocationScreens/components/CreateSiteButton.tsx
paulschreiber: It's possible the elevation API could fail.

dev-client/src/screens/CreateSiteScreen/components/CreateSiteView.tsx
paulschreiber: Yes.

dev-client/src/screens/HomeScreen/components/TemporaryLocationCallout.tsx
paulschreiber: Good catch.

Issue #1197, opened 2024-04-29
Soil ID - "Top Soil Matches" module


DerekCaelin:
As a user on the Soil ID page, see Top Soil Matches tiles component shows up to 12 soils. See parent task for context and related issues.

## References
[Carissa's implementation notes](https://docs.google.com/document/d/1Jae9g-OpKADlKiviZ_a-DO4Egrdhq3_n__lF1jrMQH4/edit)
Carissa: Nice if it comes after user story #[617](https://github.com/techmatters/terraso-product/issues/617), but not necessary (would just want the data in the same format that would be returned by the API)
The expanded view ("See more matches") is covered by #1198. 

## Designs
[Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=5800-12771&mode=design&t=bn4bR0g90ayypnJZ-0) (Site and Location visual appearance are identical) 
#### Design as of 4/29/2024, please see Figma for most recent version
![image](https://github.com/techmatters/terraso-mobile-client/assets/10799815/279b0802-acf8-47fc-9e30-ab1fa79a4c39)

## Acceptance Criteria
- [x] List displays properly for different numbers of soil matches (0, 1, 12) under "Top Soil Matches" header
- [x] Tiles look reasonable if the soil name is really long
- [x] Visible for Temporary Location and Site
- [x] Ideally: Use API that Paul made for dummy data

**Non-goal**
- [ ] Reference dummy data until Soil ID returns real info
- [ ] Pressing a soil match will not yet do anything
- [ ] Users won't yet have the option to "see fewer soil matches" to hide the matches below the top 4.

knipec:
@DerekCaelin FYI I moved "See fewer soil matches" to a non-goal for this story, so it will be part of #1198.

DerekCaelin:
@knipec exciting to test this. Is there a way to have the dummy API return 12 responses so we can check, and to return a "really long" response so we can see the wrapping (or did we decide that was unlikely)?

knipec:
@DerekCaelin Great questions. 
- 12 responses: I'll look into this when I get a chance.
- Wrapping: I showed Courtney the wrapping behavior for long soil names, but she said it's unlikely to go over two lines. I'll get a pic for you though when I get a chance.

CourtneyLee333:
[...]

PR #1478, opened 2024-06-03
Feat: Create soil match tiles using data from soil id API

knipec:
## Description
Soil ID page for a site and a temporary location should show tiles with match percentage and soil name.
NOTE: This builds off the feat/soil-id-data branch that is not yet submitted. Please let me know if there's a standard workflow for this.

### Related Issues
Implements #1197

### DIFF
--- a/dev-client/src/screens/LocationScreens/components/soilId/SoilMatchTile.tsx
+++ b/dev-client/src/screens/LocationScreens/components/soilId/SoilMatchTile.tsx
@@ -1,0 +1,69 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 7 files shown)


dev-client/src/screens/LocationScreens/components/soilId/SoilMatchTile.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const SoilMatchTile = ({soil_name, score, onPress}: Props) => {
+  const {t} = useTranslation();
+  // TODO-cknipe: dp versus px?
+
+  return (
+    <Pressable onPress={onPress}>
+      <Box
+        backgroundColor="background.secondary"
+        borderRadius="4px"
```
paulschreiber, line 35: suggestion: if we are using the same border radius in many places, can we make it a constant?

paulschreiber (approved):

dev-client/src/screens/LocationScreens/components/soilId/SoilMatchTile.tsx
knipec: Oo ya in fact I will make `tile` a theme variant

PR #1492, opened 2024-06-06
Fix: Minor UX updates to site data section on soil id screen 

knipec:
## Description
- Put steepness number inside box if steepness is represented by a number, not an image
- "Add slope" button goes to Slope Steepness screen instead of general Slope screen 
- "Slope" title changes to "Slope Steepness"
- If there are no depth intervals for a site, show a single empty row in the soil properties data table 

### Related Issues
This is follow-up work for #1200, #1230

### DIFF
--- a/dev-client/src/components/tables/soilProperties/SoilPropertiesDataTable.tsx
+++ b/dev-client/src/components/tables/soilProperties/SoilPropertiesDataTable.tsx
@@ -92,6 +92,15 @@ export const SoilPropertiesDataTable = ({rows}: Props) => {
     );
   };
 
+  const emptyRow = (
+    <Row justifyContent="flex-start">
+      <DataTableCell text="" width={columnWidthDepth} />
+      <DataTableCell text="" width={columnWidthTexture} />
+      <DataTableCell text="" width={columnWidthRockFragment} />
+      <DataTableCell text="" width={columnWidthColor} />
+    </Row>
+  );
+
   return (
     <ScrollView horizontal={true}>
       <Box>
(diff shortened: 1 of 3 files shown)


dev-client/src/components/tables/soilProperties/SoilPropertiesDataTable.tsx
```diff
@@ -92,6 +92,15 @@ export const SoilPropertiesDataTable = ({rows}: Props) => {
     );
   };
 
+  const emptyRow = (
+    <Row justifyContent="flex-start">
+      <DataTableCell text="" width={columnWidthDepth} />
+      <DataTableCell text="" width={columnWidthTexture} />
+      <DataTableCell text="" width={columnWidthRockFragment} />
+      <DataTableCell text="" width={columnWidthColor} />
```
paulschreiber, line 100: question: is `text=""` needed?

paulschreiber: question: why do we want an empty row?

paulschreiber (approved):

dev-client/src/components/tables/soilProperties/SoilPropertiesDataTable.tsx
knipec: 1. `DataTableCell` expects a required `text` property, which is a `string` type. I could go either way on making `text` optional and having DataTableCell handle the case where it's not supplied, but I feel like it's just as reasonable as-is.

2. We only want it in the case where there are no depth intervals specified (like if the site has "Custom " soil pit depths but then no depths were added, or if the site is part of a project that has "None specified" for soil pit depth intervals)

Issue #1506, opened 2024-06-07
Project screens need correct background


CourtneyLee333:
### App Version

1.0.136

### Platforms

- [X] iOS
- [X] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

The background color of the screens in Project Management should be as follows. This includes the additional screens in the workflow.

- Inputs: #ffffff
- Sites: #e0e0e0
-  Team: #ffffff
- Settings: #ffffff

CourtneyLee333:
@paulschreiber Teams tab still has light grey background below the team member list. Should be white all the way.

paulschreiber:
Fixed.

PR #1507, opened 2024-06-07
fix: set background colors for project tabs

paulschreiber:
## Description
- set background colors for project tabs
- add gray background color to theme
- localize project tab labels

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/1506

### DIFF
--- a/dev-client/src/navigation/constants.ts
+++ b/dev-client/src/navigation/constants.ts
@@ -38,7 +38,6 @@ export type TabStackParamList = {
   };
   [TabRoutes.SETTINGS]: {
     projectId: string;
-    downloadLink: string;
   };
   [TabRoutes.SITES]: {
     projectId: string;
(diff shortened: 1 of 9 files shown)


dev-client/src/navigation/constants.ts
```diff
@@ -38,7 +38,6 @@ export type TabStackParamList = {
   };
   [TabRoutes.SETTINGS]: {
     projectId: string;
-    downloadLink: string;
```
knipec, line 41: Suggestion: move the removal of the downloadLink stuff to another PR, that also includes removing the associated UI and strings

knipec (approved):

dev-client/src/navigation/constants.ts
paulschreiber: That was done elsewhere. Rebasing caused some merged conflicts, so now it's been doubly removed.

Issue #1493, opened 2024-06-06
Required Data Inputs - use correct names, implement descriptions


DerekCaelin:
Correct required data inputs names and descriptions for "Data Collection Requirements" toggles. 

## Location
Go to Project Inputs, look at "Data Collection Requirements"

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=5800-12975&t=yhjCARmYZ6fKqdPv-0)
<img width="580" alt="image" src="https://github.com/techmatters/terraso-mobile-client/assets/10799815/812726b5-c9f1-4d52-8e2d-c024b5c62d45">

## Acceptance Criteria
- [x] "Texture" --> "Soil Texture"
- [x]  "Color" --> "Soil Color"
- [x] Slope and Landscape description: `Includes slope steepness, shape, aspect, and landscape position`
- [x] Texture description `Includes texture class, percent clay, and rock fragments`

PR #1508, opened 2024-06-07
fix: update labels and descriptions for data collection methods

paulschreiber:
## Description
update labels and descriptions for data collection methods

### Related Issues
Addresses #1493.

### DIFF
--- a/dev-client/src/translations/en.json
+++ b/dev-client/src/translations/en.json
@@ -492,9 +492,9 @@
     "required_method": "{{method}} (required)",
     "collection_method": {
       "slope": "Slope and Landscape",
-      "soilTexture": "Texture",
+      "soilTexture": "Soil Texture",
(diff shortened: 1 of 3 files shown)


dev-client/src/translations/en.json
```diff
@@ -507,6 +507,10 @@
       "photos": "Photos",
       "notes": "Notes"
     },
+    "collection_method_description": {
```
tm-ruxandra, line 510: Is it intended that only these methods have descriptions? I'm seeing a bunch of warning spew in my logs about missing i18n keys for the other methods. I feel that we should avoid those warnings by some means.

paulschreiber: Yes, only some methods have descriptions. I've added a check to avoid the warnings.

tm-ruxandra (approved):

PR #1510, opened 2024-06-09
fix: enable noUncheckedIndexedAccess and strictNullChecks

paulschreiber:
## Description
Enable noUncheckedIndexedAccess and strictNullChecks.


paulschreiber:
@shrouxm this flags more possible type errors. is that useful?

Issue #1537, opened 2024-06-10
Text shouldn't look different on Android vs iOS vs Figma


knipec:
Issues:
- Text appears to be a different font between iOS and Android. This is especially noticeable looking at the zeroes on the "100%" on the SoilMatchTile
- Line spacing is also inconsistent between Figma and the app, even though we are using the same values for CSS styling attributes
- Font size also seems slightly different between Figma and the app (as font size was noted as rather small in [this comment](https://github.com/techmatters/terraso-mobile-client/pull/1581#discussion_r1643428171))
- Maybe there are others? 
- After doing this, may want to revert the change in [PR 1538](https://github.com/techmatters/terraso-mobile-client/pull/1538) to pixel-perfect match the design again (or just use non-px units in general, but that may be its own thing)

Thoughts: 
- I'm wondering if this is a nativebase thing. I vaguely recall that when using <Text> from react-native instead of native-base, the text would look different, but haven't pursued that yet.

Image showing Figma, iOS, and Android (from left to right):
![Screenshot 2024-06-10 at 1 26 11 PM](https://github.com/techmatters/terraso-mobile-client/assets/5590815/5b43879a-e7b4-4b77-8ed5-e3b9ec6584e5)

knipec:
Going thru bugs and wondering, do you consider this part of the components library work? @tm-ruxandra @CourtneyLee333

tm-ruxandra:
@knipec This is now redundant with https://github.com/techmatters/terraso-product/issues/909

knipec:
Great, closing this

PR #1538, opened 2024-06-10
fix: make 100% fit on one line in SoilMatchTile on iOS

knipec:
## Description
For some reason the fonts on iOS and Android are different, such that 74px was not sufficient to fit "100%" without wrapping. Just up the width of that section slightly for now, and hopefully [#1537](https://github.com/techmatters/terraso-mobile-client/issues/1537) will more holistically address the issue.
Screenshot 2024-06-10 at 1 31 15 PM --> Screenshot 2024-06-10 at 1 13 21 PM

### Related Issues
Follow-up work to #1197


paulschreiber (approved):

paulschreiber:
@knipec we'll likely need to use non-px units to gracefully handle text magnification app-wide.

PR #1539, opened 2024-06-10
fix: reinstate overlay sheet providers in app stack to fix crashes and conflict w/ FAB

### LINKED ISSUE #1536: Crash tapping sites filter button
## Description
In the Sites List sheet, tapping the filter button crashes the app.
```
 ERROR  'BottomSheetModalInternalContext' cannot be null!, js engine: hermes
```

There's no app code in the stack trace.

Sentry details: https://techmatters.sentry.io/issues/4955980946/

tm-ruxandra:
## Description

Fix crashes in filter screens that use overlay sheet selection components by reintroducing the root-level `BottomSheetModalProvider`. Add another `BottomSheetModalProvider` underneath the `NativeBaseProvider` so that other components can still render under floating NativeBase elements, fixing the unexpected FAB issue.

### Checklist

### Related Issues
Fixes #1536

### Verification steps

Open filter modals and verify that filter selections work as expected. Open site color entry screen and verify that "Done" button no longer floats above info modal.

### DIFF
--- a/dev-client/src/App.tsx
+++ b/dev-client/src/App.tsx
@@ -40,6 +40,7 @@ import 'terraso-mobile-client/config';
 import {Portal} from 'react-native-paper';
 import {enableFreeze} from 'react-native-screens';
 
+import {BottomSheetModalProvider} from '@gorhom/bottom-sheet';
 import * as Sentry from '@sentry/react-native';
 
 import {APP_CONFIG} from 'terraso-mobile-client/config';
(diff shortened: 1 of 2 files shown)


dev-client/src/App.tsx
```diff
@@ -78,18 +79,22 @@ function App(): React.JSX.Element {
...
-              <GeospatialProvider>
-                <Toasts />
-                <HomeScreenContextProvider>
-                  <RootNavigator />
-                </HomeScreenContextProvider>
-              </GeospatialProvider>
-            </NavigationContainer>
-          </Portal.Host>
-        </NativeBaseProvider>
+        <BottomSheetModalProvider>
```
tm-ruxandra, line 82: We need this provider to exist above `NativeBaseProvider` because we have embedded our own selection components inside NativeBase modals, and those components use the Gorhom `BottomSheetModal`. Without a provider above the NativeBase code, it cannot display selection elements. This was the cause of the crash.

dev-client/src/App.tsx
```diff
@@ -78,18 +79,22 @@ function App(): React.JSX.Element {
...
+          <NativeBaseProvider theme={theme}>
+            <Portal.Host>
+              <NavigationContainer>
+                <BottomSheetModalProvider>
```
tm-ruxandra, line 86: We need this provider to exist below the `NavigationContainer` and `NativebaseProvider` because all other bottom sheets need to render on top of NativeBase floating elements (e.g., the Floating Action Button / FAB) and also may need access to the navigation context to allow buttons to navigate the user to different parts of the app.

paulschreiber (approved):

Issue #1535, opened 2024-06-10
Implement design tutorial text for texture guide


DerekCaelin:
Update the texture guide screens to match[ Figma design](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=8336-12439&t=DvpDo5bKaKDFvXxa-0).

## Steps to find screen
1. Go to a site (unaffiliated, or affiliated with "texture" requirement)
2. Go to "soil" tab
3. in a depth, select "Texture"
4. Select info button

CourtneyLee333:
@paulschreiber I think this looks fine for now. In a future enhancement I might like to put the graphic before the question, and have the screen auto-scroll down when an answer is selected. Otherwise, questions and next steps can be overlooked. It seems odd, too, that the paragraph headings are larger than the questions. I'll make an issue for these things.

PR #1545, opened 2024-06-11
fix: add soil texture guide text

paulschreiber:
## Description
Add soli texture guide text.

Note: Text is not final. Do not merge.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/1535


dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
```diff
@@ -102,6 +103,33 @@ export const TextureGuideScreen = (props?: SoilPitInputScreenProps) => {
       <ScrollView bg="grey.300">
         <Column space="16px">
           <Column p="15px" bg="primary.contrast">
+            <Text variant="subtitle1">{t('soil.texture.guide.prepare')}</Text>
+            <BulletList
+              data={[1, 2, 3]}
+              renderItem={i => (
+                <Text variant="body1" color="text.primary">
+                  <Trans i18nKey={`soil.texture.guide.prepare_details_${i}`} />
```
knipec, line 111: FYI (or Suggestion if the design won't be changing): 
If you only have text styled in the same way, and aren't trying to say stuff like "part of this text will be styled this other way" or "we're gonna plop another component into a section of this text", you can just use the standard t() in the `<Text>` component, without needing the `<Trans>` component too.
```suggestion
                  {t(`soil.texture.guide.prepare_details_${i}`)}
```

For here and other places (except the grittyness one)

knipec (approved):

dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
paulschreiber: I'm going to leave it as-is for now. Hopefully we can fix this with a better BulletList component in the future.

PR #1549, opened 2024-06-12
fix: specify buttons as uppercase by default

paulschreiber:
## Description
Specify button text as uppercase by default.
- remove inline styling for most buttons
- style action buttons as mixed case
- style login buttons as mixed case

### DIFF
--- a/dev-client/src/theme.ts
+++ b/dev-client/src/theme.ts
@@ -130,6 +130,11 @@ export const theme = extendTheme({
       },
     },
     Button: {
+      defaultProps: {
+        _pressed: {
+          bg: 'primary.dark',
+        },
+      },
       baseStyle: {
         _disabled: {
           opacity: '100',
--- a/dev-client/src/theme.ts
+++ b/dev-client/src/theme.ts
@@ -130,14 +130,6 @@ export const theme = extendTheme({
       },
     },
     Button: {
-      defaultProps: {
-        _text: {
-          textTransform: 'uppercase',
-        },
-        _pressed: {
-          bg: 'primary.dark',
-        },
-      },
       baseStyle: {
         _disabled: {
           opacity: '100',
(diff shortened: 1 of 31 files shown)


dev-client/src/theme.ts
knipec, line 1: Question: Is this intended as purely a refactor, or should/will it change anything the user would see in the app?
For example: 
- Question/maybe Issue: Some buttons, depending on their nativebase variant (like "solid" or "outline") should have different colors when pressed. Will this override that? (I haven't gotten this branch to try it out.)
- Issue: The buttons in the ActionsModal and LoginScreen are not the only buttons that shouldn't be uppercase -- is there anywhere else? One that I know of is: Ruxandra's work for the Delete Account screen uses mixed case buttons, and I suspect this change would unintentionally make them uppercase. I haven't done a full pass of the Figma but I'm wondering if there are other places like that too?

I feel like we will ultimately want to create different LandPKS-specific button components (that wrap Button from the external library) for the various types of buttons in our app, and then have all of our code use one of those button components.
This wasn't a step that I had envisioned as part of the Buttons work. I think Courtney is still working on the Buttons part of the Components and Styles Library, so I'm wondering (read: Suggestion ;)): 
a) If it is intended purely as a refactor, maybe just don't submit this and wait until Courtney's work is done so we can implement the buttons according to the spec?
b) If it is intended to change something the user will see in the app: Would it make sense to instead maybe check in with Courtney but make an UppercaseButton component instead (or something -- maybe check in with Courtney and Ruxandra because I know they've been working on this)? I do think it's kind of unexpected behavior for developers to have the default Button not use the text as specified by the developer.

I realize this might feel like kind of hefty feedback for a change that was intended to be quick, but I feel the code change here kind of conflicts with the overall intention of trying to align design and engineering -- so I think I'm also trying to use this as a case study to reflect on what doesn't quite feel like it's working about our processes.

knipec (commented):
In the future I imagine we'd want one kind of button for

dev-client/src/theme.ts
paulschreiber: After I figured out how to set the pressed state for buttons, https://github.com/techmatters/terraso-mobile-client/pull/1550, I realized we could eliminate a lot of duplicative code where we manually specify things as uppercase. This should also make future work easier (when we implement more standard components.

- Uppercase: I'll check this against the delete buttons work now that it's landed. The snapshot tests did catch the login buttons. It's possible there are others I missed. I'll look over the Figma document again/
 
- Pressed state: that feedback should go to #1550, which needs to be approved/merged first. I did check a few of the button variants (like site card pin) with Courtney.

PR #1551, opened 2024-06-12
fix: add console integration to Sentry

paulschreiber:
## Description
Track console warnings and errors with Sentry.


tm-ruxandra:
I agree with this change in principle, but since we currently have so much warning spew in the application I am worried it may just add noise to our logs. We should also prioritize fixing our known warnings.

tm-ruxandra (approved):

PR #1554, opened 2024-06-13
fix: pass elevation data through to Learn More screen

paulschreiber:
## Description
Pass elevation data through to Learn More screen when creating sites via map callout.

### Related Issues
Part of fix for https://github.com/techmatters/terraso-product/issues/774

### DIFF
--- a/dev-client/src/screens/CreateSiteScreen/components/CreateSiteView.tsx
+++ b/dev-client/src/screens/CreateSiteScreen/components/CreateSiteView.tsx
@@ -68,7 +68,7 @@ export const CreateSiteView = ({
       const createdSite = await createSiteCallback({...site, elevation});
       if (createdSite !== undefined) {
         homeScreen?.showSiteOnMap(createdSite);
-        navigation.pop();
+        navigation.navigate('BOTTOM_TABS');
       }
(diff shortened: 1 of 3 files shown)


dev-client/src/screens/CreateSiteScreen/components/CreateSiteView.tsx
```diff
@@ -68,7 +68,7 @@ export const CreateSiteView = ({
       const createdSite = await createSiteCallback({...site, elevation});
       if (createdSite !== undefined) {
         homeScreen?.showSiteOnMap(createdSite);
-        navigation.pop();
+        navigation.navigate('BOTTOM_TABS');
```
tm-ruxandra, line 71: Can you explain the difference here between these two behaviors? / why this was necessary for this PR

paulschreiber: When you create after tapping Learn More, you need to go back more than one screen. If you create a site by tapping Create Site directly, you only need to go back one screen. In both cases, you want to go back to the home screen.

tm-ruxandra (approved):

Issue #1179, opened 2024-04-26
Update delete account button for users who have already requested deletion


DerekCaelin:
In the settings menu, signify to the user that their request to delete their account has been received and that they don't need to request it again. Please see parent issue for details.

Dependent on: https://github.com/techmatters/terraso-backend/issues/1112

## Designs
[Figma](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=6995-35081&mode=design&t=EvCSFQJWglFzt9sI-0)

## Acceptance Criteria
- [ ] Greyed out, unclickable delete button on the Settings menu. ("Pending") signifier
- [ ] As a user, I want to know that my request to delete my account has been received and that I can't submit another request.

DerekCaelin:
Ruxandra expressed an interest

CourtneyLee333:
I don't see this happening in v144 on Android or iOS.

CourtneyLee333:
Still waiting on v145 for iOS, but on Android when I press the Delete button it doesn't do an action. I see the button state change, but I remain on the Delete Account screen, get no confirmation message, and see no sign of "pending" status.

CourtneyLee333:
It's working now and looks good!

paulschreiber:
Closing this as done.

PR #1578, opened 2024-06-17
feat: delete account pending buttons

tm-ruxandra:
## Description

Update the delete account workflow to disable buttons when the deletion is pending, simplifying the hook's internals to remove unnecessary state tracking. Do some associated refactoring to make deletion / logout systems more consistently implemented.

### Checklist

### Related Issues
#1179

### Verification steps

Open the user settings screen and execute the delete account workflow. Observe that the action button does not allow the user to click it again while the request is pending. Observe that the button on the settings screen is now disabled and shows the "Pending" sub-label.

### DIFF
--- a/dev-client/src/components/modals/SignOutModal.tsx
+++ b/dev-client/src/components/modals/SignOutModal.tsx
@@ -15,24 +15,31 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
+import {useCallback} from 'react';
 import {useTranslation} from 'react-i18next';
 
-import {ListButton} from 'terraso-mobile-client/components/buttons/list/ListButton';
-import {LogoutModal} from 'terraso-mobile-client/components/modals/LogoutModal';
+import {signOut} from 'terraso-client-shared/account/accountSlice';
 
-export function LogOutButton() {
+import {ConfirmModal} from 'terraso-mobile-client/components/modals/ConfirmModal';
+import {ModalProps} from 'terraso-mobile-client/components/modals/Modal';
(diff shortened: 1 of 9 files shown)


dev-client/src/components/modals/SignOutModal.tsx
```diff
@@ -26,20 +26,20 @@ import {useDispatch} from 'terraso-mobile-client/store';
 
 export type LogoutModalProps = Pick<ModalProps, 'trigger'>;
 
-export function LogoutModal({trigger}: LogoutModalProps) {
+export function SignOutModal({trigger}: LogoutModalProps) {
```
tm-ruxandra, line 29: Change suggested by @paulschreiber on previous PR, I was passing by the settings screen again so I decided to just do it.

dev-client/src/hooks/userDeletionRequest.ts
```diff
@@ -27,13 +27,12 @@ const PREFERENCES_VALUE = 'true';
 export const useUserDeletionRequests = () => {
   const dispatch = useDispatch();
   const {data: user} = useSelector(state => state.account.currentUser);
-  const [isPending, setIsPending] = useState(
-    user!.preferences[PREFERENCES_KEY] === PREFERENCES_VALUE,
-  );
+  const isSaving = useSelector(state => state.account.preferences.saving);
```
tm-ruxandra, line 30: The state here was unnecessary; we can just look at the user's preference value since it will get updated when the dispatched request completes. (We now disable the action button while the request is pending for a smoother user experience.)

paulschreiber (approved):

Issue #1595, opened 2024-06-18
Nav bar should be behind keyboard


CourtneyLee333:

paulschreiber:
On Android, everything gets pushed up under the keyboard. @shrouxm is there a way to hide the navigation whenever the keyboard is shown?

CourtneyLee333:
So much better!

PR #1597, opened 2024-06-18
fix: hide bottom tabs when keyboard is shown

paulschreiber:
## Description
Hide bottom tabs when keyboard is shown

### Related Issues
Addresses #1595.

### DIFF
--- a/dev-client/src/screens/BottomTabsScreen.tsx
+++ b/dev-client/src/screens/BottomTabsScreen.tsx
@@ -15,11 +15,11 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import {memo, useLayoutEffect, useState} from 'react';
-import {Keyboard} from 'react-native';
+import {memo} from 'react';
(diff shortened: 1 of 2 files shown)


dev-client/src/screens/BottomTabsScreen.tsx
```diff
@@ -29,16 +30,36 @@ import {ProjectListScreen} from 'terraso-mobile-client/screens/ProjectListScreen
 import {UserSettingsScreen} from 'terraso-mobile-client/screens/UserSettingsScreen/UserSettingsScreen';
 
 export const BottomTabsScreen = memo(() => {
+  const [keyboardStatus, setKeyboardStatus] = useState(false);
```
tm-ruxandra, line 33: Let's extract this to a dedicated hook function - `useKeyboardStatus` or similar. I don't think it's appropriate for all this logic to live in this component.

tm-ruxandra (approved):

Issue #1592, opened 2024-06-18
Site card menu on Project Management screen covers warning dialog


CourtneyLee333:
Project > Sites > tap the 3 dot menu on the right side of a site card. The menu covers the dialog for the actions. Instead, the menu should close when a menu item is tapped. 

Also, there appears to be a small dot (bullet point?) next to Delete Site menu item. 

![Image](https://github.com/techmatters/terraso-mobile-client/assets/8915145/ea468f50-6b7e-4ad6-b0ff-7f918f46d09f)

CourtneyLee333:
Looks good!

PR #1600, opened 2024-06-18
fix: correct behavior of site card menu on project management screen

paulschreiber:
## Description
Correct behavior of site card menu on project management screen
- remove dot beside "delete site" menu item
- don't block dialog with menu (Android)
- don't crash (iOS)

Use React Native Paper instead of NativeBase.

### Related Issues
Addresses #1592.

### DIFF
--- a/dev-client/src/screens/ProjectSitesScreen.tsx
+++ b/dev-client/src/screens/ProjectSitesScreen.tsx
@@ -15,13 +15,14 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import {useCallback, useMemo} from 'react';
+import {useCallback, useMemo, useState} from 'react';
 import {useTranslation} from 'react-i18next';
+import {Menu} from 'react-native-paper';
 
 import {MaterialTopTabScreenProps} from '@react-navigation/material-top-tabs';
 import type {CompositeScreenProps} from '@react-navigation/native';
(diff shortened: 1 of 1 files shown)


dev-client/src/screens/ProjectSitesScreen.tsx
```diff
@@ -103,34 +86,42 @@ const SiteMenu = ({site}: SiteProps) => {
...
+          siteName: site.name,
+        })}
         actionName={t('projects.sites.remove_site_modal.action_name')}
         handleConfirm={removeSiteFromProjectCallback}
       />
 
       <ConfirmModal
         trigger={onOpen => (
-          <SiteMenuItem
-            iconName="delete"
```
tm-ruxandra, line 127: Have we confirmed that it's okay to omit the icons here? The figma designs still show them.

paulschreiber: Aha. The issue I had was to remove the `.`, which is what the icons appeared as. @CourtneyLee333 okay for this to be a regular menu w/o icons?

tm-ruxandra:
Another possible issue: the button that opens the menu has a pressed-state color that doesn't seem right (not sure if it's related to the wor
Screenshot_1718922114
k here?)

paulschreiber:
@tm-ruxandra fixed the menu color per Courtney.

tm-ruxandra (approved):

Issue #1601, opened 2024-06-20
Site screens need correct background colors


CourtneyLee333:
We need to check the following places and make sure the background colors are correct:

```[tasklist]
### Tasks
- [x] Dashboard screen - white
- [x] Tab slider -grey/200
- [x] Slope screen - grey/300
- [x] Soil screen - grey/300
```

CourtneyLee333:
Looks good, thanks!

PR #1603, opened 2024-06-20
fix: set site screen background colors

paulschreiber:
## Description
Set site screen background colors.

Also clean up project screen background color parameters to have consistent formatting.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/1601

### DIFF
--- a/dev-client/__tests__/integration/__snapshots__/LocationDashboardScreen-test.tsx.snap
+++ b/dev-client/__tests__/integration/__snapshots__/LocationDashboardScreen-test.tsx.snap
@@ -1119,7 +1119,23 @@ exports[`renders correctly 1`] = `
                               ]
                             }
                           >
-                            <RCTScrollView>
+                            <RCTScrollView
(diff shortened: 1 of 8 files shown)


dev-client/src/util.ts
```diff
@@ -63,7 +63,7 @@ export const formatPercent = (value: number) => {
 };
 
 export const formatCoordinate = (value: number) => {
-  return value.toLocaleString(undefined, {
+  return value?.toLocaleString(undefined, {
```
tm-ruxandra, line 66: What does this change (and the one above it in `StaticMapView`)  have to do with site screen colors? I think it should either be explained in the PR or moved to a different PR.

paulschreiber: It's unrelated. I hit this error during debugging and didn't think it was worth a separate PR. I'll move it.

tm-ruxandra (approved):

PR #1610, opened 2024-06-21
fix: avoid null deferences

paulschreiber:
## Description
Avoid some null deferences that pop up in debugging.

### DIFF
--- a/dev-client/src/components/StaticMapView.tsx
+++ b/dev-client/src/components/StaticMapView.tsx
@@ -111,7 +111,7 @@ export const StaticMapView = ({
   const cameraSettings = useMemo(
     () =>
       ({
-        centerCoordinate: coordsToPosition(coords),
+        centerCoordinate: coords ? coordsToPosition(coords) : undefined,
         zoomLevel: zoomLevel,
         animationMode: 'none',
         animationDuration: 0,
--- a/dev-client/src/util.ts
+++ b/dev-client/src/util.ts
@@ -63,7 +63,7 @@ export const formatPercent = (value: number) => {
 };
 
 export const formatCoordinate = (value: number) => {
-  return value.toLocaleString(undefined, {
+  return value?.toLocaleString(undefined, {
     maximumFractionDigits: COORDINATE_PRECISION,
   });
 };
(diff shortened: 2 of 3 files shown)


dev-client/src/util.ts
```diff
@@ -63,7 +63,7 @@ export const formatPercent = (value: number) => {
 };
 
 export const formatCoordinate = (value: number) => {
-  return value.toLocaleString(undefined, {
+  return value?.toLocaleString(undefined, {
```
tm-ruxandra, line 66: I don't think this is correct. The `value` parameter should be declared as optional. This will also return `undefined` for a zero coordinate value, which seems incorrect since zero is a valid lat/lon value.

tm-ruxandra (changes requested):
See comment on `formatCoordinate` function.

dev-client/src/util.ts
paulschreiber: value is not optional here. but if it's undefined, I don't want to try to call `.toLocaleString()` and crash.

This works fine for `0`. Optional chaining only fails on undefined or null, not other falsy values.

Did I misunderstand?

dev-client/src/components/StaticMapView.tsx
```diff
@@ -111,7 +111,7 @@ export const StaticMapView = ({
   const cameraSettings = useMemo(
     () =>
       ({
-        centerCoordinate: coordsToPosition(coords),
+        centerCoordinate: coords ? coordsToPosition(coords) : undefined,
```
knipec, line 114: Question: What were the errors caused by this? I'm confused because `coords` is defined as the `Coords` type, which (as far as I know) can't be undefined/null/falsy. So I'd expect this change to not really have any effect, but maybe I'm missing something?

paulschreiber: This seems to happen with hot reloads when building/debugging in the simulator. It's possible/likely there is a better way to fix this.

paulschreiber:
The fixes in Soil ID ranking seem to be for [Sentry 5563928079](https://techmatters.sentry.io/issues/5563928079/?environment=staging&project=4506299761623040&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=1), which is likely the cause of https://github.com/techmatters/terraso-mobile-client/issues/1694

PR #1613, opened 2024-06-22
feat: Implement welcome screen on first app open

knipec:
## Description
When user first opens the app, before they log in, the welcome screen should show up.
Once they press the "Get Started" button, we will not show them the welcome screen again, unless they delete app data.

### Related Issues
Implements [https://github.com/techmatters/terraso-product/issues/813](https://github.com/techmatters/terraso-product/issues/813) 

### Screenshot
I'll run it by Courtney so may have design tweaks, but here's what it looks like:
Screenshot 2024-06-21 at 5 34 36 PM


dev-client/src/hooks/useStorage.ts
```diff
@@ -0,0 +1,29 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {MMKVLoader, useMMKVStorage} from 'react-native-mmkv-storage';
+
+const MMKV = new MMKVLoader().initialize();
```
knipec, line 20: Looking for takes on if we should consolidate the `MMKV` object for `tokenStorage` in config/index.ts with this?

dev-client/src/screens/WelcomeScreen.tsx
```diff
@@ -0,0 +1,98 @@
...
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {useCallback} from 'react';
+import {useTranslation} from 'react-i18next';
+import {ScrollView} from 'react-native-gesture-handler';
+
+import {Button} from 'native-base';
+
+import LandPksSVG from 'terraso-mobile-client/assets/LandPKS-from-Terraso-logo.svg';
```
paulschreiber, line 24:
```suggestion
import LandPKSTerrasoLogo from 'terraso-mobile-client/assets/LandPKS-from-Terraso-logo.svg';
```

paulschreiber (approved):

dev-client/src/hooks/useStorage.ts
knipec: Keeping them separate for now, as this does not need encryption and it can be argued they are separate enough concepts.

Issue #1640, opened 2024-06-25
Slope screen needs divider between inputs


CourtneyLee333:
The top level screen for Slope in a site needs the light grey divider between Steepness and Shape. See [Figma here](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=5800-13583&t=kal2xQUEnGpihGpF-1), and reference the implemented Soil tab.

CourtneyLee333:
Should be ready in v150.

Issue #688, opened 2023-12-11
Restyle the text input field component


ltseng:
Based on these [Figma designs](https://www.figma.com/file/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?type=design&node-id=2647-10445&mode=design&t=bbYFQNbSJmIzK0Ih-0), please restyle all the text input fields. see https://github.com/techmatters/terraso-product/issues/453 for prior discussion.

We need to make sure that all text input fields that have been built adhere to this new design as well. We also need to pay attention to the complexities of the interactions with these text input fields, like how/when "required" labels show.

DerekCaelin:
@paulschreiber thanks for working on this! In 142 I see max length for site and project titles, but input fields are still not disallowing characters over the max, and I need to deselect the project description field to see the character count when I am over max.

paulschreiber:
@DerekCaelin we are going to address text field requirements separately. This PR is to restyle the the fields, not change their behaviour.

DerekCaelin:
OK, I've removed the behavior requirements from this issue for clarity. Behavior updates are in: https://github.com/techmatters/terraso-mobile-client/issues/1318#issue-2293846203

@CourtneyLee333 ready for your design review.

CourtneyLee333:
Looks good except for:
This one doesn't display the label once you tap the field, so I'm not sure if it was updated:
![Screenshot_20240612-100633](https://github.com/techmatters/terraso-mobile-client/assets/8915145/d5cfeea4-a5f7-42fd-950c-3e69fe173517)

This one should also have the fields on two lines to avoid the issue, but could be handled in a separate issue:
![Screenshot_20240612-100633](https://github.com/techmatters/terraso-mobile-client/assets/8915145/1fc6a04e-9555-4835-a6ea-2979ece70853)

CourtneyLee333:
@paulschreiber Assuming that this is a stop-gap measure until we implement the full component (including [styles and behaviors as described in the component library](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7992-21173&t=kal2xQUEnGpihGpF-1)), I think these are okay. The background grey looks a bit dark - should be 6% opacity black.

PR #1643, opened 2024-06-26
fix: update login screen, texture guide, background colors, divider

paulschreiber:
## Description
* update login screen to match [new designs](https://www.figma.com/design/h4xdqfTq1c84YEn6AZcfHf/Splash-Screen-to-Log-In-Animation?node-id=2-56&t=31mUduoewqR4pY3P-0) (add logo, update text strings, left-aligned, green, uppercase button text (on Android))
* fix text input field background color https://github.com/techmatters/terraso-mobile-client/issues/688
* in texture guide, move videos above questions and reduce question font size https://github.com/techmatters/terraso-mobile-client/issues/1535#issuecomment-2189430152
* add divider to slope screen inputs https://github.com/techmatters/terraso-mobile-client/issues/1640


dev-client/src/screens/LoginScreen.tsx
```diff
@@ -125,22 +128,26 @@ export const LoginScreen = () => {
...
+          <TerrasoLogo width="122px" height="39px" />
+        </Row>
       </Column>
     </Column>
   );
 };
 
 const styles = StyleSheet.create({
+  loginButton: {
+    backgroundColor: 'white',
```
paulschreiber, line 144: This doesn't work with `background.default` or `primary.contrast`. Would like to understand why/fix.

dev-client/src/screens/SlopeScreen/SlopeScreen.tsx
```diff
@@ -68,14 +69,15 @@ export const SlopeScreen = ({siteId}: {siteId: string}) => {
       <DataInputSummary
```
knipec, line 69: Issue: It appears the Figma design calls for a divider here too
```suggestion
      <Divider />
      <DataInputSummary
```

dev-client/src/screens/LoginScreen.tsx
knipec, line 1: Issue: Looks a little bottom-heavy on my iOS emulator device -- is this expected? 
Wonder if we may want to treat all the content from the logo down to the buttons as a box, and use flex or center to position that between the top and the "from Terraso" content.
Screenshot 2024-06-27 at 11 26 28 AM

paulschreiber: I reworked the login screen layout to better support multiple screen sizes:
image

knipec (approved):

paulschreiber:
Here's an iPad version for completeness:
image

Issue #944, opened 2024-03-14
Project-selection dropdown for adding sites to projects does not consider user permissions


ltseng:
### App Version

94

### Platforms

- [X] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

I see the option on the Create Site form to add a site to a project where I'm a project viewer. The Save button fails silently when I attempt to save the site, but this project should not even appear in the list of available projects

### Steps to Reproduce

1. Be a viewer in a project
2. Open the Create Site form from the map
3. Open the "Add to Project" dropdown menu

### Expected behavior

Projects where I am a viewer should not display in the list

### Actual behavior

Projects where I am a viewer do display in the list

### Additional context

https://github.com/techmatters/terraso-mobile-client/assets/2730009/abc94e58-8957-4f11-aee0-618e7b79db1c

DerekCaelin:
Since this is still "to do" as Ruxandra goes on vacation, I'm pulling this issue into the June sprint with the expectation that we may need to reshuffle once she returns.

CourtneyLee333:
Works great!

PR #1648, opened 2024-06-27
fix: filter projects by role for site transfer selection

tm-ruxandra:
## Description

Add a hook allowing us to filter projects based on the current user's role for the project, and use it in `ProjectSelect` and `CreateSiteForm` so that users aren't able to create sites for projects to which they cannot add.

### Checklist

### Related Issues
#944

### Verification steps

Create a new site and verify that projects for which the current user is a viewer are not included in the select list, while projects for which the user is a manager or contributor are included.

### DIFF
--- a/dev-client/src/hooks/permissionHooks.ts
+++ b/dev-client/src/hooks/permissionHooks.ts
@@ -31,7 +31,7 @@ export const useProjectUserRolesFilter = (
        * Filter returns whether we can find a membership for the project
        * that matches the current user and has a role in the accepted list
        */
-      !!Object.values(project.memberships).find(
+      Boolean(Object.values(project.memberships).find(
         membership =>
           currentUser?.id === membership.userId &&
           userRoles.includes(membership.userRole),
--- a/dev-client/src/hooks/permissionHooks.ts
+++ b/dev-client/src/hooks/permissionHooks.ts
@@ -34,7 +34,7 @@ export const useProjectUserRolesFilter = (
       !!Object.values(project.memberships).find(
         membership =>
           currentUser?.id === membership.userId &&
-          userRoles.indexOf(membership.userRole) >= 0,
+          userRoles.includes(membership.userRole),
       );
   } else {
     return _ => true;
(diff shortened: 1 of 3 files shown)


dev-client/src/hooks/permissionHooks.ts
```diff
@@ -0,0 +1,38 @@
...
+  const currentUser = useSelector(state => state.account.currentUser.data);
+  if (userRoles) {
+    return project =>
+      !!Object.values(project.memberships).find(
+        membership =>
+          membership.userId === currentUser?.id &&
+          userRoles.indexOf(membership.userRole) >= 0,
+      );
+  } else {
+    return _ => true;
```
tm-ruxandra, line 36: I decided that this was friendlier to the usage code rather than finding a way to conditionally invoke the hook, which is a little cumbersome in React since it wants hook invocations to be consistent between renders.

dev-client/src/hooks/permissionHooks.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (userRoles) {
+    return project =>
+      /*
+       * Filter returns whether we can find a membership for the project
+       * that matches the current user and has a role in the accepted list
+       */
+      !!Object.values(project.memberships).find(
+        membership =>
+          currentUser?.id === membership.userId &&
+          userRoles.indexOf(membership.userRole) >= 0,
```
paulschreiber, line 37: suggestion: this is easier to read / understand for me.
```suggestion
          userRoles.includes(membership.userRole),
```

dev-client/src/hooks/permissionHooks.ts
```diff
@@ -0,0 +1,42 @@
...
+  userRoles?: ProjectRole[],
+): ProjectFilter => {
+  const currentUser = useSelector(state => state.account.currentUser.data);
+  if (userRoles) {
+    return project =>
+      /*
+       * Filter returns whether we can find a membership for the project
+       * that matches the current user and has a role in the accepted list
+       */
+      !!Object.values(project.memberships).find(
```
paulschreiber, line 34: I like explicitly making something a Boolean. The `!!` is easy to overlook/misread. Will need a corresponding `)` at the end.
```suggestion
      Boolean(Object.values(project.memberships).find(
```

paulschreiber (approved):
Approved (but note suggestions).

PR #1663, opened 2024-06-28
fix: build and version cleanup

paulschreiber:
## Description
- separate out build and version
- update IOS bundle ID
- update app name

### DIFF
--- a/.github/workflows/build-shared-android.yml
+++ b/.github/workflows/build-shared-android.yml
@@ -62,7 +62,7 @@ jobs:
       ENV: ${{ inputs.ENV }}
       TERRASO_BACKEND: ${{ inputs.TERRASO_BACKEND }}
       APP_BUILD: ${{ inputs.APP_BUILD }}
-      APP_VERSION: ${{ inputs.APP_VERSION }}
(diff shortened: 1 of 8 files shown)


.github/workflows/build-shared-android.yml
```diff
@@ -57,7 +61,8 @@ jobs:
       STRICT: ${{ inputs.STRICT }}
       ENV: ${{ inputs.ENV }}
       TERRASO_BACKEND: ${{ inputs.TERRASO_BACKEND }}
-      APP_VERSION: ${{ inputs.APP_VERSION }}
+      APP_BUILD: ${{ inputs.APP_BUILD }}
+      APP_VERSION: ${{ vars.APP_VERSION }}
```
knipec, line 65: Questions: What's `vars`? When we want to make a new version number, where/how do we do that?

knipec (approved):

.github/workflows/build-shared-android.yml
paulschreiber: In GitHub settings, we have both secrets and vars. To bump the version, we edit the settings there. (We can change this later if we decide to put the version in source.)

PR #1683, opened 2024-07-01
fix: suppress reanimated warning

### LINKED ISSUE #1548: Reanimated generates reduceMotion console warning
This console warning is generated
```
 WARN  [Reanimated] Tried to modify key `reduceMotion` of an object which has been already passed to a worklet. See 
https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#tried-to-modify-key-of-an-object-which-has-been-converted-to-a-shareable 
for more details.
```

paulschreiber:
## Description
Suppress suppress reanimated warning.

### Related Issues
Fixes #1548.

### DIFF
--- a/dev-client/src/App.tsx
+++ b/dev-client/src/App.tsx
@@ -67,6 +67,7 @@ Mapbox.setAccessToken(APP_CONFIG.mapboxAccessToken);
 
 LogBox.ignoreLogs([
   'In React 18, SSRProvider is not necessary and is a noop. You can remove it from your app.',
+  /^[Reanimated] Tried to modify key `reduceMotion` of an object which has been already passed to a worklet/,
 ]);
 
 const store = createStore();
--- a/dev-client/src/App.tsx
+++ b/dev-client/src/App.tsx
@@ -67,7 +67,8 @@ Mapbox.setAccessToken(APP_CONFIG.mapboxAccessToken);
 
 LogBox.ignoreLogs([
   'In React 18, SSRProvider is not necessary and is a noop. You can remove it from your app.',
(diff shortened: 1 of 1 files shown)


tm-ruxandra:
Is there a reason you've decided to suppress the warning rather than address its cause? If it's a known issue without a workaround I'm okay with doing this, but I would rather fix the actual cause of the warning problem.

dev-client/src/App.tsx
```diff
@@ -67,6 +67,7 @@ Mapbox.setAccessToken(APP_CONFIG.mapboxAccessToken);
 
 LogBox.ignoreLogs([
   'In React 18, SSRProvider is not necessary and is a noop. You can remove it from your app.',
+  /^[Reanimated] Tried to modify key `reduceMotion` of an object which has been already passed to a worklet/,
```
knipec, line 70: Question/Issue: If this is a regular expression, do you need to escape the `[` and `]`?

paulschreiber:
This is a bug in a library that we can't fix ourselves. We're suppressing so it doesn't clutter the UI. It still gets logged to the console.

dev-client/src/App.tsx
knipec, line 1: Question: This warning is happening because of a bug in a library we depend on, right? 
If so, do we intend to remove this if the issue is fixed in the relevant library? 
If so, is there a good way we can remember to remove it when/if that happens?

paulschreiber: I've subscribed to the issues in the underlying repository and should be notified when they get resolved. I'll leave a TODO comment as well.

knipec (approved):

Issue #1635, opened 2024-06-24
Soil ID match tile percentage should avoid wrapping in more cases


knipec:
### Repro 
On certain devices, on the "Soil ID" page, the match percentage on the `SoilMatchTile` will wrap instead of staying on one line, even if there is a reasonable amount of space to accommodate it. 
It will more likely happen with "100%", but depending on the phone's size settings it may happen with 2-digit numbers as well
Decide what layout logic to use to satisfy: 
- Percentage doesn't wrap (unless the screen is very narrow, or the font size is larger than we want to support)
- Columns still line up

![image](https://github.com/techmatters/terraso-mobile-client/assets/5590815/4a41daf7-c577-4d11-9166-d189d7979a6e)

### Context 
Discussion in https://github.com/techmatters/terraso-mobile-client/issues/1197 
Potential solution: Instead of a fixed pixel width, use a percent width, or perhaps dp?

paulschreiber:
Would `adjustsFontSizeToFit` work here?

knipec:
> Would `adjustsFontSizeToFit` work here?

It is another option, but I think it looks weird because it could make things that are a higher percent match be a smaller size:
![Screenshot 2024-06-24 at 4 12 22 PM](https://github.com/techmatters/terraso-mobile-client/assets/5590815/6cea1b6a-6270-4315-9c42-a6602e10f686)

CourtneyLee333:
Can I see what a single digit percent looks like?

CourtneyLee333:
Also, it seems that we could just decrease the font size. We still have the issue of what happens on a small screen with the font size turned up. I'd rather address this particular instance in the context of larger responsive design strategy throughout the app. It's very unlikely, and maybe impossible, for a 100% match to be returned.

paulschreiber:
Here is an example
![image](https://github.com/techmatters/terraso-mobile-client/assets/86784/1bc8a6e4-0d65-4df5-aa42-2783d6f0c53f)

CourtneyLee333:
Noting that, though we had been told a 100% match would be impossible, it is actually quite common from what we've seen in the app so far.

CourtneyLee333:
Perhaps something to think through - do we need to show percentage numbers? Do they help or just make things confusing. We could use a number ranking system instead.

knipec:
Also seeing "100" with the "%" sign missing with lat/long 37.4303, -122.12658: 
[...]

PR #1696, opened 2024-07-03
fix: Decrease font size of soil match percent, use adjustsFontSizeToFit

knipec:
## Description
Another attempt at preventing the "100%" from wrapping (or truncating) -- this time by decreasing the font size and using adjustsfontSizeToFit onto one line.

### Related Issues
Addresses #1635 
However, I still don't understand why:
a) With 30px or 32px font size, we'd get "100" without the "%" sign, even with adjustsFontSizeToFit?
b) Why font size and Box width are both specified in fixed pixels but not 

Before --> After:
 --> 

And also any individual percentage, like the "100%", will get smaller if it would overflow its box, like this (but note this is not the new 24px font size):


paulschreiber (approved):

knipec:
Merging so folks can try it on their devices, and open to Courtney's feedback!

CourtneyLee333:
It looks great to me on Android phone and iPad. Thanks!

PR #1714, opened 2024-07-05
feat: improve UX of location permission requests

### LINKED ISSUE #1645: Map "My Location" icon button needs to prompt for location permission
Any time a user has not given location permission to the app and they tap the "My Location" IconButton on the map, they should be prompted to give the app permission. I believe we can reuse the [modal used for camera permission seen here in Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7711-20430&t=mDy56PEOs97i2XXC-1), with the following text:

LandPKS needs to access your location
For the best map experience, grant LandPKS access to your location in device settings. 
(Cancel) (Go to settings)

shrouxm:
## Description
Improves UX of location permission requests by:
- not asking for location permission until after user signs in
- re-requesting location permission when user clicks on the location icon on the map

### Checklist

### Related Issues
Fixes #1645

In the course of working on this PR, I discovered two new bugs: #1749 and #1750. I confirmed they are pre-existing (not introduced by this PR), and could not find an easy workaround, so I'm just filing them and leaving them in the backlog.


dev-client/src/screens/HomeScreen/components/MapSearch.tsx
```diff
@@ -211,14 +214,22 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
...
+              <IconButton
+                name="my-location"
+                _icon={{color: 'action.active'}}
+                bgColor="white"
+                padding={2}
+                borderRadius={15}
+                onPress={onRequest}
+              />
+            )}
+          </PermissionsRequestModal>
```
knipec, line 232: Sorta-suggestion: I've seen a few places in our codebase this model where the component hierarchy has a modal that is opened by pressing a button, and the button defined within its onRequest. It reads a little convoluted to me, because I tend to expect to see the button in the hierarchy because that's what's visible.

If you'd prefer not to refactor it that way here, perhaps this can be a future code standard discussion -- but wanted to note it!

knipec (approved):

dev-client/src/screens/HomeScreen/components/MapSearch.tsx
shrouxm: discussed on call, doing a rename for now but considering changing structurally in future

Issue #1547, opened 2024-06-12
Cache elevation values on client


DerekCaelin:
## Description
Before calling the elevation API:
- round to five decimal places
- check local cache for value

After calling the elevation API:
- cache value in a redux store

paulschreiber:
This will be available in 161.

paulschreiber:
Closing.

PR #1736, opened 2024-07-08
feat: cache elevation data in client redux slice

tm-ruxandra:
## Description

Add a new slice to the terraso mobile client which handles elevation caching based on 5 decimal places of precision. The data hook will attempt to load elevation once (does not retry on failure).

### Checklist

### Related Issues
#1547

### Verification steps

Select the same location twice on the home screen (within 5 decimal places of precision). Verify that the elevation value is already loaded.


dev-client/src/model/elevation/elevationTypes.ts
```diff
@@ -0,0 +1,22 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+export type ElevationKey = `(${number}, ${number})`;
+export type ElevationRecord = {
+  value?: number;
+  fetching: boolean;
```
tm-ruxandra, line 21: I included the ability to track when an elevation request is fetching in case we want to e.g. display a "no data available" message at some point.

dev-client/src/model/elevation/elevationFunctions.ts
```diff
@@ -15,21 +15,10 @@
...
-    const elevation = await getElevation(coords.latitude, coords.longitude);
-    if (elevation !== undefined) {
-      setSiteElevationValue(elevation);
-    }
-  }, [coords]);
+import {ElevationKey} from 'terraso-mobile-client/model/elevation/elevationTypes';
 
-  return siteElevationValue;
+export const elevationKey = (coords: Coords): ElevationKey => {
+  return `(${coords.longitude.toFixed(5)}, ${coords.latitude.toFixed(5)})` as ElevationKey;
```
paulschreiber, line 23: issue: use `formatCoordinate()` helper.

paulschreiber (approved):
I've flagged one issue, but approved (so you won't have to re-request approval).

Issue #1740, opened 2024-07-09
when searching, long text extends beyond search bar when search is deselected


DerekCaelin:
### App Version

160

### Platforms

- [X] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

When typing a long string into the search bar, and then deselecting the search bar or tapping "search", the text extends beyond the vertical bounds of the search bar. It seems to wrap.

### Steps to Reproduce

1. On the home screen, type a long string, such as `34.97277, -106.68278`
2. select "search" on the keyboard
3. see issue

### Expected behavior

Text remains in the search bar but is truncated to fit within the horizontal bounds of the search bar

### Actual behavior

Text extends beyond the search bar, as if wrapping. 

### Additional context

![IMG_2CE3BE799310-1](https://github.com/techmatters/terraso-mobile-client/assets/10799815/7d7414c8-3929-4856-a2f8-3ee0e3968502)

paulschreiber:
This only happens when text magnification is greater than 100%.

DerekCaelin:
1. I think my text magnification is at the default
2. Isn't this a bug, even if I have changed my text magnification?

paulschreiber:
Yes, it's a bug even if magnification is changed. We hope to support responsive text sizing in a future version.

PR #1741, opened 2024-07-09
fix: increase map search bar width

paulschreiber:
## Description
Increase map search bar width.

### Related Issues
This mitigates #1740 (but doesn't solve it entirely).

### DIFF
--- a/dev-client/src/screens/HomeScreen/components/MapSearch.tsx
+++ b/dev-client/src/screens/HomeScreen/components/MapSearch.tsx
@@ -194,7 +194,7 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
                 placeholder={t('search.placeholder')}
                 style={{
                   ...searchBarStyles.search,
-                  ...searchBarStyles.wideSearch,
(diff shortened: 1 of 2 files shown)


dev-client/src/screens/HomeScreen/components/MapSearch.tsx
```diff
@@ -192,7 +192,10 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
                 }}
                 value={query}
                 placeholder={t('search.placeholder')}
-                style={searchBarStyles.search}
+                style={{
+                  ...searchBarStyles.search,
+                  ...searchBarStyles.wideSearch,
```
knipec, line 197: Suggestion: It feels weird to me to use both styles, especially when they both have a width property. Even though I know the latter one will override it, I'd prefer either 
- using two styles but pulling out a `standardWidthSearch` or something so the width part is separate, or 
- maybe creating a separate `wideSearch` that duplicates the other style properties, or 
- if you prefer this approach, maybe just name it `wideSearchOverride` instead of `wideSearch`, so it's clear it's not the same type of thing as `search`.

Issue #1651, opened 2024-06-27
Location and Soil Properties Info Sheets on Soil Info Sheet open too tall and X button does not work


CourtneyLee333:
This appears to be the same issue we had the previous implementation of the Data Privacy sheet. 

![Image](https://github.com/techmatters/terraso-mobile-client/assets/8915145/e6a5d806-8c35-4c03-85a4-1d1832dcf4dc)

![Image](https://github.com/techmatters/terraso-mobile-client/assets/8915145/5f748c44-575f-44f3-a593-65ec447f78ac)

knipec:
Triage notes: 
- Components work is going to touch overlay sheets. Prefer to fix this via components work (#697), and not need this fix for Capri

paulschreiber:
@knipec can web fix the close button now and the sheet later?

knipec:
I'm not sure -- would be nice to have for Capri, but we think it's a P2 so not critical for Capri because you can swipe the sheet down instead of pressing the X

tm-ruxandra:
@paulschreiber @knipec the issue is that the sheet's unexpected height causes the close button to be out of tapping range for users on some devices. I can poke around to try and find a good fix.

paulschreiber:
Please verify in 168.

DerekCaelin:
This looks good to me. Over to @CourtneyLee333, who filed the issuse.

CourtneyLee333:
This passes for me too!

PR #1742, opened 2024-07-10
fix: make header height context available to all overlay sheets

tm-ruxandra:
## Description

We had issues with nested overlay sheets opening to inconsistent heights, which in turn made their close buttons inaccessible. This was because the `HeaderHeight` context that tells them their opening limit is scoped to the `ScreenScaffold` component, which is context that nested sheets do not have (since they are opening in the `BottomSheetModalProvider` outside of the scaffold). This fixes the issue by making `HeaderHeight` context available to the entire app, with `ScreenScaffold` just setting the current value as a side effect.

### Checklist

### Related Issues
#1651

### Verification steps

Open info overlay sheets, including the nested ones in Soil ID, and verify that all open to the expected height and can be closed on both IOS and Android.

### DIFF
--- a/dev-client/src/screens/ScreenScaffold.tsx
+++ b/dev-client/src/screens/ScreenScaffold.tsx
@@ -15,12 +15,12 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import {useCallback, useState} from 'react';
+import {useCallback, useEffect, useState} from 'react';
 import {LayoutChangeEvent, StatusBar, StyleSheet, View} from 'react-native';
 import {SafeAreaView, useSafeAreaInsets} from 'react-native-safe-area-context';
 
 import {Box, Column} from 'terraso-mobile-client/components/NativeBaseAdapters';
-import {HeaderHeightContext} from 'terraso-mobile-client/context/HeaderHeightContext';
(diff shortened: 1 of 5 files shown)


dev-client/src/screens/ScreenScaffold.tsx
```diff
@@ -34,14 +34,19 @@ export const ScreenScaffold = ({
...
+    [setAppBarHeight],
+  );
   const safeAreaTop = useSafeAreaInsets().top;
 
-  const onLayout = useCallback(
-    (e: LayoutChangeEvent) => setHeaderHeight(e.nativeEvent.layout.height),
-    [setHeaderHeight],
+  const {setHeaderHeight} = useHeaderHeight();
+  useEffect(
+    () => setHeaderHeight(safeAreaTop + (appBarHeight ?? 0)),
```
knipec, line 48: Oo nice that all the things called "header height" mean the same thing now :)

knipec (approved):
Just to confirm understanding: The first OverlaySheet would be a child of the ScreenScaffold, so would receive the headerHeight via the context, but subsequent nested OverlaySheets would not receive it because they'd exist at some other level of the component hierarchy that was not a child of the ScreenScaffold -- is that accurate?

tm-ruxandra:
@knipec that's correct.

Issue #1289, opened 2024-05-10
Make ExternalLink consistent with design


tm-ruxandra:
Standardize usage of links to external URLs, using `ExternalLink` component. Update component to [match design.](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7242-19281&m=dev).

Go through the mobile codebase and replace any non-standard ad-hoc equivalents with these components which correspond to known design spec elements.

paulschreiber:
question: would it make sense to have `<CloseButton size="...">` instead?

tm-ruxandra:
> question: would it make sense to have `<CloseButton size="...">` instead?

This was the case originally, however I ended up deciding it would be better to have them as two different components. First, there's no material value in sharing the code between them (it's trivial). Second, the buttons are used in completely different contexts - it's not like they're two variants of one button with similar usage (like an action button colored red or green if its action is deletion or creation). I think it'll avoid more confusion in the long term to talk about a "close button" or "big close button" vs. "close button" having to remember if it's supposed to have a particular size in some context.

DerekCaelin:
Related issue: https://github.com/techmatters/terraso-mobile-client/issues/1395

DerekCaelin:
@tm-ruxandra I checked the data privacy info sheet. To my eye the ExternalLink matches the design. Over to @CourtneyLee333

CourtneyLee333:
@tm-ruxandra Can you tell me what build to find this in?

tm-ruxandra:
@CourtneyLee333 should be in anything after v170beta

CourtneyLee333:
Looks great!

PR #1778, opened 2024-07-16
feat: spec compliance for ExternalLink

tm-ruxandra:
## Description

Reimplement `ExternalLink` to be compliant with spec -- to assist with future work, this involved spinning off a common `TextButton` component which supports this component as well as expected future work in the components documentation.

### Checklist

### Related Issues
#1289

### Verification steps

Open sections of the application containing external links. Verify that they match the component specification. (Note that this is not the same as `InternalLink`, which will eventually open an embedded web view rather than a browser window.)

### DIFF
--- a/dev-client/src/components/buttons/TextButton.tsx
+++ b/dev-client/src/components/buttons/TextButton.tsx
@@ -1,0 +1,99 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


dev-client/src/components/buttons/TextButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+  leftIcon,
+  rightIcon,
+  disabled,
+  onPress,
+}: TextButtonProps) => {
+  const [pressed, setPressed] = useState(false);
+  const colors = disabled ? COLORS.disabled : COLORS[type];
+
+  return (
+    <Button
```
tm-ruxandra, line 51: The styling here is very cumbersome -- am open to an alternative implementation, but we will also be revisiting all buttons code in the future so we don't need to make a definitive call right now.

shrouxm (approved):
LGTM! agree the styling is cumbersome, but wouldn't be surprised if there isn't a better way with the nativebase button. did you try using RN paper's text button? https://callstack.github.io/react-native-paper/docs/components/Button/

tm-ruxandra:
@shrouxm I experimented a little but was unfamiliar with the API especially for things like icons. We can revisit this and see if it's a candidate for more generally replacing the NativeBase button.

PR #1788, opened 2024-07-17
fix: gracefully handle areas where elevation data doesn’t exist

paulschreiber:
## Description
Gracefully handle areas where elevation data doesn’t exist: return undefined instead of throwing an exception.


paulschreiber:
Subsumed by #1789.

Issue #848, opened 2024-02-08
Styling feedback for slope data input


ltseng:
- [x] The modal for "enter steepness percentage or degree" should show error text when user enters a degree number outside the range. I think turning the helper text red would suffice.
<img src="https://github.com/techmatters/terraso-mobile-client/assets/2730009/92cdb953-a6ef-4b0d-b327-61abcf5842fb" width="300px">

- [x] Limit manual input for percentage to 999%, and degrees field input to 99 degrees (i.e. 2 digit numeric input). See discussion below for more context.
<img src="https://github.com/techmatters/terraso-mobile-client/assets/2730009/84c9e5eb-4867-41b0-8280-f022dabd9f89" width="300px">

- [ ] I'm concerned that the grey font is too light in the field placeholders, so I'm hoping that gets updated when they update the form styles in general.
- [x] Hide the Done button on the slope data input screens from project/site viewers
- [x] Hide the manual and slope meter buttons on the slope steepness screens for project/site viewers.
- [x] On iPads, the "or" in the "Enter steepness or percentage degree" headline for the manual slope steepness entry modal is omitted

<img src="https://github.com/techmatters/terraso-mobile-client/assets/2730009/13cef606-b71f-4ecf-a041-894efe68336f">

ltseng:
@CourtneyLee333 to respond to your point about maximum steepness percentage, I've seen a couple references saying that anything over 60 degrees is basically a rock face. The iPhone's accelerometer seems to go up to 5729% or 89 degrees. I'd be happy to set the maximum at somewhere in between 😂 .

CourtneyLee333:
We could limit that field to 3 numbers. Would that solve the problem? Will people actually need to enter something over 999%? 

We might also want to limit the degrees field to two numbers, since 89 is the max allowed.

ltseng:
I think 999% would be a safe maximum for percentage :) I'll update the issue description now.

PR #1834, opened 2024-07-23
fix: on slope data screen, hide edit buttons from viewers and limit percentage input

paulschreiber:
## Description
On slope data screen:
- hide edit buttons from viewers
- limit percentage to a maximum of 999%
- add maxlength attributes to percentage and degree text fields

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/848

Not adjusting the grey color, which will be handled with text field component work.

### DIFF
--- a/dev-client/src/screens/SlopeScreen/SlopeSteepnessScreen.tsx
+++ b/dev-client/src/screens/SlopeScreen/SlopeSteepnessScreen.tsx
@@ -43,6 +43,8 @@ import {
   Row,
   Text,
 } from 'terraso-mobile-client/components/NativeBaseAdapters';
+import {RestrictBySiteRole} from 'terraso-mobile-client/components/RestrictByRole';
+import {SiteRoleContextProvider} from 'terraso-mobile-client/context/SiteRoleContext';
 import {AppBar} from 'terraso-mobile-client/navigation/components/AppBar';
 import {useNavigation} from 'terraso-mobile-client/navigation/hooks/useNavigation';
 import {ScreenScaffold} from 'terraso-mobile-client/screens/ScreenScaffold';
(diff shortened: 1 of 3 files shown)


dev-client/src/screens/SlopeScreen/SlopeSteepnessScreen.tsx
```diff
@@ -43,6 +43,8 @@ import {
   Row,
   Text,
 } from 'terraso-mobile-client/components/NativeBaseAdapters';
+import {RestrictBySiteRole} from 'terraso-mobile-client/components/RestrictByRole';
```
knipec, line 46: Suggestion: In lieu of any larger-scope changes to add the provider to screens more generally, it could be helpful to just make a comment next to the `RestrictBySiteRole` definition like "Note: If you use this, some ancestor must be wrapped in the SiteRoleContextProvider". Same with `RestrictByProjectRole`. Just something to point future developers in the right direction if they run into the same thing we did.

knipec (approved):

dev-client/src/screens/SlopeScreen/components/ManualSteepnessModal.tsx
```diff
@@ -139,6 +140,7 @@ export const ManualSteepnessModal = ({siteId}: Props) => {
                 helpText={t('slope.steepness.percentage_help')}
                 placeholder={t('slope.steepness.percentage_placeholder')}
                 textInputLabel={t('slope.steepness.percentage_placeholder')}
+                maxLength={3}
```
knipec, line 143: Question: Is this redundant with the min/max specification above?

paulschreiber: Nope. This _prevents_ someone from typing in 9999. The min/max specification only shows an error afterwards.

Issue #1827, opened 2024-07-22
App hangs searching for lat/lng pair


paulschreiber:
### App Version

1.0.1 (172)

### Platforms

- [ ] iOS
- [X] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

App hangs searching for a lat/ng pair.

Sentry profile: https://techmatters.sentry.io/issues/5632803701/?project=4506299761623040

### Steps to Reproduce

1. Start typing `36.30218,-83.19369` into the search box (type, don't paste the whole thing in)
2. Around `36.30218,-83.19` the app hangs

DerekCaelin:
- Relevant issues: https://github.com/techmatters/terraso-mobile-client/issues/772
- Look to google maps as the method of putting in the lat/long
- solution may involve using a debounce system

paulschreiber:
We use [use-debounce](url) in the web app.

DerekCaelin:
@paulschreiber I was never able to reproduce on my devices... can you review this?

DerekCaelin:
Pulling into August since the QA still needs to happen.

paulschreiber:
I can n longer reproduce. Likely mitigated by https://github.com/techmatters/terraso-mobile-client/pull/1848

PR #1848, opened 2024-07-25
fix: debounce for search queries

tm-ruxandra:
## Description

Wrap the map query suggestion system in a 500ms debounce to avoid hammering the map or related downstream APIs. Also refactor the component to extract the query suggestion system to a self-contained hook which the component uses. Additionally, wrap relevant functions in `useCallback` and handle `zoomTo` calls through a single side-effect driven by the suggestion system's selected coordinates rather than scattered through async functions. We believe that the hang issue was caused by some combination of these factors (many async calls affecting zoom position) and that this will mitigate the issue, but were not able to find a direct repro case.

### Checklist

### Related Issues
#1827 

### Verification steps

Enter string and coordinate-based queries in the sites screen search bar and verify that behavior is still as expected, with the addition of a 500ms debounce window before input results in on-screen changes.


dev-client/src/hooks/useMapSuggestions.ts
```diff
@@ -0,0 +1,90 @@
+import {useCallback, useState} from 'react';
```
paulschreiber, line 1: issue: add copyright notice.

dev-client/src/hooks/useMapSuggestions.ts
```diff
@@ -0,0 +1,95 @@
...
+      newAbortController.signal,
+    );
+    setAbortController(null);
+    return newSuggestions;
+  }, []);
+
+  const querySuggestions = useCallback(
+    async (queryText: string) => {
+      if (isValidCoordinates(queryText)) {
+        const [latitude, longitude] = queryText.split(',').map(Number);
```
paulschreiber, line 50: suggestion: validate the coordinates before submitting them. (we have `isValidLatitude`/ `isValidLongitude` helpers already).

but: not sure what to do if they're invalid. Ask @CourtneyLee333?

This also is somewhat out of scope, so we could do that work in another PR.

dev-client/src/hooks/useMapSuggestions.ts
```diff
@@ -0,0 +1,95 @@
...
+            }
+            return null;
+          });
+          setSuggestions([]);
+        }
+      }
+    },
+    [makeSuggestionsApiCall],
+  );
+  const debouncedQuerySuggestions = useDebouncedCallback(querySuggestions, 500);
```
paulschreiber, line 78: suggestion: put the debounce time in a constant.

dev-client/src/hooks/useMapSuggestions.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import {MAP_QUERY_MIN_LENGTH} from 'terraso-mobile-client/constants';
+import {
+  initMapSearch,
+  Suggestion,
+} from 'terraso-mobile-client/services/mapSearchService';
+import {isValidCoordinates} from 'terraso-mobile-client/util';
+
+/*
+ * Logic extracted from screens/HomeScreen/components/MapSearch.tsx; see its revisions
```
knipec, line 15: Oo I haven't been doing this when extracting things into new files but I like it

dev-client/src/hooks/useMapSuggestions.ts
tm-ruxandra: I don't understand your suggestion - we're invoking `isValidCoordinates` right here which in turn invokes `isValidLatitude(latitude) && isValidLongitude(longitude)`.

knipec (approved):

dev-client/src/hooks/useMapSuggestions.ts
paulschreiber: Oops.

Issue #1719, opened 2024-07-05
Make MenuList and MenuItem consistent with component design


CourtneyLee333:
Menus throughout the app need to implement the MenuList and MenuItem components per the [design](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7604-45783&t=Esm9mQnDuie9QHKp-1).

tm-ruxandra:
@CourtneyLee333 It'd be helpful if you could make some minor updates to the docs now that things are implemented:
- The "destructive" variant ended up being part of the main component, so it can just be called `destructive`
- I added a property to control whether text is uppercase (which it is in the user settings menu) which would be nice to have documented
- It would also be nice to document that menu items support sub-label text (like the "deletion pending" text on the settings menu)

DerekCaelin:
@tm-ruxandra in which build should we review MenuList and MenuItem?

tm-ruxandra:
@DerekCaelin Should be out on v177beta now.

DerekCaelin:
Looking at the team member list and the settings screen, this appears to match the design, but I'll trust @CourtneyLee333's keener eye here.

CourtneyLee333:
Looks great!

Issue #1877, opened 2024-07-29
Make chip consistent with component design


CourtneyLee333:
Chips should be consistent across the app with the [design](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=9334-12527&t=1rygqWdA7ochfVlJ-1).

knipec:
Should be ready to QA in v197

DerekCaelin:
This looks right to me - over to @CourtneyLee333 for her review

CourtneyLee333:
Looks good on Android, iOS is currently crashing on launch.

Issue #1774, opened 2024-07-08
Make FormOverlaySheet consistent with component design


CourtneyLee333:
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7289-34451&t=FWzt402zPDCk2duY-1)

tm-ruxandra:
Per discussion with @CourtneyLee333 

-`Select` component should use this new `FormOverlaySheet`
- Soil depth interval editor should be migrated to `InfoSheet`
- Soil settings editor should be migrated to `InfoSheet`
- Camera/photo source picker should be standalone component, no underlying reused sheet component

DerekCaelin:
My uncultured eyes think this looks good. I found a FormOverlaySheet sheet in selecting project depths.

CourtneyLee333:
Looks great!

PR #1878, opened 2024-07-29
feat: menu components

tm-ruxandra:
## Description

[Per design spec](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7604-45783&m=dev): consolidate the user settings buttons, project user list, and select item lists into a single pair of components called `MenuList` and `MenuItem`. Both are implemented without any direct `NativeBase` dependencies. The new component supports all functionality required by the spec, plus a legacy feature to include the "leave project" button in the user list as a child component. They also support standard accessibility attributes like `accessibilityLabel`, `accessibilityRole`, accessibilityState` etc.

Also refactor some of the `UserList` components to make them easier to reuse now that the ad-hoc menu item implementation has been removed, extracting `ProfilePic` and `RolePill`.

### Checklist

### Related Issues
#1719

### Verification steps

Visit the three affected screens and verify that items are compliant with spec on Figma. I also encourage setting dummy data to check edge-cases like long names, lists that exceed screen height, etc.


dev-client/src/components/inputs/Select.tsx
```diff
@@ -152,11 +100,43 @@ export const Select = <T, Nullable extends boolean>({
...
+            </Text>
+          </Row>
+        </Pressable>
+        <MenuList>
+          {optionsWithNull.map((option, i) => {
+            const itemLabel = option ? renderValue(option) : unselectedLabel!;
+            const selected = option === value;
+            return (
+              <MenuItem
+                key={itemLabel + i}
```
tm-ruxandra, line 127: Would appreciate a double-check here that this is ok; the item lists tend not to change often + the component is simple so the cost of re-rendering if the list gets reordered etc seemed low, vs. requiring unique IDs as an input to `Select` which seemed out-of-scope for this ticket.

dev-client/src/components/icons/Icon.tsx
```diff
@@ -33,7 +34,7 @@ export type IconProps = Omit<
 > &
   NativeBaseProps & {
     size?: keyof typeof theme.components.Icon.sizes | number;
-    color?: ThemeColor | string;
+    color?: ThemeColor | ColorValue | string;
```
tm-ruxandra, line 37: This allows us to take the output of `convertColorProp` as an input here, which is more convenient for some usages.

dev-client/src/components/menus/MenuItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+      return styles.default;
+    }
+  }, [disabled, variant]);
+
+  return (
+    <TouchableRipple
+      onPress={onPress!}
+      disabled={disabled}
+      style={selected ? styles.selected : undefined}
+      accessibilityRole="menuitem"
```
tm-ruxandra, line 66: These attributes seem common in library code that I referenced (e.g., the menu components in Native Paper.) We should start thinking about how & when to use them effectively in future work.

dev-client/src/components/menus/MenuItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+import {Text, TouchableRipple} from 'react-native-paper';
+
+import {Icon, IconName} from 'terraso-mobile-client/components/icons/Icon';
+import {convertColorProp} from 'terraso-mobile-client/components/util/nativeBaseAdapters';
+
+export type MenuItemVariant = 'default' | 'destructive';
+
+export type MenuItemProps = {
+  variant?: MenuItemVariant;
+  uppercase?: boolean;
```
tm-ruxandra, line 29: Open to a better way of specifying this; some usages specified in the design are uppercased (the settings screen) and some aren't, so having a simple flag handle it seemed reasonable. I didn't want to add arbitrary text styling props in the interest of discouraging complex / unplanned usages.

dev-client/src/components/inputs/Select.tsx
knipec: This seems fine to me, at least for the way we use it. May be worth a comment just in case in the future we do end up with a `renderValue` that's expensive in conjunction with a reorder-able list -- but I agree that seems out-of-scope for now.

knipec: Mostly just thinking aloud here -- I guess another potential issue is reordering in conjunction with multiple items with the same `renderValue`s, or an item whose `renderValue` is the same as the unselectedLabel... but that also seems rather unlikely

dev-client/src/components/inputs/Select.tsx
knipec, line 1: This is incredibly minor, but it looks like with a long scrolling list, the corners of the BottomSheetScrollView
- are square on the green "DONE" top row
- are square on the selected item
- are rounded on all other items

I'd be fine saying it's not worth fixing, but wanted to note it

Screenshot 2024-07-29 at 5 37 12 PM

Screenshot 2024-07-29 at 5 39 41 PM

Screenshot 2024-07-29 at 5 37 04 PM

dev-client/src/components/menus/MenuItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+    } else if (variant === 'destructive') {
+      return styles.destructive;
+    } else {
+      return styles.default;
+    }
+  }, [disabled, variant]);
+
+  return (
+    <TouchableRipple
+      onPress={onPress!}
```
knipec, line 63: Question: What guarantees `onPress` being non-null / non-undefined, since it's marked optional in the props?

dev-client/src/components/menus/MenuItem.tsx
knipec: Added an article to the reading group list to start thinking about this :)

dev-client/src/components/menus/MenuItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+        )}
+      </View>
+    </TouchableRipple>
+  );
+}
+
+const styles = StyleSheet.create({
+  root: {
+    flexDirection: 'row',
+    margin: 16,
```
knipec, line 111: Side question: Do you envision moving away from the standardized sizing in theme.ts, and do them in this one-off per-component way? I feel like it's nice to have some set of standards but don't have takes on how (if theme.ts is a nativebase concept)
Was thinking about this because noticing this is the 'space.md' size (which is what we use in the ScreenContentSection, which has the same flavor of "just keep content 16px from the edge of the screen")

dev-client/src/components/menus/MenuList.tsx
```diff
@@ -16,15 +16,14 @@
...
-
 export type ButtonListProps = React.PropsWithChildren;
 
-export function ButtonList({children}: ButtonListProps) {
+export function MenuList({children}: ButtonListProps) {
   return (
-    <Column mt="12px" mb="24px" space="6px">
+    <View style={styles.root} accessibilityRole="menu">
       <Divider />
       {Children.map(children, child => {
```
knipec, line 28: 1. First I was wondering: Do you not need a `key` for the Divider and maybe also children?

2. Then I googled and haven't read thru this, but seems like React docs discourage use of the `Children` API: https://react.dev/reference/react/Children#alternatives

dev-client/src/components/menus/MenuItem.tsx
tm-ruxandra: I think we could eventually have constants in place for these concepts (like we're still pulling in `theme.ts` for colors) but for now I can't think of a better solution. I think it can be something to think about as we continue implementing component stuff.

dev-client/src/components/menus/MenuList.tsx
tm-ruxandra: Interesting - will take a look.

dev-client/src/components/inputs/Select.tsx
tm-ruxandra: It might be a side-effect of how background colors are handled here. I can try to fix it via that, otherwise I'd ask for you to note this in https://github.com/techmatters/terraso-mobile-client/issues/1774

dev-client/src/components/inputs/Select.tsx
tm-ruxandra: After thinking about it, you raise a good point here. I'll see if I can just require that client code provides unique keys for select items.

dev-client/src/components/menus/MenuItem.tsx
paulschreiber: React Native Paper also supports theming (but differently than NativeBase). We have a little bit of RNP theming in `theme.ts` as well.

question: would it make sense to use or extend the ReactNativePaper `Menu.Item` here?

dev-client/src/components/inputs/Select.tsx
tm-ruxandra: Seems like adding a background color fixed it.

dev-client/src/components/menus/MenuList.tsx
tm-ruxandra: After reading the above, I've removed the use of `Children` in the list component.

dev-client/src/screens/ProjectTeamScreen/components/RolePill.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {Badge} from 'terraso-mobile-client/components/NativeBaseAdapters';
+
+type RolePillProps = {
+  membership: ProjectMembership;
+};
+
+export const RolePill = ({membership}: RolePillProps) => {
+  const {t} = useTranslation();
+  return (
+    <Badge
```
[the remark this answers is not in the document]
paulschreiber: I've also asked @CourtneyLee333 if it make sense to call this Chip in the Component spec, since that's what Material UI calls it. (Apple doesn't have an equivalent item; Token is closest.)

dev-client/src/components/menus/MenuItem.tsx
```diff
@@ -0,0 +1,167 @@
...
+import {convertColorProp} from 'terraso-mobile-client/components/util/nativeBaseAdapters';
+
+export type MenuItemVariant = 'default' | 'destructive';
+
+export type MenuItemProps = {
+  variant?: MenuItemVariant;
+  uppercase?: boolean;
+  label: string;
+  subLabel?: string;
+  icon?: IconName | React.ReactElement;
```
paulschreiber, line 32: question: do we an to allow any kind of ReactElement, or should we restrict this to some kind of Icon component?

dev-client/src/components/menus/MenuItem.tsx
```diff
@@ -0,0 +1,167 @@
...
+  pill?: React.ReactNode;
+  selected?: boolean;
+  disabled?: boolean;
+  onPress?: PressableProps['onPress'];
+};
+
+export function MenuItem({
```
paulschreiber, line 39: question: should this be `export const MenuItem = ...`?

dev-client/src/screens/ProjectTeamScreen/components/RolePill.tsx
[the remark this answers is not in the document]
tm-ruxandra: We have a separate ticket to standardize chips / pills / whatever in https://github.com/techmatters/terraso-mobile-client/issues/1877

dev-client/src/components/menus/MenuItem.tsx
tm-ruxandra: The current non-icon-name example is a `ProfilePic` extracted from the user info component, which made me think it's fine to just be an arbitrary component. Coming up with a taxonomy to encompass both icons and profile pictures seemed excessive.

dev-client/src/components/menus/MenuItem.tsx
tm-ruxandra: Yes.

dev-client/src/components/content/images/ProfilePic.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {Image} from 'native-base';
```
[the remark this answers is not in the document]
tm-ruxandra: Can do.

dev-client/src/components/menus/MenuItem.tsx
tm-ruxandra: On the whole, my opinion is that a unified strategy for theming post-NativeBase is a large-scale task of its own and is out-of-scope here. But we should keep it in mind as we implement more components; might be a later-stage task as our component library matures.

knipec (approved):
Nice! Great to be able to get rid of the nativebase Image too :)

PR #1887, opened 2024-07-31
docs: Add a github issue template that sets project = LandPKS

knipec:
Occasionally we'd file a mobile-client issue and forget to add it to the LandPKS project, which means it doesn't show up in our backlog views. 

Side question: Is there a way to confirm this works before merging it?


paulschreiber:
What about adding an automation to put all mobile-client issues into the project? Would that also meet your needs?

knipec:
> What about adding an automation to put all mobile-client issues into the project? Would that also meet your needs?

Yup, done in 1:1! :)

Issue #1808, opened 2024-07-19
Prompt for location permission appears on signin or welcome screens instead of on map screen 


paulschreiber:
LandPKS requests permission to access locations in the signin or welcome screen, but we expect this to happen when the map screen is first viewed. 

## Version
175 beta

[x] Android
[x] iOS

## Steps to reproduce
1. Uninstall and reinstall app
2. Open the app

## Expected behavior
Prompt for location permission on map screen (not login or welcome screens)

## Actual Behavior
Prompt for location on signin or welcome screen

knipec:
Should be available to verify in build 178

paulschreiber:
Verified as fixed.

PR #1892, opened 2024-08-01
fix: Ask for location permissions on home screen after login

knipec:
## Description
App used to do location stuff in RootNavigator, resulting in asking user for location permission on the welcome screen. Now do the same stuff, but instead on the BottomTabsScreen (which would ask for permissions effectively on the home screen after login)

Relatedly: I don't think _everything_ you could do to `locationManager` implicitly launches a permissions check -- like you can do locationManager.setMinDisplacement and it doesn't -- but it seems like stuff to do with using the user's location does.

**Side note**: There is still this warning before user has given permissions (was happening from RootNavigator code, now in BottomTabsScreen because this change moves the code there), but that seems like an orthogonal concern:
Screenshot 2024-08-01 at 10 04 39 AM

### Related Issues
Addresses #1808 

### Verification steps
1. Uninstall and reinstall app
2. Open the app (shouldn't see location permission prompt)
3. Get past the welcome screen and log in (now should see location permission prompt)

### DIFF
--- a/dev-client/src/navigation/navigators/RootNavigator.tsx
+++ b/dev-client/src/navigation/navigators/RootNavigator.tsx
@@ -18,16 +18,13 @@
 import {useEffect} from 'react';
 
 import {NativeStackNavigationOptions} from '@react-navigation/native-stack';
-import {Location, locationManager} from '@rnmapbox/maps';
 
 import {
   fetchUser,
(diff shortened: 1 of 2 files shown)


paulschreiber (approved):
This is great. Thanks!

Issue #1350, opened 2024-05-16
Frontend should hide site data entry for Viewers


knipec:
## Repro: 
1. Create a Project with a Site. 
2. Add another user (who you can log in as) to the project as a Viewer
3. Log out and log in as that other user from the previous step
4. Go to the site and one of the data entry screens (like slope > steepness)
**Expected**: No UI that makes me think I can actually input data. 
**Actual**: A "Done" button. A button for manual entry that I can put numbers into but doesn't work. A slope meter button that... well on my machine it causes a crash but maybe that's just the state of my branch?

## Additional info:
This is not as egregious because the permissions checks on the backend will prevent the changes from making it to the database. Apparently this also prevents selections from being reflected in the app UI, so that's nice behavior, but in a kinda hacky way.

But I shouldn't see editing-specific UI if I am in a role that cannot make edits. I believe this is the full list of such UI elements in the Site tabs that should be altered or removed for Viewers:
- Slope > Steepness: (1) "Done" button, (2) "Manual" button, (3) "Slope meter" button
- Slope > Shape: (1) "Done" button
- Soil: (1) ... menu for each depth interval
- Soil > Vertical Cracks: (1) dropdown, (2) grayed out "Done" button
- Soil > Texture/Rock fragment: (1) Texture dropdown, (2) "Use guide" button?
- Soil > Color: (1) "Enter values" button, (2) "Use guide" button, (3) button to take or upload a photo
- Notes: (1) pencil icon to edit a site note
- Also, something about the interaction between the frontend and backend seems to handle the Pressable images in the Rock Fragment, Slope Steepness, and Shape screens. But would be nice to have frontend code that makes it explicit.

paulschreiber:
Slope Steepness fixed with https://github.com/techmatters/terraso-mobile-client/pull/1834

DerekCaelin:
This appears resolved!

PR #1923, opened 2024-08-05
fix: remove site data entry for viewers

paulschreiber:
## Description
Remove site data entry for viewers
- for soil color, hide delete button and done button from viewers
- hide texture menu, use guide button and done button from viewers
- hide vertical cracks menu and done button from viewers
- hide slope shape done button for viewers
- disable steepness, shape and texture ImageRadio buttons for viewers

also:
- avoid key error when a single label is blank/undefined
- style disabled `<Select>` item

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/1350

### DIFF
--- a/dev-client/src/model/permissions/permissions.ts
+++ b/dev-client/src/model/permissions/permissions.ts
@@ -15,6 +15,7 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
+import {ProjectMembershipProjectRoleChoices} from 'terraso-client-shared/graphqlSchema/graphql';
 import {SiteUserRole} from 'terraso-client-shared/selectors';
 
 export const matchesOne =
@@ -36,6 +37,11 @@ export const SITE_EDITOR_ROLES: SiteUserRole[] = [
   {kind: 'project', role: 'MANAGER'},
   {kind: 'project', role: 'CONTRIBUTOR'},
   {kind: 'site', role: 'OWNER'},
+];
+
+export const PROJECT_EDITOR_ROLES: ProjectMembershipProjectRoleChoices[] = [
+  'MANAGER',
+  'CONTRIBUTOR',
 ];
 
 export const isProjectViewer = (userRole: SiteUserRole | null) => {
--- a/dev-client/src/model/permissions/permissions.ts
+++ b/dev-client/src/model/permissions/permissions.ts
@@ -32,3 +32,18 @@ export const isSiteManager = matchesRole([
   {kind: 'project', role: 'MANAGER'},
 ]);
 
+export const SITE_EDITOR_ROLES: SiteUserRole[] = [
+  {kind: 'project', role: 'MANAGER'},
+  {kind: 'project', role: 'CONTRIBUTOR'},
+  {kind: 'site', role: 'OWNER'},
+];
+
+export const isProjectViewer = (userRole: SiteUserRole | null) => {
+  return Boolean(userRole && ['VIEWER'].includes(userRole.role));
+};
+
+export const isProjectEditor = (userRole: SiteUserRole | null) => {
+  return Boolean(
+    userRole && ['MANAGER', 'CONTRIBUTOR', 'OWNER'].includes(userRole.role),
+  );
+};
--- a/dev-client/src/model/permissions/permissions.ts
+++ b/dev-client/src/model/permissions/permissions.ts
@@ -1,0 +1,34 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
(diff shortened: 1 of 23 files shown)


dev-client/src/screens/SoilScreen/ColorScreen/ColorScreen.tsx
```diff
@@ -66,6 +71,16 @@ export const ColorScreen = (props: SoilPitInputScreenProps) => {
...
 
+  const userRole = useSelector(state =>
+    selectUserRoleSite(state, props.siteId),
+  );
+
+  const canDelete = useMemo(
+    () =>
+      userRole && ['MANAGER', 'CONTRIBUTOR', 'OWNER'].includes(userRole.role),
+    [userRole],
+  );
```
paulschreiber, line 82: `useSiteRoleContext()` couldn't be used here because site wasn't in scope. The site ID is in `props.`siteId. Would another approach be better?

tm-ruxandra: Could you extract this & the similar logic you've added in other places to a helper method somewhere in the `permissions` directory? I think it's fine if the site context isn't available but we can at least make this logic more centralized.

dev-client/src/screens/SoilScreen/ColorScreen/ColorScreen.tsx
```diff
@@ -85,43 +100,63 @@ export const ColorScreen = (props: SoilPitInputScreenProps) => {
...
+                  )}
+                />
+                <Paragraph variant="body1">{t('soil.color.info.p2')}</Paragraph>
+                <Paragraph variant="body1">{t('soil.color.info.p3')}</Paragraph>
+              </InfoOverlaySheetButton>
+            </Row>
+            <Box flex={1} />
+            <RestrictBySiteRole
+              role={[
+                {kind: 'project', role: 'MANAGER'},
```
tm-ruxandra, line 125: Might be nice to extract this specific combination of parameters to a named constant somewhere, in a dedicated permissions file/folder as noted above.

dev-client/src/screens/SoilScreen/TextureScreen.tsx
```diff
@@ -140,54 +152,78 @@ export const TextureScreen = (props: SoilPitInputScreenProps) => {
...
+              </InfoOverlaySheetButton>
+            </Row>
+            <RestrictBySiteRole
+              role={[
+                {kind: 'project', role: 'MANAGER'},
+                {kind: 'project', role: 'CONTRIBUTOR'},
+                {kind: 'site', role: 'OWNER'},
+              ]}>
+              <Box height="sm" />
+              <Select
```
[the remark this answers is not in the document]
paulschreiber: Good idea. Done.

dev-client/src/model/permissions/permissions.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+import {ProjectMembershipProjectRoleChoices} from 'terraso-client-shared/graphqlSchema/graphql';
+import {SiteUserRole} from 'terraso-client-shared/selectors';
+
+export const matchesOne =
+  <T>(cmp: (a: T, b: T) => boolean) =>
+  (examples: T[]) =>
+  (sample: T) =>
+    examples.filter(x => cmp(x, sample)).length > 0;
```
tm-ruxandra, line 25: Is there a reason you didn't use `examples.find` ?

paulschreiber: I didn't write that code — it came from https://github.com/techmatters/terraso-mobile-client/commit/66b62974e. Happy to fix that as well if you want to make the change with a GItHub suggestion.

tm-ruxandra: `examples.find(x => cmp(x, sample))` would do it I think.

tm-ruxandra (approved):

dev-client/src/model/permissions/permissions.ts
knipec, line 25: Suggestion re: Ruxandra's comment -- According to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), I think it's just 
```suggestion
    examples.find(x => cmp(x, sample)) !== undefined;
```
which would make sense to me since we only expect it to match and return one, so it can return once it finds one

dev-client/src/model/permissions/permissions.ts
knipec, line 1: Nice! I like this

dev-client/src/model/permissions/permissions.ts
```diff
@@ -0,0 +1,59 @@
...
+export const matchesRole = matchesOne(
+  (a: SiteUserRole, b: SiteUserRole) => a.kind === b.kind && a.role === b.role,
+);
+
+export const isSiteManager = matchesRole([
+  {kind: 'site', role: 'OWNER'},
+  {kind: 'project', role: 'MANAGER'},
+]);
+
+export const SITE_EDITOR_ROLES: SiteUserRole[] = [
```
knipec, line 36: Nitpick: Can Contributors only edit soil data on a site, but not stuff like the name of a site? If so, is there a slightly more accurate name? Won't block approval on this though :)

dev-client/src/screens/SoilScreen/components/SoilSurfaceScreen.tsx
```diff
@@ -57,30 +60,37 @@ export const SoilSurfaceScreen = ({siteId}: Props) => {
...
-      </Column>
-      <Box position="absolute" bottom="0" right="0">
-        <DoneButton isDisabled={!cracking} />
-      </Box>
+      <SiteRoleContextProvider siteId={siteId}>
+        <Column padding="md">
+          <Heading variant="h6">
+            {t('soil.collection_method.verticalCracking')}
+          </Heading>
+          <RestrictBySiteRole role={SITE_EDITOR_ROLES}>
```
knipec, line 68: Suggestion: Also disable this instead of hiding it?

Issue #1352, opened 2024-05-17
Color input photo module returns nonvalid results for 7YR 3/1 sample


DerekCaelin:
### App Version

125

### Platforms

- [X] iOS
- [x] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

With the same or similar cropping of an image, color input photo module will return different results.

### Steps to Reproduce

1. Download [test color image 7.5YR ](https://drive.google.com/file/d/1TKOg9S9_bsyGhlZ_vHeTH0dThfxUngpT/view?usp=drive_link) onto testing device
2. Go to a site where you are a manager or contributor
3. Select to the "soil" tab and select a color input
4. Select test photo
5. Crop to reference image
6. Crop to 3/1 leaving no white space
7. Select "moist" and "yes" for radio buttons
8. Select analyze
9. see issue

#### Cropping on iPad Image
![image](https://github.com/techmatters/terraso-mobile-client/assets/10799815/240d2c06-dbb0-416b-a966-6015256e84e2)

### Expected behavior

Return results within 1 step of 7.5YR 3/1. 

### Actual behavior

Returns a range of incorrect values, including:
- 5P 8.5/3
- 5R 2/1
- 5PB 5/1
- 10PB 9.5/3
- 10YR 2/1

![image](https://github.com/techmatters/terraso-mobile-client/assets/10799815/b0d03efa-1608-4c2a-8ec3-f336ce99fcd5)
![image](https://github.com/techmatters/terraso-mobile-client/assets/10799815/3c0d4b61-fa84-48b9-983d-4be8dac387aa)
![image](https://github.com/techmatters/terraso-mobile-client/assets/10799815/29d52c82-5c37-4890-a530-5998c3f53556)

### Additional context

Following the same steps on Android the test returns a valid result. On android, if I leave marginal whitespace in the crop zone, the invalid result (5PB 5/1) occurs.

CourtneyLee333:
Updates: I just did some more testing on iPad and got results all over the place. I used the same 7YR 3/1 sample, and experimented with the radio buttons for Moist and Light (though that didn't seem to affect a pattern I could discern.) I used a single depth at each of two different sites. I used the same cropping on all of them. [See the results in Slack.](https://tech-matters.slack.com/archives/C06BQTCAUJF/p1716382018087999)

shrouxm:
[...]

PR #1936, opened 2024-08-08
fix: ios color bug

shrouxm:
## Description
Please review by commit. The first two are no-op refactors. The last commit tweaks the way we're calling the color quantization algorithm to be more robust to unexpected pixels on the edges of images. [A bug on iOS](https://github.com/techmatters/terraso-mobile-client/issues/1352) reliably produces such pixels, but we'd also like to be robust in these cases regardless in case the user gives us a sloppy croppy.

I tested this on my iphone and can no longer reproduce #1352, but if someone else could also attempt to reproduce that would bring me peace of mind.

### Checklist

### Related Issues
https://github.com/techmatters/terraso-mobile-client/issues/1352

### Verification steps
Should no longer be able to reproduce https://github.com/techmatters/terraso-mobile-client/issues/1352

### DIFF
--- a/dev-client/src/model/color/colorDetection.ts
+++ b/dev-client/src/model/color/colorDetection.ts
@@ -32,7 +32,7 @@ import {
 } from 'terraso-mobile-client/model/color/types';
 
 const SOIL_COLOR_SIMILARITY_THRESHOLD = 5;
-const COLOR_COUNT = 16;
+const COLOR_COUNT = 5;
 
 // Correction values obtain from spectrophotometer Observer. = 2°, Illuminant = D65
 export const REFERENCES = {
@@ -108,12 +108,17 @@ export const dominantColor = (pixels: RGBA[]): RGB => {
   });
 
   // Quantize.js performs median cut algorithm, and returns a palette of the "top 16" colors in the picture
-  var cmap = quantize(pixelArray, COLOR_COUNT);
-  if (cmap === false) {
+  var colorMap = quantize(pixelArray, COLOR_COUNT);
+  if (colorMap === false) {
     throw new Error('Unexpected color algorithm failure!');
   }
 
-  return cmap.palette()[0];
+  const colorsWithCounts = colorMap.vboxes.map(
+    vbox => [vbox.color, vbox.vbox.count()] as const,
+  );
+  const [dominantColor] = colorsWithCounts.sort(([, c1], [, c2]) => c2 - c1)[0];
+
+  return dominantColor;
 };
 
 const nearestSoilColor = (color: MunsellHVC) =>
--- a/dev-client/src/model/color/colorDetection.ts
+++ b/dev-client/src/model/color/colorDetection.ts
@@ -34,6 +34,7 @@ import {
 const SOIL_COLOR_SIMILARITY_THRESHOLD = 5;
 const COLOR_COUNT = 16;
 
+// Correction values obtain from spectrophotometer Observer. = 2°, Illuminant = D65
 export const REFERENCES = {
   CAMERA_TRAX: [210.15, 213.95, 218.42],
   CANARY_POST_IT: [249.92, 242.07, 161.42],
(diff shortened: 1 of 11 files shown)


dev-client/src/model/color/colorDetection.ts
```diff
@@ -107,12 +73,55 @@ export const getColor = (
...
+        return [[r, g, b] as quantize.RgbPixel];
+      }
+    }
+    return [];
+  });
+
+  // Quantize.js performs median cut algorithm, and returns a palette of the "top 16" colors in the picture
+  var cmap = quantize(pixelArray, COLOR_COUNT);
+  if (cmap === false) {
+    throw new Error('Unexpected color algorithm failure!');
```
paulschreiber, line 113: suggestion: log some data (pixel array?) to sentry.

dev-client/src/model/color/types.ts
```diff
@@ -0,0 +1,39 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+export type RGBA = [number, number, number, number];
+export type RGB = [number, number, number];
```
paulschreiber, line 19: question: is it worth creating a RGBValue type that is a subset of number, (i.e. 0-255)? https://stackoverflow.com/questions/39494689/is-it-possible-to-restrict-number-to-a-certain-range

dev-client/src/model/color/colorDetection.ts
```diff
@@ -0,0 +1,146 @@
...
+import {SOIL_COLORS} from 'terraso-mobile-client/model/color/soilColors';
+import {
+  ColorResult,
+  MunsellHVC,
+  RGB,
+  RGBA,
+} from 'terraso-mobile-client/model/color/types';
+
+const SOIL_COLOR_SIMILARITY_THRESHOLD = 5;
+const COLOR_COUNT = 5;
```
paulschreiber, line 35: suggestion: explain how these numbers are derived.

dev-client/src/model/color/colorDetection.ts
```diff
@@ -34,6 +34,7 @@ import {
 const SOIL_COLOR_SIMILARITY_THRESHOLD = 5;
 const COLOR_COUNT = 16;
 
+// Correction values obtain from spectrophotometer Observer. = 2°, Illuminant = D65
```
knipec, line 37: Not being familiar with this code, I don't understand this comment.

dev-client/src/model/color/colorDetection.ts
```diff
@@ -108,12 +108,17 @@ export const dominantColor = (pixels: RGBA[]): RGB => {
   });
 
   // Quantize.js performs median cut algorithm, and returns a palette of the "top 16" colors in the picture
```
knipec, line 110: Nit: no longer "top 16"

dev-client/src/model/color/colorDetection.ts
```diff
@@ -108,12 +108,17 @@ export const dominantColor = (pixels: RGBA[]): RGB => {
...
     throw new Error('Unexpected color algorithm failure!');
   }
 
-  return cmap.palette()[0];
+  const colorsWithCounts = colorMap.vboxes.map(
+    vbox => [vbox.color, vbox.vbox.count()] as const,
+  );
+  const [dominantColor] = colorsWithCounts.sort(([, c1], [, c2]) => c2 - c1)[0];
+
+  return dominantColor;
```
knipec: Could extract this into a testable function like `getDominantColor` or something. 

Also, out of curiosity, what's a vbox?

dev-client/src/model/color/colorDetection.ts
knipec, line 1: I might need a bit of explanation/reminder to understand why this fixes the bug -- like, were the quantized colors not sorted by count of "pixels in original image closest to that color"? Is the color_count part of the fix, or just an optimization?

dev-client/src/model/color/types.ts
shrouxm: no i don't think so:
- we're never manually constructing RGBs so there wouldn't be any situation where that could actually validate anything
- it would make the definitions much more difficult to understand
- it could be a typechecker performance papercut which could add up if we did things like that too often

dev-client/src/model/color/colorDetection.ts
shrouxm: yeah, there seems to be a bug in the quantization library where under obscure conditions that are coincidentally reliably triggered by the iOS image decoding bug it will return the pixels not in order of the most common. the color count change was from some ad-hoc testing I was doing, the original value of 16 was passed down from ancestral code and I never understood, and while testing different ways to fix this bug I tried tweaking it and 5 seemed to be more robust. (i'm aware that this is all very loosey-goosey :sob: but i don't know how to do better without spending a ton of time on this, i feel like implementing a domain-specific color detection algorithm could be someone's whole master's thesis)

dev-client/src/model/color/colorDetection.ts
shrouxm: i don't fully understand, the juicy details are explained in [this paper](http://leptonica.org/papers/mediancut.pdf), but vaguely speaking it's a range of pixel values with associated frequency count

knipec (approved):

Issue #1984, opened 2024-08-15
delete account confirmation window: missing ")"


DerekCaelin:
### App Version

181 beta

### Platforms

- [ ] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

On the "Delete account screen", there is a missing parenthesis in the instructional text

### Steps to Reproduce

1. sign in
2. go to settings screen
3. tap "delete account"
4. see issue

### Expected behavior

Final parenthesis needed in the instructional text.

### Actual behavior

Parenthesis is missing:

<img src="https://github.com/user-attachments/assets/ac96311c-795d-4f94-8839-c1b7e3adc0b6" width="300px">

DerekCaelin:
This is resolved!

PR #2014, opened 2024-08-19
fix: add missing parenthesis to delete dialog text

paulschreiber:
## Description
Add missing parenthesis to delete dialog text.

Addresses https://github.com/techmatters/terraso-mobile-client/issues/1984


paulschreiber:
I didn't wait for triage on this because it was only a string change.

knipec (approved):

Issue #1852, opened 2024-07-25
Create unit tests for Soil ID ranking logic


tm-ruxandra:
Build on the unit testing infrastructure introduced in https://github.com/techmatters/terraso-product/issues/776 to add tests for the logic in`model/soilId`.

DerekCaelin:
Marked done Aug 2024

PR #2016, opened 2024-08-21
Test: Create unit tests for soil ID ranking logic

knipec:
## Description
Add tests for getTopMatch and sorting, for location-based and data-based soil matches.

Note: we expect the input to come with the correct ranks (this test, as well as the product code, will not detect if a set of matches' `rank` and `score` contradict each other; it will just use `rank` and ignore `score`)

### Related Issues
Implements #1852 

### Verification steps
Run `npm test -- soilIdRanking.test.tsx`

### DIFF
--- a/dev-client/src/model/soilId/soilIdRanking.test.tsx
+++ b/dev-client/src/model/soilId/soilIdRanking.test.tsx
@@ -1,0 +1,298 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {
+  getSortedDataBasedMatches,
+  getSortedLocationBasedMatches,
+  getTopMatch,
+} from 'terraso-mobile-client/model/soilId/soilIdRanking';
+
+const locationBasedSoil0 = {
+  dataSource: 'Data source 0',
+  distanceToNearestMapUnitM: 5,
+  match: {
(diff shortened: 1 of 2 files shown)


dev-client/src/model/soilId/soilIdRanking.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {
+  getSortedDataBasedMatches,
+  getSortedLocationBasedMatches,
+  getTopMatch,
+} from 'terraso-mobile-client/model/soilId/soilIdRanking';
+
+const locationBasedSoil0 = {
```
knipec, line 24: No harm to defining these outside the scope of `describe`, right?

dev-client/src/model/soilId/soilIdRanking.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+      taxonomySubgroup: 'Taxonomy subgroup 2',
+    },
+  },
+};
+
+describe('location based matches', () => {
+  test('get top match by rank', () => {
+    const inputSoilIdResults = {
+      locationBasedMatches: [
+        locationBasedSoil0,
```
tm-ruxandra, line 192: This test defines these large constants for test data, but most of the tests are only sensitive to the `rank` property. I think that this suite would be easier to understand if it got test data from a helper factory method like `locationBasedMatchWithRank(rank: number)`. This is both to factor out unnecessary duplication (the other properties in the object basically don't matter except for identifying uniqueness) and to enhance readability (it would be obvious in the unit test that the matches in question have specific ranks that affect the test; right now, I have to scroll to the top of the file and scan three large constants in each test to know what each one's rank is).

dev-client/src/model/soilId/soilIdRanking.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+        locationBasedSoil1,
+        locationBasedSoil2,
+      ],
+      dataBasedMatches: [],
+    };
+    expect(getTopMatch(inputSoilIdResults)).toEqual(locationBasedSoil0);
+  });
+
+  test('get top match by rank when reordered', () => {
```
tm-ruxandra, line 201: I think this test is redundant with the previous one; it's fine IMO to just have a single test verifying that out-of-order elements still get sorted/picked correctly.

tm-ruxandra (changes requested):
The tests themselves are fine; had some comments about structure/redundancy of setup code.

tm-ruxandra (approved):
Looks great!

PR #2063, opened 2024-08-28
fix: update sentry configuration

paulschreiber:
## Description
Update sentry metro and app config.

### DIFF
--- a/dev-client/app.config.ts
+++ b/dev-client/app.config.ts
@@ -79,7 +79,6 @@ if (typeof APP_BUILD === 'string') {
 }
 
 const defaultConfig: ExpoConfig = {
-  ...config,
   name: 'LandPKS Soil ID',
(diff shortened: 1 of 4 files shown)


knipec:
I don't have a lot of context on this one, so unless other folks want to review this, Paul could you walk me through it on a call when you're back from vacation?

knipec (approved):
Sounds like we really don't know very deeply what this does, but talked to Paul about context for this change: 
When we updated to a new Sentry version, we couldn’t get it to build with iOS. We asked Sentry and they told us to not use the serializer we'd been using and instead use the default Expo serializer, which is what drove #2018. With that interaction, the Sentry folks also provided the changes to the metro.config.js file, and recommended we use that. Their documentation also instructed us to make the changes here to app.config.ts, which someone from sentry approved in conversation with Paul.

PR #2071, opened 2024-08-29
feat: add permissions unit tests

### LINKED ISSUE #1853: Add unit tests for permission hooks code
Build on the unit testing infrastructure introduced in https://github.com/techmatters/terraso-product/issues/776 to add tests for the logic in `hooks/permissionHooks.` The core filter logic will need to be extracted from the React selector/callback infrastructure.

paulschreiber:
## Description
Extract `userHasProjectRole` in to a function and refactor `permissionHooks` to use this function.

Add tests for `userHasProjectRole`.

### Related Issues
Fixes #1853.

### DIFF
--- a/dev-client/src/model/permissions/permissions.test.ts
+++ b/dev-client/src/model/permissions/permissions.test.ts
@@ -1,0 +1,109 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {User} from 'terraso-client-shared/account/accountSlice';
+import {Project, ProjectRole} from 'terraso-client-shared/project/projectSlice';
+
+import {userHasProjectRole} from 'terraso-mobile-client/model/permissions/permissions';
+
+const emptyUser = null;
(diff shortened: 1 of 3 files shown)


dev-client/src/model/permissions/permissions.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  profileImage: '',
+  preferences: {
+    group_notifications: 'true',
+    story_map_notifications: 'true',
+    language: 'en-US',
+  },
+};
+
+const emptyProject: Project = {
+  id: '3',
```
tm-ruxandra, line 40: Can we condense these project objects to a helper factory method? It's difficult to parse many large objects with fields that aren't all relevant to the test, I'd rather see like `const sampleProject = projectWithMemberships(...);` or something like that.

dev-client/src/model/permissions/permissions.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    expect(result).toBeFalsy();
+  });
+
+  test('user is in project with no role', () => {
+    const result = userHasProjectRole(sampleUser, sampleProject);
+
+    expect(result).toBeTruthy();
+  });
+
+  test('user is in project with role', () => {
```
tm-ruxandra, line 90: I would add another case for 'user is in project with different role than expected', e.g. the filter is for `MANAGER` but the user is only `VIEWER`.

tm-ruxandra (changes requested):
Generally looks good, some changes requested on test structure.

dev-client/src/model/permissions/permissions.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    sites: {},
+    archived: false,
+  };
+};
+
+const sampleProject = projectWithMembersjips({
+  sample: {userId: '1', userRole: 'MANAGER', id: '2'},
+});
+
+const sampleProject2 = projectWithMembersjips({
```
tm-ruxandra, line 75: 1) Let's move these const declarations to the test functions themselves, so that it's easier to see the specific memberships relevant to the test. `sampleProject` and `sampleProject2` still don't communicate much about the attributes under the projects under test, I'd rather they just be inline in the tests.
2) Typo in method name, should be `projectWithMemberships`

tm-ruxandra (approved):

PR #2072, opened 2024-08-29
fix: use ExternalLink for soil info pages

paulschreiber:
## Description
Use ExternalLink for soil info pages. 

Also clean up external link to use a non-default export.

### Related Issues
Addresses #1023.

### DIFF
--- a/dev-client/src/components/links/InternalLink.tsx
+++ b/dev-client/src/components/links/InternalLink.tsx
@@ -29,7 +29,7 @@ type InternalLinkProps = {
   url: string;
 };
 
-export function InternalLink({label, onPress, url}: InternalLinkProps) {
+export const InternalLink = ({label, onPress, url}: InternalLinkProps) => {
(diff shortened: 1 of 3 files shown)


dev-client/src/components/links/InternalLink.tsx
```diff
@@ -29,7 +29,7 @@ type InternalLinkProps = {
   url: string;
 };
 
-export default function InternalLink({label, onPress, url}: InternalLinkProps) {
+export function InternalLink({label, onPress, url}: InternalLinkProps) {
```
tm-ruxandra, line 32: Probably the true standard declaration for our other components would be `export const InternalLink = (` rather than `export function`.

tm-ruxandra (changes requested):
Looks good overall; minor change on the updated component declaration since we're making the change here.

tm-ruxandra (approved):

PR #2074, opened 2024-08-29
fix: add trailing slash to graphQLEndpoint

paulschreiber:
## Description
Add trailing slash to graphQLEndpoint.


dev-client/jest/integration/setup.ts
knipec, line 1: I assume there were no consequences to this, just a little clean-up thing because it's recommended to have trailing slashes for django endpoints?

paulschreiber: Yeah. Django needs the slash and was complaining (in some other cases).

knipec (approved):

Issue #2070, opened 2024-08-29
Move soilID Redux slice into mobile client codebase


tm-ruxandra:
Currently, the Redux slice the handles soil sample data and Soil ID interactions lives in `terraso-client-shared`. This causes architectural challenges for mobile-only considerations for offline mode operation. Migrate this code (Redux slices under `project`, `site`, and `soilId`)  into the `terraso-mobile-client` repository, and apply minor refactoring to make further offline research tasks more manageable.

shrouxm:
closing because i think this is complete

PR #2101, opened 2024-09-03
refactor: move mobile-only redux slice code out of shared repo

tm-ruxandra:
## Description
The `soilId`, `site`, and `project` slices are only used by the mobile application, and offline support work will require them to use data flows that do not apply to the shared mobile/web codebase. (For example, redirecting reads/writes to local storage rather than the network.)

Migrate relevant code from `terraso-client-shared`, with the following fixups:
- The `createStore` method has been modified to have correct type safety.
- The existing `selectors.ts` file's tests are now categorized as integration tests since they depend on rendering the full app stack.
- Minor module reorganization for better readability.

### Checklist
- [x] New tests added

### Related Issues
#2070

### Verification steps

Run both application and tests and verify that both function as expected.


dev-client/src/store/index.ts
```diff
@@ -21,28 +21,44 @@ import {
...
-export type AppState = StateFromStoreFactory<typeof createStore> &
-  StateFromReducersMapObject<typeof reducers>;
+export type AppState = StateFromReducersMapObject<typeof reducers>;
 export type AppDispatch = DispatchFromStoreFactory<typeof createStore>;
 
 export const useSelector: TypedUseSelectorHook<AppState> = reduxUseSelector;
 export const useDispatch: () => AppDispatch = reduxUseDispatch;
 
-export const createStore = createStoreFactory(reducers);
+export const createStore = (intialState?: Partial<AppState>) =>
```
tm-ruxandra, line 58: Bypassing the `createStoreFactory` method allows us to have correct types for the resulting store object; see the TODO on that method's implementation in the shared repo.

dev-client/src/screens/SiteNotesScreen/SiteNotesScreen.tsx
```diff
@@ -36,14 +34,6 @@ import {SiteInstructionsCard} from 'terraso-mobile-client/screens/SiteNotesScree
 import {SiteNoteCard} from 'terraso-mobile-client/screens/SiteNotesScreen/components/SiteNoteCard';
 import {useSelector} from 'terraso-mobile-client/store';
 
-export type SiteNote = {
```
tm-ruxandra, line 39: AFAICT this type was never used anywhere.

shrouxm (approved):
LGTM, i agree with all the code organization decisions you made. it's hard to follow changes in files that are being moved across repos so just to be sure, is the following accurate?
- `projectSlice.ts` and `siteSlice.ts` were copied mostly unmodified, other than moving some types into associated `types.ts` files in `client-shared`
- `soilIdFunctions.ts`, `soilDataConstants.ts`, `soilIdHooks.ts`, `soilIdSlice.ts`, `selectors.ts` were copied unmodified other than import rewrites

tm-ruxandra:
@shrouxm That's correct, I split out the type declarations from the slices but otherwise all code was moved over only with fixups to import declarations / lint errors.

Issue #1893, opened 2024-08-01
Dev console warning: uid does not have location permissions


knipec:
## Repro steps
1. Uninstall and reinstall app
2. Open the app
3. Get past the welcome screen and log in (now should see location permission prompt)
  **Expected**: Just the permission prompt with no problems
  **Actual**:
Console warning: "Error: uid 10195 does not have any of [android.permission.ACCESS_FINE_LOCATION, android.permission.ACCESS_COARSE_LOCATION]"

![Screenshot 2024-08-01 at 10 04 39 AM](https://github.com/user-attachments/assets/c6493fdf-9fe2-4a03-b8de-5869752b77c3)

Points to `getLastKnownLocaation()`, which does 
```
      try {
        lastKnownLocation =
          await MapboxGLLocationManager.getLastKnownLocation();
      } catch (error) {
        console.warn('locationManager Error: ', error);
      }
```

## Thoughts
- In `BottomTabsScreen`, should we check before calling getLastKnownLocation if the user has granted permissions?
- Mapbox docs mention [PermissionsManager](https://docs.mapbox.com/android/maps/guides/user-location/permissions/), is that appropriate to use here?

Issue #1780, opened 2024-07-16
on Android, user can't see own location on first load of map


DerekCaelin:
### App Version

169

### Platforms

- [No] iOS
- [YES] Android
- [No] iOS simulator
- [No] Android emulator

### Description

First-time Android users can't see their own position on the map, and the "my location" button fails silently. Reloading the app once permissions have been granted allows the map to see the user's location. 

### Steps to Reproduce

On Android:
1. Uninstall, reinstall and open app
2. Pass through interstitial
3. Sign in with any identity
4. Open home screen and accept permissions request
5. See issue (fails to pan/zoom to your location)
6. Tap "my location" button
7. See issue

### Expected behavior

The map will know and center on the user's location
Tapping on "My location" button will center the map on the user's location

### Actual behavior

The user provides permission, but no point representing the user appears on the map, and the "my location" button does nothing. 

### Additional context

- Upon closing and re-opening the app, the map can pull the user's own location
- on iOS, permissions are granted on the intersitial screen, but on android, Permissions come when viewing the map. Theory: the app tries and fails to get the user's location, asks for permission, but doesn't try again to get location once it has permission.

DerekCaelin:
Possibly related to https://github.com/techmatters/terraso-mobile-client/issues/1749

DerekCaelin:
Assigning to Garo since they have dealt with permissions issues and are looking at related task #1806

DerekCaelin:
Hi @knipec I see this is in QA. The bug appears to still occur in 187. Is the fix in the current build?

knipec:
Sorry @DerekCaelin - should be in v188

DerekCaelin:
@knipec Sorry for the delay in reviewing this. After granting permissions to access my location, the app successfully zooms in to my location. The "My location" button works.

However, on first view, the blue marker signifying my location doesn't appear. I need to close and re-launch the app to see see the blue marker.

knipec:
@DerekCaelin Oh yeah, weird -- I can repro that behavior on a physical Android device, but not on the emulator or an iOS device. 
Would you be ok with filing a new bug for that, and pulling the new bug into October (or leave it out if you think it's lower priority than other things)?

knipec:
@DerekCaelin Hm actually I can only inconsistently repro it on a physical Android device (after deleting and re-installing the app) -- are you seeing it consistently?

DerekCaelin:
It is consistent, yes. I'll file another bug.

Issue #1749, opened 2024-07-11
bug: app permissions do not update immediately when changed in settings


shrouxm:
### App Version

1.0.0 (160)

### Platforms

- [ ] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

The expo permissions hooks do not seem to update when permissions are modified via either platform's settings page for the app. See steps to reproduce for a more detailed account. I will be filing a bug with expo so hopefully it can get fixed upstream, and tried some workarounds to fix but none that didn't create other weird behaviors so I'm giving up for now.

### Steps to Reproduce

1. Navigate to a screen that has a permissions interaction (e.g. home screen, soil color screen in the photo workflow).
2. Deny the relevant permission, or turn it off in settings.
3. Press the button that initiates the permissions interaction (e.g. location button on home screen map, "Use camera" button in photo workflow).
4. Deny the permission again if the system prompt appears until the "go to settings" modal appears.
5. Hit the "go to settings" button.
6. Enable the relevant permission.
7. Return to the app and hit the button that initiates the permissions interaction again.

### Expected behavior

The button should do what it's supposed to do since the user has now granted permission.

### Actual behavior

The button displays the "go to settings" modal again.

### Additional context

This affects:
- the home screen
- the soil color data input's photo workflow
- the slope meter screen

I tried to workaround by requesting permissions at the time the user presses the button rather than via the hook, but couldn't get that to work in a way that fits the different behaviors of all the screens this affects.

Some other possible workarounds:
- use a combination of the hook and permission requests at time of button press
- somehow use [RN's `AppState` API](https://reactnative.dev/docs/appstate) to resolve the bug downstream

Silva404:
`const [status, requestPermission, asyncStatus] = ImagePicker.useCameraPermissions();`

if you check the first status, its not updated but the asyncStatus is and after you start using the async status, the sync status is finally updated, not sure why, but if you dont request the async, the sync one is always stale.

Anyways, requesting the ImagePicker.getMediaLibraryPermissionsAsync as a fallback for the sync status worked for me, the same applies for the [ImagePicker.getCameraPermissionsAsync](https://docs.expo.dev/versions/latest/sdk/imagepicker/#imagepickergetcamerapermissionsasync)

DerekCaelin:
Might be dependent on an upstream fix, let's see if we can find a workaround. If there is an outside dependency, leave a note.

knipec:
[...]

PR #2102, opened 2024-09-05
fix: allow zooming to user location on first map view 

knipec:
**The issue** 
I was only able to repro on a Release version on a physical device, and you can’t console.log, and wasn't able to clearly figure out what series of operations caused the issue. 
- But my best guess is that it was some sort of race condition with requesting permissions in HomeScreen and getting the user’s coordinates with the locationManager in BottomTabsScreen. 
- It seemed to only reproduce after my change to fix bug #1808, which moved the locationManager code from RootNavigator to BottomTabsScreen. This would put its execution much closer to HomeScreen’s.

**The fix** 
The approach in this fix was to guarantee we will ask for permissions whenever we ask for the user’s last known location. I do this by moving the locationPermission logic from HomeScreen to BottomTabsScreen, where the locationManager logic is.
- This resolves the issue, but there may be a better solution, where the locationManager logic lives in RootNavigator again, rather than BottomTabsScreen, and only does a "get" instead of a "request" for permissions -- but still asks for location permissions when the user encounters the HomeScreen. This would help if we want to use the logic outside of BottomTabsScreen.
- But I’ve spent a while looking into the original bug (1808) and still don’t understand what was triggering the permissions request before the home screen (and I am unable to reproduce it when reverting my fix for it). 

**Aside**
There are sadly various unpleasant things going on with permissions that I did not fix (Issues #1749 and #1893). I renamed a couple things accordingly per “honest names are better than dishonest names”.

### Related Issues
Addresses #1780

### DIFF
--- a/dev-client/src/components/modals/PermissionsRequestWrapper.tsx
+++ b/dev-client/src/components/modals/PermissionsRequestWrapper.tsx
@@ -46,18 +46,21 @@ export const PermissionsRequestWrapper = ({
 }: Props) => {
   const {t} = useTranslation();
   const ref = useRef<ModalHandle>(null);
-  const [permissions, requestPermissions] = usePermissions();
+  // Issue #1808: These permissions could be outdated if
+  // a) nobody has done a "get" or "request" on them since they changed, or
+  // b) a descendant component did the "get" or "request", so there is no reason for the given component to re-render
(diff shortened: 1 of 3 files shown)


paulschreiber (commented):
Looks good to me. Will defer to Garo.

PR #2104, opened 2024-09-05
feat: form sheet component

tm-ruxandra:
## Description

Bring `FormOverlaySheet` in compliance with design spec, simplifying its implementation to remove unsupported use-cases. Specifically:
- Sheet 'done' header is now a reusable named component included in the sheet itself, rather than embedded in `Select` component
- Sheet header stays pinned while content scrolls
- Sheet now snaps to both 50% and 100% of screen
- Sheet now scrolls content with specified amount of padding at bottom

Supporting these changes:
- Move the `NavigationContainer` higher in the `App.tsx` stack to enable proper content scrolling in the sheet component when embedded in a modal.
- Also migrate the `ImagePicker` and site soil modals to use `StandaloneOverlaySheet` and `InfoSheet` respectively per discussion with design
- Extract some common logic to make the three sheet variants more concise in implementation.

### Checklist

### Related Issues
#1774

### Verification steps

Examine instances of overlay sheets and check that they still retain expected behavior. (For `Select` components, `InfoSheet` components including for site depth intervals, and for the camera/image picker for site soil colors.)


dev-client/src/components/sheets/StandaloneOverlaySheet.tsx
```diff
@@ -0,0 +1,68 @@
...
+export type StandaloneOverlaySheetProps = React.PropsWithChildren<{
+  trigger?: ModalTrigger;
+}>;
+
+/*
+ * Simple overlay sheet with no special header or sizing. Just fits to its
+ * contents, making them scrollable if needed. Intended for one-off cases
+ * where design doesn't indicate more specific sheet behavior.
+ */
+export const StandaloneOverlaySheet = forwardRef<
```
tm-ruxandra, line 44: This is functionally a one-off case that handles `ImagePicker`; it fits neither of the other two use-cases, and it seemed cleaner to encapsulate the guts of interacting with `GorhomBottomSheetModal` without implying a broader design. My discussion with @CourtneyLee333 indicated that some application components may just be these simple standalone sheets, so this component exists to serve such cases.

dev-client/src/App.tsx
```diff
@@ -91,12 +91,12 @@ function App(): React.JSX.Element {
   return (
     <GestureHandlerRootView style={style}>
       <Provider store={store}>
-        <HeaderHeightContext.Provider value={{headerHeight, setHeaderHeight}}>
-          <BottomSheetModalProvider>
-            <PaperProvider theme={paperTheme}>
-              <NativeBaseProvider theme={theme}>
-                <Portal.Host>
-                  <NavigationContainer>
+        <NavigationContainer>
```
tm-ruxandra, line 94: The `BottomSheetScrollView` wasn't behaving correctly in the sites screen filters unless I moved the `NavigationContainer` to be above the outermost `BottomSheetModalProvider`.

dev-client/src/hooks/gorhomHooks.ts
```diff
@@ -0,0 +1,54 @@
...
+  useMemo,
+  useRef,
+} from 'react';
+
+import {BottomSheetModal as GorhomBottomSheetModal} from '@gorhom/bottom-sheet';
+
+import {ModalHandle} from 'terraso-mobile-client/components/modals/Modal';
+
+/* Allows binding a GorhomBottomSheetModal to forwarded refs of our own ModalHandle type */
+export const useGorhomModalHandleRef = (
```
tm-ruxandra, line 31: This seemed like a good way to factor out the duplicated code between our three bottom sheets while preserving the simplicity of their single-purpose implementations. Am open to feedback for other approaches, though.

dev-client/src/screens/SoilScreen/components/EditDepthModal.tsx
```diff
@@ -176,7 +176,7 @@ export const EditDepthModal = ({
           onPress={onOpen}
         />
       )}
-      Header={
+      heading={
         <Heading variant="h6">
           {mutable ? t('soil.depth.edit_title') : renderDepth(t, thisDepth)}
```
tm-ruxandra, line 181: I was tempted to switch this to a `TranslatedHeader`, but since the immutable case directly renders a string (which that component isn't yet designed to handle) I thought it'd be better to defer this until we tackle our typography design tickets.

dev-client/src/components/sheets/FormOverlaySheetHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+}: FormOverlaySheetHeaderProps) => {
+  const {t} = useTranslation();
+
+  return (
+    <Pressable
+      onPress={onDone}
+      accessibilityRole="button"
+      accessibilityLabel={t('general.done')}
+      style={styles.header}>
+      <Text
```
tm-ruxandra, line 39: Wasn't totally sure what to do with styles here; ultimately left it as-is even though the container is defined with stylesheets now since typography components are a separate ticket.

dev-client/src/components/inputs/Select.tsx
```diff
@@ -107,49 +101,30 @@ export const Select = <T, Nullable extends boolean>({
...
-            );
-          })}
-        </MenuList>
-      </BottomSheetScrollView>
+      <MenuList>
+        {optionsWithNull.map(option => {
+          const itemLabel = option ? renderValue(option) : unselectedLabel!;
+          const selected = option === value;
+          let key;
+          if (optionKey) {
```
paulschreiber, line 109: This is much better than a double ternary.

dev-client/src/components/sheets/StandaloneOverlaySheet.tsx
tm-ruxandra: I don't think so. There's nothing about it that's specific to that component and there may be other similar cases in the future.

dev-client/src/screens/SoilScreen/SoilScreen.tsx
```diff
@@ -94,18 +95,16 @@ export const SoilScreen = ({siteId}: {siteId: string}) => {
         alignItems="center">
         <Heading variant="h6">{t('soil.pit')}</Heading>
         {!projectSettings && (
-          <FormOverlaySheet
-            Header={
-              <Heading variant="h6">{t('soil.soil_preset.header')}</Heading>
-            }
+          <InfoSheet
```
knipec, line 98: Is it by-design that this opens to 100% of the available height, even if the contents would allow 50%? 

Same question applies to EditDepthModal

knipec (approved):

dev-client/src/components/sheets/FormOverlaySheet.tsx
```diff
@@ -25,100 +25,53 @@ import {
...
+    const {modalRef, methods} = useGorhomModalHandleRef(ref);
 
     return (
       <>
-        {trigger && (
-          <Pressable onPress={methods.onOpen}>
-            {trigger(methods.onOpen)}
-          </Pressable>
-        )}
+        {trigger && trigger(methods.onOpen)}
```
knipec, line 52: The indirection with trigger and refs feels challenging to read -- it's not a result of your change, but wondering if there is some commenting here that could make it faster for a reader to understand

dev-client/src/screens/SoilScreen/SoilScreen.tsx
tm-ruxandra: Yes, this was confirmed with @CourtneyLee333

dev-client/src/components/sheets/FormOverlaySheet.tsx
tm-ruxandra: I'll see what I can do, I agree that it's confusing.

Issue #1931, opened 2024-08-07
Research - Redux state persistence library (NOT for data syncing)


DerekCaelin:
- Regardless of syncing, we just want to preserve redux state between launches, so that offline local changes to data are saved if you never have connectivity
- Serializing what's in memory, load it back from disk when app opens again

## References
[Offline Research](https://docs.google.com/presentation/d/1eHO52LqsSsOV6tyBQrEUA857VHsjGAdxQEDYWcWQLaY/edit#slide=id.g2f0a32b466d_0_13)

## Acceptance Criteria
- [ ] How would we break this down into smaller prototype chunks? (And create separate issues if appropriate)
- [ ] Make recommendation for Redux state persistence library

shrouxm:
i've written up my research in [this doc](https://docs.google.com/document/d/1cMRc5SD3XIQ1jrtR-AfAqdEXsag5hbJ32MMzM5FTp40/edit?usp=sharing)

shrouxm:
doc has been reviewed and approved by everyone, and its action items transcribed to github issues, so this is done!

PR #2105, opened 2024-09-05
feat: redux state persistence

### LINKED ISSUE #2164: feat: implement redux persistence prototype behind offline flag
We feel good about the approach defined in https://github.com/techmatters/terraso-mobile-client/pull/2105. To make it merge-able, we'd like to hide the implementation behind the offline feature flag, and add at least one integration test of the functionality.

shrouxm:
## Description
Minimal implementation of redux state persistence! Currently the behavior is behind the offline feature flag.

This small amount of code is surprisingly effective. With just these changes, the app now goes to directly to the home screen and (if you comment out the hardwired `fetchSoilDataForUser` call) immediately displays all the user's data on launch:

[Launching straight to home screen](https://github.com/user-attachments/assets/b18167d6-283c-4de1-87fd-7ca0ebaed9e8)

### Checklist

### Related Issues
Related to #1931 
Fixes #2164

### DIFF
--- a/dev-client/src/store/index.ts
+++ b/dev-client/src/store/index.ts
@@ -21,7 +21,11 @@ import {
   TypedUseSelectorHook,
 } from 'react-redux';
 
-import {configureStore, StateFromReducersMapObject} from '@reduxjs/toolkit';
+import {
+  configureStore,
+  Middleware,
+  StateFromReducersMapObject,
+} from '@reduxjs/toolkit';
 
 import {
   DispatchFromStoreFactory,
@@ -35,6 +39,15 @@ import {reducer as preferencesReducer} from 'terraso-mobile-client/model/prefere
 import projectReducer from 'terraso-mobile-client/model/project/projectSlice';
 import siteReducer from 'terraso-mobile-client/model/site/siteSlice';
 import soilIdReducer from 'terraso-mobile-client/model/soilId/soilIdSlice';
+import {kvStorage} from 'terraso-mobile-client/persistence/kvStorage';
+
+export const PERSISTED_STATE_KEY = 'persisted-state';
+const serializationMiddleware: Middleware = store => next => action => {
+  const result = next(action);
+  const newState = store.getState();
+  kvStorage.setMap(PERSISTED_STATE_KEY, newState);
+  return result;
+};
 
 const reducers = {
   ...sharedReducers,
@@ -55,7 +68,9 @@ export const useDispatch: () => AppDispatch = reduxUseDispatch;
 export const createStore = (intialState?: Partial<AppState>) =>
   configureStore({
     middleware: getDefaultMiddleware =>
-      getDefaultMiddleware().concat(handleAbortMiddleware),
+      getDefaultMiddleware()
+        .concat(handleAbortMiddleware)
+        .concat(serializationMiddleware),
     reducer: reducers,
     preloadedState: intialState,
   });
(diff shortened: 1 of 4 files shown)


dev-client/src/store/index.ts
```diff
@@ -32,12 +33,21 @@ import {
...
+import {MMKV} from 'terraso-mobile-client/hooks/useStorage';
 import {reducer as elevationReducer} from 'terraso-mobile-client/model/elevation/elevationSlice';
 import {reducer as mapReducer} from 'terraso-mobile-client/model/map/mapSlice';
 import {reducer as preferencesReducer} from 'terraso-mobile-client/model/preferences/preferencesSlice';
+import projectReducer from 'terraso-mobile-client/model/project/projectSlice';
+import siteReducer from 'terraso-mobile-client/model/site/siteSlice';
+import soilIdReducer from 'terraso-mobile-client/model/soilId/soilIdSlice';
+
+export const PERSISTED_STATE_KEY = 'persisted-state';
+const serializationMiddleware: Middleware = store => next => action => {
```
tm-ruxandra, line 45: Very elegant, I like this a lot.

dev-client/src/store/persistence.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+test('persistence middleware saves state to disk', () => {
+  jest.isolateModules(() => {
+    const {kvStorage} = require('terraso-mobile-client/persistence/kvStorage');
+    kvStorage.setBool('FF_offline', true);
+
+    const {
+      persistenceMiddleware,
+      loadPersistedReduxState,
+    } = require('terraso-mobile-client/store/persistence');
```
knipec, line 39: Question: Could you explain the use of `require` and how it differs from import here?

dev-client/src/store/persistence.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    expect(store.getState()).toEqual({counter: 1});
+
+    store.dispatch(increment());
+
+    expect(loadPersistedReduxState()).toEqual({counter: 2});
+  });
+});
+
+test('persistence middleware does nothing without feature flag', () => {
```
knipec, line 79: Suggestion/nit: Might help to be explicit about the feature flag off, so when we turn it on there are fewer tests to fix up

knipec (approved):

tm-ruxandra (approved):

dev-client/src/store/persistence.test.ts
shrouxm: answers:
- you can only use `import` at the top level of a file, not within a callback, otherwise i would use `import`
- yeah exactly! MMKV has functionality to reset the store after each test, but
- the `terraso-mobile-client/store/persistence` module imports the `terraso-mobile-client/config/featureFlags` module, so we need to re-initialize it on each test to be able to control the state of the feature flags

i can see this getting out of hand if we're testing lots of functionality that's gated behind feature flags, but this works for now and we can always do something smarter if it turns into a big DX issue

dev-client/src/store/persistence.test.ts
shrouxm: fair point! i was assuming that we'd just remove the feature flag when we're ready to have it on by default, but can't hurt to be careful + explicit here

dev-client/src/store/persistence.test.ts
knipec: Would it help to add to featureFlags.ts something like `enableFlag_forTestOnly`, that would add a given flag to `ENABLED_FLAGS` and allow us to bypass the feature flag-related kvStorage stuff?

Issue #2031, opened 2024-08-22
Implement a network status hook


knipec:
Now that we've done the initial investigation and prototyping (from #1930), do some stuff for real

## Acceptance criteria
- Create a hook that can be used throughout the app, that will provide network status (online, offline, or unknown/null). See [Whitespectre writeup](https://docs.google.com/document/d/1vc2GrmcZLsLPuqeQ5kWJQCxv_8ErfJr6FUR2b3Wr16w/edit) for guidance 
- Try out the proxy debugger that Whitespectre folks suggested --> This is now #2127
- App monitors network status when 'active' (meaning it's in the foreground)
- App stops monitoring network status when not 'active'
- Use it in the Soil ID sheet "Top Soil Matches" section to show a "You are offline" box (see image below)
- Consider: any tests we can write for this work?
- Should any of this be behind a feature flag? I think it'd be ok not to.

![Image](https://github.com/user-attachments/assets/7763a64b-a1cd-4950-be36-09f15c792588)

knipec:
Should be available in next build (v191beta)

DerekCaelin:
This worked for me! 
1. While online, open app
2. Open a site, go to the Soil ID page
3. Go into airplane mode
4. See change 

This appears to be both a working status hook and resolve:
- https://github.com/techmatters/terraso-mobile-client/issues/1360

Over to @CourtneyLee333

CourtneyLee333:
It works on Android and iOS! 

One design nit-pick: the alert box doesn't have the pale orange background as specified in the design. Should we go ahead a create a design component for this since we are introducing a new component to the app?

knipec:
@CourtneyLee333 Could you point me to the design with the pale orange background? 
All I see on the [Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=3800-22059&node-type=instance&t=jayxGV4uVBjAfph5-0) linked on #1360 is this palette: 
![Screenshot 2024-09-25 at 10 04 34 AM](https://github.com/user-attachments/assets/5bf5ab89-922c-41a0-90e7-eb57abbe4392)

CourtneyLee333:
@knipec oh, you're right! I forgot that I had updated the design from a previous version that had a pale orange background. All good, carry on.

PR #2135, opened 2024-09-11
feat: Show an offline message using a hook to report if app is offline or not

knipec:
## Description
- `isOffline` exists as state in a context provider, so there is a centralized source of truth.
- Setting up the netinfo event listener happens on app start. Listening for netinfo events will stop when the app is not "active", and will resume when the app becomes "active" again. This logic is also centralized in the context provider. The logic itself is heavily based on [Whitespectre's example code here](https://docs.google.com/document/d/1vc2GrmcZLsLPuqeQ5kWJQCxv_8ErfJr6FUR2b3Wr16w/edit)
- There is a trivial hook to use the isOffline state. We use it in the Soil ID sheet to show a "You are offline" message instead of the list of top soil matches.
- Add a small unit test

### Related Issues
Implements #2031

### DIFF
--- a/dev-client/src/context/ConnectivityContext.tsx
+++ b/dev-client/src/context/ConnectivityContext.tsx
@@ -1,0 +1,96 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {createContext, useEffect, useRef, useState} from 'react';
+import {AppState} from 'react-native';
+
+import {
+  addEventListener,
+  NetInfoSubscription,
+} from '@react-native-community/netinfo';
+
+export type ConnectivityContextType = {
+  isOffline: boolean | null;
+  setIsOffline: (newValue: boolean | null) => void;
+};
+
+export const ConnectivityContext =
+  createContext<ConnectivityContextType | null>(null);
+
+export const ConnectivityContextProvider = ({
+  children,
+}: React.PropsWithChildren) => {
+  const [isOffline, setIsOffline] = useState<boolean | null>(null);
+  const netInfoSubscriptionRef = useRef<NetInfoSubscription | null>(null);
+  console.log('re-render ConnectivityContextProvider');
+
+  useEffect(() => {
+    console.log('setup reachability monitoring');
+
+    const stopMonitoringReachability = () => {
+      if (netInfoSubscriptionRef.current) {
+        netInfoSubscriptionRef.current?.();
+        netInfoSubscriptionRef.current = null;
+      }
+    };
+
+    const startMonitoringReachability = () => {
+      stopMonitoringReachability();
+
+      netInfoSubscriptionRef.current = addEventListener(state => {
+        // TODO: Remove this
+        console.log(
+          'Connected =',
+          state.isConnected,
+          ' | Reachable =',
+          state.isInternetReachable,
+        );
+
+        // TODO: make a test for this cuz
+        //   false && null = false
+        //   null && false = null
+        const isOnline = state.isConnected && state.isInternetReachable;
+        setIsOffline(isOnline === null ? null : !isOnline);
+      });
+    };
+
+    if (AppState.currentState === 'active') {
+      startMonitoringReachability();
+    }
+
+    console.log('add app state listener');
+    const appStateListener = AppState.addEventListener('change', state => {
+      if (state === 'active') {
+        startMonitoringReachability();
+      } else {
+        stopMonitoringReachability();
+      }
+    });
+
+    return () => {
+      console.log('remove app state listener');
+      stopMonitoringReachability();
+      appStateListener.remove();
+    };
+  }, []);
+
+  return (
+    <ConnectivityContext.Provider value={{isOffline, setIsOffline}}>
+      {children}
+    </ConnectivityContext.Provider>
+  );
+};
--- a/dev-client/src/context/ConnectivityContext.tsx
+++ b/dev-client/src/context/ConnectivityContext.tsx
@@ -25,7 +25,6 @@ import {
 
 export type ConnectivityContextType = {
   isOffline: boolean | null;
-  setIsOffline: (newValue: boolean | null) => void;
 };
 
 export const ConnectivityContext =
@@ -89,7 +88,7 @@ export const ConnectivityContextProvider = ({
   }, []);
 
   return (
-    <ConnectivityContext.Provider value={{isOffline, setIsOffline}}>
+    <ConnectivityContext.Provider value={{isOffline}}>
       {children}
     </ConnectivityContext.Provider>
   );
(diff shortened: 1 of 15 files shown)


dev-client/src/hooks/connectivityHooks.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import {useContext} from 'react';
+
+import {ConnectivityContext} from 'terraso-mobile-client/context/ConnectivityContext';
+
+export const useIsOffline = () => {
+  const context = useContext(ConnectivityContext);
+
+  if (context === null) {
```
knipec, line 25: This feels clunky. Is there a way to ensure the context is hooked up to the state?

dev-client/src/context/ConnectivityContext.tsx
```diff
@@ -0,0 +1,96 @@
...
+        const isOnline = state.isConnected && state.isInternetReachable;
+        setIsOffline(isOnline === null ? null : !isOnline);
+      });
+    };
+
+    if (AppState.currentState === 'active') {
+      startMonitoringReachability();
+    }
+
+    console.log('add app state listener');
```
paulschreiber, line 75: issue: remove the debug logging.

dev-client/src/context/ConnectivityContext.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const ConnectivityContextProvider = ({
+  children,
+}: React.PropsWithChildren) => {
+  const [isOffline, setIsOffline] = useState<boolean | null>(null);
+  const netInfoSubscriptionRef = useRef<NetInfoSubscription | null>(null);
+  console.log('re-render ConnectivityContextProvider');
+
+  useEffect(() => {
+    console.log('setup reachability monitoring');
```
paulschreiber, line 42:
```suggestion
```

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -161,3 +167,17 @@ const NoMapDataAlertMessageContent = () => {
     </Box>
   );
 };
+
+const OfflineMessageBox = () => {
```
tm-ruxandra, line 171: I think it would be good to make this a dedicated component in another file.

dev-client/src/context/ConnectivityContext.tsx
```diff
@@ -0,0 +1,96 @@
...
+import {AppState} from 'react-native';
+
+import {
+  addEventListener,
+  NetInfoSubscription,
+} from '@react-native-community/netinfo';
+
+export type ConnectivityContextType = {
+  isOffline: boolean | null;
+  setIsOffline: (newValue: boolean | null) => void;
```
tm-ruxandra, line 28: IMO we shouldn't be exposing the state setter via the context like this; if possible it'd make more sense to be read-only, I think. (Unless there's a specific situation you've envisioned for setting it outside of this context provider?)

tm-ruxandra (changes requested):
The overall structure looks good and intuitive! A few comments before I'd want it merged in.

dev-client/src/context/ConnectivityContext.tsx
knipec: Ah, good call, thanks!

dev-client/src/context/connectivity/ConnectivityContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const appStateListener = AppState.addEventListener('change', state => {
+      if (state === 'active') {
+        startMonitoringReachability();
+      } else {
+        stopMonitoringReachability();
+      }
+    });
+
+    return () => {
+      stopMonitoringReachability();
```
knipec, line 75: This line not in the example code from Whitespectre folks. I'm guessing that's just because it's unnecessary if this provider wraps the app, but not harmful, is that correct?

dev-client/src/hooks/connectivityHooks.ts
```diff
@@ -0,0 +1,33 @@
...
+    console.warn(
+      'We expect to have set up the isOffline state before using this hook',
+    );
+    return null;
```
knipec, line 29: The null check feels a bit clunky. Is there a better way to enforce that we expect the `isOffline` state is created (by the `ConnectivityContextProvider`) before anything else uses the `ConnectivityContext`?

dev-client/src/screens/LocationScreens/components/soilId/MessageBoxesInsteadOfSoilMatches.tsx
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+import {useTranslation} from 'react-i18next';
+
+import {ExternalLink} from 'terraso-mobile-client/components/links/ExternalLink';
+import {AlertMessageBox} from 'terraso-mobile-client/components/messages/AlertMessageBox';
+import {ErrorMessageBox} from 'terraso-mobile-client/components/messages/ErrorMessageBox';
+import {Box, Text} from 'terraso-mobile-client/components/NativeBaseAdapters';
+
+export const NoMapDataAlertMessageBox = () => {
```
paulschreiber, line 25: suggestion: each of these components should be in its own file.

dev-client/src/context/connectivity/connectivityLogic.ts
```diff
@@ -0,0 +1,27 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+export const getIsOffline = ({
```
paulschreiber, line 18: question: do we want as isOnline function?

knipec: Sure, I'll extract it but not export it since nothing outside the file uses it so far

dev-client/src/hooks/connectivityHooks.ts
tm-ruxandra: Instead of `createContext<ConnectivityContextType | null>(null);` with a nullable context value, you could just set the type to  `SiteUserRole` and have the initial value be some default `ConnectivityContextType` object. (Basically the null-object pattern rather than actual `null`). Not sure what the implications would be of assuming that the context should be online / offline in its default uninitialized state, but presumably it'd be easy to settle on a sensible convention.

dev-client/src/context/connectivity/ConnectivityContext.tsx
tm-ruxandra: I think that's right? But if you're unsure we can wait for comment from them.

tm-ruxandra (approved):
LGTM (pending any other outside input you want!)

dev-client/src/context/connectivity/ConnectivityContext.tsx
lucaslt89: @knipec that's correct, it's not harmful to have this line and actually a safe net in case you want to unmount the context provider for any reason.

dev-client/src/context/connectivity/connectivityLogic.ts
```diff
@@ -0,0 +1,37 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+const getIsOnline = ({
```
lucaslt89, line 18: this `getIsOnLine` seems unnecessary? unless you plan to use it in your `ConnectivityContext` later on it could be part of the `getIsOffline` function:
```
export const getIsOffline = ({
  isConnected,
  isInternetReachable,
}: {
  isConnected: boolean | null;
  isInternetReachable: boolean | null;
}) => {
  return !isConnected || !isInternetReachable;
};
```

knipec: I mostly just had it for enforcement around when `getIsOffline` should return `null`, and that code has slightly different behavior when `isConnected` is true but `isInternetReachable` is null. I'm not sure how often we'll be in that state or how important it is to track, but since netinfo distinguishes between true / false / null, I figured we should too.

Issue #2100, opened 2024-09-03
Create feature flag infrastructure to gate offline functionality


knipec:
## Intention
We want to do development and testing on the new offline feature, while shipping releases that don't yet expose offline functionality to users.
Previously we'd been talking about having a long-lived feature branch like feat/offline or something. After some discussion, sounds like the preference is to have a feature flag. This will reduce the complexity of merging between branches, as well as potential confusion of which version we're testing against.

## Acceptance Criteria
- A feature flag exists. By default it is off.
- Users can see and toggle the feature flag state when in development or staging environments (not production)
- The team knows how to use it, and developers know how to write code with it
- It will be easy to enable when we are ready to ship the offline feature 
- It will be easy to remove entirely after it has been enabled in prod for a while

Optionally, but not part of this, a later stage could: 
- Make the feature flag system support multiple feature flags
- Figure out how to restart the app for you when you hit the toggle, or allow the developer to specify which flags require a restart

knipec:
Pulling this into the sprint to enable checking in work specific to offline, that is not ready to ship

knipec:
FYI 
- Developers can use the flag to gate new functionality: `if (isFlagEnabled('FF_offline')) { // do new thing }`
- Users in a non-production environment can now toggle the feature flag from the settings screen. The user must restart the app for the new feature flag value to take effect.
<img src=https://github.com/user-attachments/assets/ea7910dc-0623-4869-89f1-f772a5d0aec0 width=350>

- #2160 now tracks the semi-distant future work of removing the feature flag from the codebase once the feature is released and we're like, "it's on forever now and we're never gonna want to turn this off". @DerekCaelin Feel free to change the milestone of that item.

DerekCaelin:
Seems like the right milestone

DerekCaelin:
@knipec in what build can I test this?

knipec:
@DerekCaelin should be in 191

DerekCaelin:
@knipec I see the flag in settings in the current build! Is that all I need to test?

knipec:
@DerekCaelin Almost! Also that when you switch the flag between on/off, it's remembered. And when you clear app storage it's forgotten and set back to the default.

DerekCaelin:
Great! I've done that, and this passes.

PR #2139, opened 2024-09-13
feat: Add feature flag to gate offline functionality

knipec:
## Description
- Adds a feature flag for offline: `FF_offline`
- Allows a user in a non-production environment to toggle the feature flag. The user must restart the app for the new feature flag value to take effect. (This seemed the easiest thing to implement right now, but still allowing for user control.)

### Related Issues
Implements #2100 

### Screenshot
There should be no visible difference in a production environment, but looks like this in a non-prod environment:

### DIFF
--- a/dev-client/src/components/FeatureFlagControl.tsx
+++ b/dev-client/src/components/FeatureFlagControl.tsx
@@ -28,6 +28,7 @@ import {APP_CONFIG, MMKV} from 'terraso-mobile-client/config';
 export const FeatureFlagControl = () => {
   // It may cause issues to toggle a feature flag state while the app is running,
   // so expect the user to restart the app if they want to change the feature flag state.
+  // This should never be exposed in production.
 
   const currentFlagState = APP_CONFIG.FF_offline;
   const [nextFlagState, setNextFlagState] = useState<boolean>(
@@ -41,16 +42,20 @@ export const FeatureFlagControl = () => {
   };
 
   return (
-    <View>
-      <Heading mb="10px">Feature Flags</Heading>
-      <Text bold={true}>FF_offline</Text>
-      <Text>{`Currently: ${currentFlagState ? 'ON' : 'OFF'}`}</Text>
-      <View style={styles.nextFlagStateView}>
-        <Text>{`On next startup, flag will be: `}</Text>
-        <Switch value={nextFlagState} onValueChange={onToggle} />
-        <Text>{` ${nextFlagState ? 'ON' : 'OFF'}`}</Text>
-      </View>
-    </View>
+    <>
+      {APP_CONFIG.environment !== 'production' && (
+        <View>
+          <Heading mb="10px">Feature Flags</Heading>
+          <Text bold={true}>FF_offline</Text>
+          <Text>{`Currently: ${currentFlagState ? 'ON' : 'OFF'}`}</Text>
+          <View style={styles.nextFlagStateView}>
+            <Text>{`On next startup, flag will be: `}</Text>
+            <Switch value={nextFlagState} onValueChange={onToggle} />
+            <Text>{` ${nextFlagState ? 'ON' : 'OFF'}`}</Text>
+          </View>
+        </View>
+      )}
+    </>
   );
 };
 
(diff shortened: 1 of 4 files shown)


dev-client/src/components/FeatureFlagControl.tsx
```diff
@@ -0,0 +1,69 @@
...
+import {StyleSheet, View} from 'react-native';
+import {Divider, Switch} from 'react-native-paper';
+
+import {
+  Heading,
+  Text,
+} from 'terraso-mobile-client/components/NativeBaseAdapters';
+import {APP_CONFIG, MMKV} from 'terraso-mobile-client/config';
+
+export const FeatureFlagControlOrEmptyInProd = () => {
```
tm-ruxandra, line 27: I think it's fine to just call this component `FeatureFlagControl`, the use-case of hidden in prod is implicit imo.

tm-ruxandra (approved):

dev-client/src/components/FeatureFlagControl.tsx
```diff
@@ -0,0 +1,69 @@
...
+import {APP_CONFIG, MMKV} from 'terraso-mobile-client/config';
+
+export const FeatureFlagControl = () => {
+  // It may cause issues to toggle a feature flag state while the app is running,
+  // so expect the user to restart the app if they want to change the feature flag state.
+  // This should never be exposed in production.
+
+  const currentFlagState = APP_CONFIG.FF_offline;
+  const [nextFlagState, setNextFlagState] = useState<boolean>(
+    MMKV.getBool('FF_offline') ?? APP_CONFIG.FF_offline,
```
paulschreiber, line 34: suggestion: put this string in a const somewhere.

dev-client/src/components/FeatureFlagControl.tsx
```diff
@@ -0,0 +1,69 @@
...
+  Text,
+} from 'terraso-mobile-client/components/NativeBaseAdapters';
+import {APP_CONFIG, MMKV} from 'terraso-mobile-client/config';
+
+export const FeatureFlagControl = () => {
+  // It may cause issues to toggle a feature flag state while the app is running,
+  // so expect the user to restart the app if they want to change the feature flag state.
+  // This should never be exposed in production.
+
+  const currentFlagState = APP_CONFIG.FF_offline;
```
paulschreiber, line 32: question: is there a standard naming convention for feature flags? Should it be in all-caps? Should it have a more descriptive name, like FF_OFFLINE_SUPPORT_ENABLED?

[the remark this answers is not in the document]
knipec: I don't think there's a standard. I don't expect us to have many feature flags, so I chose to keep it short, since we will use it in every place where we want to branch behavior. But if we were planning to have several flags I might opt for more description. It should also be pretty easy to change in the future if we end up wanting to change its name.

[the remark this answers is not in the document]
knipec: Actually, I'll plan to add some description of what the flag is

knipec: Resolving because this comment is duplicated above :)

dev-client/src/components/FeatureFlagControl.tsx
knipec: Ok I got a little carried away but I think my most recent commit effectively addresses this 😅

dev-client/src/components/FeatureFlagControl.tsx
```diff
@@ -0,0 +1,87 @@
...
+    setNextFlagState(!nextFlagState);
+  };
+
+  return (
+    <View>
+      <Text bold={true}>{flag}</Text>
+      <Text>{featureFlags[flag].description}</Text>
+      <Text>{`Currently: ${currentFlagState ? 'ON' : 'OFF'}`}</Text>
+      <View style={styles.nextFlagStateView}>
+        <Text>{`On next startup will be: `}</Text>
```
paulschreiber, line 74:
```suggestion
        <Text>{`On next app launch will be: `}</Text>
```

PR #2166, opened 2024-09-18
feat: add Spanish localization

paulschreiber:
## Description
Add Spanish localization.

### DIFF
--- a/dev-client/src/translations/index.ts
+++ b/dev-client/src/translations/index.ts
@@ -17,9 +17,22 @@
 
 import {initReactI18next} from 'react-i18next';
 
+import {getLocales} from 'expo-localization';
+
 import i18n from 'i18next';
 
 import en from 'terraso-mobile-client/translations/en.json';
+import es from 'terraso-mobile-client/translations/es.json';
+
+const fallbackLanguage = 'en';
+let deviceLanguage = getLocales()[0].languageCode;
+
+// getLocales()[0].languageCode can be string|null, and
+// lng needs string|undefined. Check for null and set to the fallback
+// (could also set to undefined).
+if (deviceLanguage === null) {
+  deviceLanguage = fallbackLanguage;
+}
 
 i18n.use(initReactI18next).init({
   compatibilityJSON: 'v3',
@@ -29,10 +42,15 @@ i18n.use(initReactI18next).init({
         ...en,
       },
     },
+    es: {
+      translation: {
+        ...es,
+      },
+    },
   },
-  lng: 'en',
   debug: __DEV__ && process.env.NODE_ENV !== 'test',
-  fallbackLng: 'en',
+  lng: deviceLanguage,
+  fallbackLng: fallbackLanguage,
   interpolation: {
     // react already escapes HTML tags by default
     escapeValue: false,
--- a/dev-client/app.config.ts
+++ b/dev-client/app.config.ts
@@ -132,6 +132,7 @@ export default ({config}: ConfigContext): ExpoConfig => ({
   },
   plugins: [
     ['expo-apple-authentication'],
+    ['expo-localization'],
     ['expo-screen-orientation', {initialOrientation: 'PORTRAIT'}],
     [
       '@sentry/react-native/expo',
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -31,6 +31,7 @@
         "expo-dev-client": "^4.0.26",
         "expo-image-manipulator": "~12.0.5",
         "expo-image-picker": "~15.0.7",
+        "expo-localization": "^15.0.3",
         "expo-location": "~17.0.1",
         "expo-media-library": "~16.0.4",
         "expo-screen-orientation": "~7.0.5",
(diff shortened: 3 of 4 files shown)


dev-client/src/translations/en.json
knipec, line 1: Question: Just confirming there were no substantial changes in this file? Github's diff makes it very hard to tell :)

dev-client/scripts/localization/transform-to.mjs
```diff
@@ -77,35 +77,26 @@ const transform = (process, from, i18Transform) => {
 // PO transform
 const toPoOptions = {
   project: 'Terraso',
+  ctxSeparator: '',
```
knipec, line 80: Question: This is saying that we will never use the concept of context, right?

dev-client/src/translations/es.json
```diff
@@ -0,0 +1,875 @@
...
+            "title": "¿Qué sigue?",
+            "subtitle": "Versiones futuras incluirán:",
+            "bullet_3": "Propiedades adicionales del suelo, como la estructura y el carbono orgánico del suelo",
+            "bullet_4": "Fotos del sitio"
+        },
+        "version_includes": {
+            "title": "Esta version incluye:",
+            "bullet_3": "Notas acerca del sitio",
+            "bullet_4": "El manejo de projectos"
+        },
```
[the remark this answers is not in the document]
paulschreiber: That's a bug. Fixed.

dev-client/src/translations/en.json
paulschreiber: pass w=1 to GitHub and you can see the tiny diff:
https://github.com/techmatters/terraso-mobile-client/pull/2166/files?w=1

dev-client/scripts/localization/transform-to.mjs
paulschreiber: YEs. If we do, we will have to come up with a different separator (default is `_`) and update this configuration.

dev-client/src/translations/index.ts
```diff
@@ -17,9 +17,22 @@
...
+
+// getLocales()[0].languageCode can be string|null, and
+// lng needs string|undefined. Check for null and set to the fallback
+// (could also set to undefined).
+if (deviceLanguage === null) {
+  deviceLanguage = fallbackLanguage;
+}
 
 i18n.use(initReactI18next).init({
   compatibilityJSON: 'v3',
```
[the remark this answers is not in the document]
paulschreiber: Good catch. I've updated this.

dev-client/src/translations/es.json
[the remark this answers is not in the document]
knipec: The es.json file was generated from the es.po file though, right? It looks like the es.po file is in a different order than the en.po file -- like `msgid "welcome##version_includes##title"` is on line 14 in english and 833 in spanish. I'm guessing that's why it's ordered differently here, does that sound right?

I'm guessing it's not a huge issue but do you know how/why they would have got re-ordered? I forget if we supply the translator with an empty Spanish PO file, or if they create it themself?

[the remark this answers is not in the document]
paulschreiber: It works like this:
en.json
-> en.po (conversion script)
-> POEditor -> translator fills in translations
-> es.po (POEditor export)
-> es.json (conversion script)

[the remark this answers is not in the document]
knipec: (Chatted with Paul, who will look into this more once we get all the translations.)

dev-client/src/util.ts
```diff
@@ -24,20 +25,15 @@ import {
 } from 'terraso-client-shared/utils';
 
 import {COORDINATE_PRECISION} from 'terraso-mobile-client/constants';
+import {fallbackLocale} from 'terraso-mobile-client/translations';
 
 export const getSystemLocale = () => {
```
knipec, line 30: Suggestion: Could the locale and the language logic live next to each other? Since they look very similar and are about similar things

dev-client/src/translations/index.ts
```diff
@@ -17,22 +17,42 @@
...
+
 import i18n from 'i18next';
 
 import en from 'terraso-mobile-client/translations/en.json';
+import es from 'terraso-mobile-client/translations/es.json';
+
+export const fallbackLanguage = 'en';
+export const fallbackLocale = 'en-US';
+
+let deviceLanguage = getLocales()[0]?.languageCode;
```
knipec, line 30: Question: I forget if/where we landed on MVP behavior as a team -- is this the final behavior that Derek and Courtney agreed to, or were we considering adding additional logic for how a user would specify their language of choice (for MVP or later)?

.github/workflows/localization.yml
knipec, line 1: Question: This won't block checkin if it identifies missing localization keys, right? It's just an FYI, and we should just make sure it passes before a release to the public? 

(My thought is: Glancing at missing-check.mjs, it looks like it compares json files. But when we make new changes to code, we'll add to the en.json but probably not to the es.json file.)

paulschreiber: Yes, that's how it works.

dev-client/src/translations/index.ts
paulschreiber: We're starting without in-app UI for switching.

dev-client/src/util.ts
paulschreiber: I created `localization.ts` for related functions and constants.

knipec (approved):

PR #2167, opened 2024-09-18
fix: Changing location permissions should update components immediately

knipec:
## Description
**Issue**: Various issues with foreground permissions and zooming to the user's location on the map were caused by the expo-location library's `useForegroundPermissions` hook:
1. It only updates permissions on call to get(). 
2. When get() is called, the updated permissions are only local to component using the hook; it does not cause other components to re-render. This means components could be using outdated permissions. 

**Solution**: This PR makes and uses a hook that will update all components when foreground permissions are changed 
within the app (e.g., null to denied, or denied to granted), OR 
outside the app (in app settings there’s “Allow only while using the app”, “Ask every time”, and “Don’t allow”). 

**FYI** I do not recommend looking through the commits; it's chaos in there 😅 

### Related Issues
- Fixes an issue in with my earlier change addressing #1780 
- Is a starting point for #1749 
- (Similar approach to netinfo work in [PR #2135](https://github.com/techmatters/terraso-mobile-client/pull/2135))

### DIFF
--- a/dev-client/src/context/AppPermissionsContext.tsx
+++ b/dev-client/src/context/AppPermissionsContext.tsx
@@ -26,12 +26,16 @@ import {
 
 export type ForegroundPermissionsType = {
   permissions: LocationPermissionResponse | null;
-  get: () => Promise<LocationPermissionResponse>;
-  request: () => Promise<LocationPermissionResponse>;
-} | null;
+  get: (() => Promise<LocationPermissionResponse>) | null;
+  request: (() => Promise<LocationPermissionResponse>) | null;
+};
 
 export const ForegroundPermissionsContext =
-  createContext<ForegroundPermissionsType>(null);
+  createContext<ForegroundPermissionsType>({
+    permissions: null,
+    get: null,
+    request: null,
+  });
 
 export const ForegroundPermissionsProvider = ({
   children,
--- a/dev-client/src/context/AppPermissionsContext.tsx
+++ b/dev-client/src/context/AppPermissionsContext.tsx
@@ -38,47 +38,37 @@ export const ForegroundPermissionsProvider = ({
   const [permissions, request, get] = useForegroundPermissions();
   const [updatedPermissions, setUpdatedPermissions] =
     useState<LocationPermissionResponse | null>(permissions);
-  console.log('---- Permissions Provider re-rendering ----');
 
   const updatedRequest = useCallback(async () => {
     const response = await request();
-    console.log('    Calling updatedRequest(), got', response.status);
     setUpdatedPermissions(response);
     return response;
   }, [request, setUpdatedPermissions]);
 
   const updatedGet = useCallback(async () => {
     const response = await get();
-    console.log('    Called updatedGet(), got', response.status);
     setUpdatedPermissions(response);
     return response;
   }, [get, setUpdatedPermissions]);
 
   useEffect(() => {
-    console.log(
-      'Considering adding a new app state listener... --> updatedPermissions',
-      updatedPermissions?.status,
-    );
     // Don't start listening until someone asks about location permissions.
-    // If app switches from background to foreground, update permissions in case they changed
-    if (updatedPermissions) {
-      console.log('YES, add AppState listener');
+    //   updatedPermissions will only be non-null if permissions have been gotten/requested so far
+    //   For this to be fully correct, all components must use the updated hook, rather than useForegroundPermissions
 
+    if (updatedPermissions) {
+      // If app switches from background to foreground, update permissions in case they changed
       const onAppStateChange = async (state: AppStateStatus) => {
-        console.log('App state changed to', state);
         if (state === 'active') {
-          console.log('    Calling updatedGet()');
           updatedGet();
         }
       };
-
       const appStateListener = AppState.addEventListener(
         'change',
         onAppStateChange,
       );
 
       return () => {
-        console.log('Destroying AppState listener');
         appStateListener.remove();
       };
     }
--- a/dev-client/src/context/AppPermissionsContext.tsx
+++ b/dev-client/src/context/AppPermissionsContext.tsx
@@ -15,8 +15,8 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import {createContext, useCallback, useEffect, useRef, useState} from 'react';
-import {AppState, AppStateStatus, NativeEventSubscription} from 'react-native';
+import {createContext, useCallback, useEffect, useState} from 'react';
+import {AppState, AppStateStatus} from 'react-native';
 
 import {
   LocationPermissionResponse,
@@ -38,7 +38,6 @@ export const ForegroundPermissionsProvider = ({
   const [permissions, request, get] = useForegroundPermissions();
   const [updatedPermissions, setUpdatedPermissions] =
     useState<LocationPermissionResponse | null>(permissions);
-  const appStateListener = useRef<NativeEventSubscription | null>(null); // Do we still need a ref for this?
   console.log('---- Permissions Provider re-rendering ----');
 
   const updatedRequest = useCallback(async () => {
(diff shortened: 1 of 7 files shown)


dev-client/src/hooks/appPermissionsHooks.ts
```diff
@@ -0,0 +1,47 @@
...
+// when app permissions are updated. Instead, the permissions object stays stale until the next call to
+// get() or request().
+// This hook wraps the library-supplied hooks and should cause components to update when permissions are
+// updated in the app or after they are updated outside the app.
+export type UpdatedPermissionsHookReturnType = [
+  PermissionResponse | null, // permission
+  () => Promise<PermissionResponse>, // request
+  () => Promise<PermissionResponse>, // get
+];
+export type UpdatedPermissionsHookType = () => UpdatedPermissionsHookReturnType;
```
knipec, line 34: Soliciting advice: The types feel a little sloppy between here and in AppPermissionsContext.tsx. 
FYI
- `UpdatedPermissionsHookReturnType` aims to match types compatible with `PermissionsRequestWrapper` (which matches the expo permission hooks' types)
- `ForegroundPermissionsType` is similar, but allows additional nulls for get and request because I wanted something easy to set up the context with. 
- The context provider should populate with actual values when it is mounted on app launch, so the hook expects them not to be null.

I’m also torn about if `ForegroundPermissionsType` should deviate further from `UpdatedPermissionsHookReturnType`, since it would be nice to have an object with named keys, but the array matches the way everything else is structured.

dev-client/src/screens/BottomTabsScreen.tsx
```diff
@@ -38,19 +37,24 @@ import {useDispatch} from 'terraso-mobile-client/store';
...
 
   useEffect(() => {
-    if (!potentiallyOutdatedLocationPermission?.granted) {
-      requestLocationPermission();
+    const requestAndAwaitPermissions = async () => {
+      await requestLocationPermissions();
+    };
+
+    if (!locationPermissions?.granted) {
+      requestAndAwaitPermissions();
```
paulschreiber, line 49: question: can you call ` await requestLocationPermissions()` directly?

knipec: Ah thanks! 
Details: You can't call `await requestLocationPermissions()` directly in the `useEffect` because using await requires the calling function to be async and you can't make a useEffect async, so if you want to call something with `await` you need to make a separate async function and call it this way ([source](https://devtrium.com/posts/async-functions-useeffect)). 
However, I think we actually can just call `requestLocationPermissions()` here because a) the updated permissions hook should handle the permissions changing, and b) we weren't actually waiting for that to complete before starting the next useEffect in the first place.

dev-client/src/hooks/appPermissionsHooks.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// As of Sept 2024, the app permissions hooks supplied by the expo libraries don't trigger component updates
+// when app permissions are updated. Instead, the permissions object stays stale until the next call to
+// get() or request().
+// This hook wraps the library-supplied hooks and should cause components to update when permissions are
+// updated in the app or after they are updated outside the app.
+export type UpdatedPermissionsHookReturnType = [
+  PermissionResponse | null, // permission
+  () => Promise<PermissionResponse>, // request
+  () => Promise<PermissionResponse>, // get
```
shrouxm, line 32: question: do we ever need to call `get` outside of `AppPermissionsContext`?

dev-client/src/screens/BottomTabsScreen.tsx
```diff
@@ -38,19 +37,20 @@ import {useDispatch} from 'terraso-mobile-client/store';
 export const BottomTabsScreen = memo(() => {
   const dispatch = useDispatch();
   const {keyboardStatus} = useKeyboardStatus();
-  const [potentiallyOutdatedLocationPermission, _, requestLocationPermission] =
-    useForegroundPermissions();
+  const [locationPermissions, requestLocationPermissions, _] =
+    useUpdatedForegroundPermissions();
```
shrouxm, line 41: question: am i right in understanding that the expo hooks return `[permission, get, request]` and the new hook returns `[permission, request, get]`? should we keep the same order?

dev-client/src/context/AppPermissionsContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  (() => Promise<LocationPermissionResponse>) | null, // get
+];
+
+export const ForegroundPermissionsContext =
+  createContext<ForegroundPermissionsType>([null, null, null]);
+
+export const ForegroundPermissionsProvider = ({
+  children,
+}: React.PropsWithChildren) => {
+  const [permissions, request, get] = useForegroundPermissions();
```
shrouxm, line 38: thoughts: 

i'm trying to think through if there are any cracks for a permission change to slip through. it seems like in order for that to happen, the permissions would have to change:
- while the app is active but,
- not as a result of us calling this provider's "request" method

so i guess the question is, is there any reason this would happen besides our own programming error of calling expo's `useForegroundPermissions`? 

i can think of obscure hypotheticals (e.g. android and iOS both have modes where you can split the screen and open multiple apps simultaneously, so you could hypothetically change the settings while the app is inactive), but none that seem like blockers for this PR since it addresses the current open bugs and the user always has the "close and re-open" workaround

but maybe worth noting in a comment as a potential avenue of investigation if any more permissions bugs crop up?

dev-client/src/context/AppPermissionsContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const updatedGet = useCallback(async () => {
+    const response = await get();
+    setUpdatedPermissions(response);
+    return response;
+  }, [get, setUpdatedPermissions]);
+
+  useEffect(() => {
+    // Don't start listening until someone asks about location permissions.
+    //   updatedPermissions will only be non-null if permissions have been gotten/requested so far
+    //   For this to be fully correct, all components must use the updated hook, rather than useForegroundPermissions
```
shrouxm, line 57: suggestion: also put this Very Important comment somewhere more visible than inside an implementation method. maybe next to `useUpdatedForegroundPermissions` since that would be what consumers of this API will actually use?

shrouxm (approved):
an hero! thanks for doing this, no major issues i noticed! and seems pretty straightforward to generalize to other permission types :+1:

dev-client/src/hooks/appPermissionsHooks.ts
shrouxm: thoughts:
- rather than initializing with `[null, null, null]`, you could initialize with `null`, and  `ForegroundPermissionsType` could become `UpdatedPermissionsHookReturnType | null`, or `UpdatedPermissionsHookReturnType` could become `NonNullable<ForegroundPermissionsType>`
- i agree about it seeming nicer to have named keys, i understand why react does this with `useState` but don't like how it seems to get cargo culted onto everyone's hook definitions when it doesn't make as much sense

dev-client/src/hooks/appPermissionsHooks.ts
knipec: Is this about what level of the component hierarchy the ForegroundPermissionsProvider should live at, in App.tsx? If so, I'm not 100% sure, but I can't think of anything in particular. And I don't believe anything currently relies on it; the only thing I could think of was GeospatialContext.tsx, which uses state.map.userLocation, which I believe starts null, and only gets updated by BottomTabsScreen if location permissions have been granted.

Lmk if I'm not actually answering your question though!

dev-client/src/screens/BottomTabsScreen.tsx
knipec: Nope! The expo hooks return [permission, request, get] 
https://docs.expo.dev/versions/latest/sdk/location/#useforegroundpermissionsoptions 

Ha, this got me too while I was working on this -- I'd been under the same impression and spent a while being very confused because I'd been calling request() when I thought I was calling get() and vice versa. 😅 

This is primarily why I'm like "hmm should I just bite the types bullet and return an object with named types instead of a list?"

knipec: Ok I have done it

dev-client/src/hooks/appPermissionsHooks.ts
shrouxm: it was more like, "if no one is using this can we keep the API surface smaller". but i'm not sure if that might break usage with `PermissionsRequestWrapper`? not a big deal!

dev-client/src/screens/BottomTabsScreen.tsx
shrouxm: i see, so this is actually fixing a bug? since the `request` swaps places in this diff? & i don't feel a way about it if you don't want to bother with the object with named types so up to you!

knipec: Ah, yes, it was fixing a bug! Sorry I think my eyes had glazed over the actual code in question here, oops

And I have just finished bothering with the named things!

shrouxm (approved):
nice, looks cleaner, ty!

PR #2177, opened 2024-09-19
refactor: consolidate MMKV usage

### LINKED ISSUE #2161: refactor: use canonical MMKV instance
Currently we initialize the MMKV in two places in the app:
- https://github.com/techmatters/terraso-mobile-client/blob/f197e61cd11310a06dafe75ef6ae5862a076ee23/dev-client/src/hooks/useStorage.ts#L20
- https://github.com/techmatters/terraso-mobile-client/blob/22d8653f1dc075b524f188789606615074f142ba/dev-client/src/config/index.ts#L27

The MMKV docs recommend initializing only once, and we use separate settings in both instances. We should define a single place where MMKV is loaded.

We also need to understand how the `withEncryption` functionality works, so as part of completing this work we should document that alongside the initialization.

shrouxm:
## Description
Consolidates our usage of MMKV.

### Checklist

### Related Issues
Fixes #2161

### Verification steps

### DIFF
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -98,7 +98,6 @@
         "react-devtools": "^5.3.1",
         "react-native-svg-transformer": "^1.5.0",
         "react-test-renderer": "^18.3.1",
-        "ts-jest": "^29.2.5",
         "ts-morph": "^23.0.0",
         "ts-node": "^10.9.2",
         "typescript": "^5.6.2"
@@ -10046,12 +10045,6 @@
         "node": ">=4"
       }
     },
-    "node_modules/async": {
-      "version": "3.2.6",
-      "resolved": "https://registry.npmjs.org/async/-/async-3.2.6.tgz",
-      "integrity": "sha512-htCUDlxyyCLMgaM3xXg0C0LW2xqfuQ6p05pCEIsXuyQ+a1koYKTuBMzRNwmybfLgvJDMd0r1LTn4+E0Ti6C2AA==",
-      "dev": true
-    },
     "node_modules/async-limiter": {
       "version": "1.0.1",
       "license": "MIT"
@@ -10586,17 +10579,6 @@
       },
       "engines": {
         "node": "^6 || ^7 || ^8 || ^9 || ^10 || ^11 || ^12 || >=13.7"
-      }
-    },
-    "node_modules/bs-logger": {
-      "version": "0.2.6",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "fast-json-stable-stringify": "2.x"
-      },
-      "engines": {
-        "node": ">= 6"
       }
     },
     "node_modules/bser": {
@@ -12383,20 +12365,6 @@
       "version": "1.1.1",
       "license": "MIT"
     },
-    "node_modules/ejs": {
-      "version": "3.1.10",
-      "dev": true,
-      "license": "Apache-2.0",
-      "dependencies": {
-        "jake": "^10.8.5"
-      },
-      "bin": {
-        "ejs": "bin/cli.js"
-      },
-      "engines": {
-        "node": ">=0.10.0"
-      }
-    },
     "node_modules/electron": {
       "version": "23.3.13",
       "dev": true,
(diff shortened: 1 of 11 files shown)


tm-ruxandra (approved):

dev-client/src/config/index.ts
```diff
@@ -81,5 +81,4 @@ export const APP_CONFIG = {
   googleClientId,
   googleRedirectURI,
   microsoftRedirectURI,
-  FF_offline: willFlagBeEnabledOnReload('FF_offline'),
```
shrouxm, line 86: note: i moved this to `featureFlags.ts` because:
- we already expose a function which reads off these `APP_CONFIG` values and it seems better to not have more than one way to do that
- it makes it easier to test `featureFlags.ts` as an isolated module

shrouxm:
waiting for feedback on tests before merging

dev-client/src/config/index.ts
knipec: Yay, I like it!

dev-client/src/persistence/kvStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+// generating an encryption key and storing it on the keychain,
+// which we would have to do manually if we were using the other
+// library. at time of writing there is not a well-documented
+// migration story from one to the other.
+//
+// MMKV does not necessarily write to disk after every operation,
+// it will keep data in memory and only flush it when necessary
+// for performance reasons. I'm not sure if the RN API exposes a
+// way to manually flush to disk, there's a `clearMemoryCache`
+// method which presumably does this but haven't confirmed.
```
knipec, line 41: Question: I assume it will flush on app close -- is preventing data loss if the app crashes the main impetus for considering flushing manually? Are there other use cases you have in mind?

knipec (approved):

dev-client/package-lock.json
```diff
@@ -98,7 +98,6 @@
         "react-devtools": "^5.3.1",
         "react-native-svg-transformer": "^1.5.0",
         "react-test-renderer": "^18.3.1",
-        "ts-jest": "^29.2.5",
```
paulschreiber, line 101: issue: you removed this file by accident.

paulschreiber (changes requested):

dev-client/jest/unit/setup.ts
```diff
@@ -15,4 +15,11 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-// fill me in if you need special config for unit testing
+let mmkvMock = require('react-native-mmkv-storage/jest/dist/jest/memoryStore.js');
+mmkvMock.mock(); // Mock the storage
+
+beforeEach(() => {
```
tm-ruxandra, line 21: I don't know if I agree with making this a globally-configured setting in our unit tests; is it possible to just set this up for the kv storage and feature flag test files? The logic seems so minimal and we don't know what the cost is of clearing the mmkv mock for every test case.

dev-client/package-lock.json
shrouxm: not an accident, it's no longer used

dev-client/src/persistence/kvStorage.ts
shrouxm: that's what i was thinking about!

dev-client/package-lock.json
paulschreiber: sorry for the mixup. GitHub's UI made it looked like to remove the whole file.

paulschreiber (approved):

dev-client/jest/unit/setup.ts
shrouxm: i looked into it a bit, it looks like the cost of clearing the mock is more or less setting a handful of pointers to null and leaving some objects to be GC'd

it is possible to do it in individual test files, but it can be awkward because you need to manually make sure the imports resolve in the right order since the mock expects to be required in a setup file like this before any of the library's use sites are resolved

if it turns out to be a problem it should be straightforward to manually inline it though

Issue #2230, opened 2024-10-02
Prototype - load new data from server (October)


knipec:
Acceptance Criteria
- [ ] Investigate / Prototype:
  - [x] A) General system for screens requiring data
  - [ ] B) Prototype: Identify when response has an error, and do a pull 
  - [ ] How to test
- [ ] Identify next steps
  - Test
  - A) When user is viewing a screen associated with an entity someone else deleted, then syncs
    - Update design to support the entity still existing, but you can no longer edit it -- like if you were editing a site note, or a project, and you were turned from a manager to a viewer 
    - Eng: Audit each screen to take id's as input, specify required data, and specify what to do when it doesn't exist
    - UX: Do we show something to the user?
  - B) When user tries to make an edit to an entity someone else deleted, without a sync
    - [[TO DECIDE]]
  - Do the pulls at the right times
  - What does the pull return from server? --> Recommendation: Start with everything, then optimize later
  - Minor issues
    - Color dropdowns don’t update
    - UX: Confirm our behavior with remembering but hiding data at old depth intervals makes sense
  - UX: Do we show an indicator that the pull is happening? Or when it last happened? (Post-MVP?) 
- [ ] Agreement and understanding from team
- [ ] Create new items in GitHub

PR #2231, opened 2024-10-02
feat: Add sync button & handle deleted required entities on a few screens after pulling data from server

knipec:
## Description
Spinoff work from #2230 
- Adds a sync button to most screens (behind feature flag) that queries for everything
- Update a few screens:
  - Use the new component wrapper `RestrictByRequirements`. It is intended to wrap screens and if any of its required data is missing, it will prevent the screen from rendering, and will do an action.
  - Make sure screens take id's and use redux to get data. If data is passed in as props, it may become stale if data is pulled from server while on that screen.
- Refactoring to break `LocationDashboardScreen` into `SiteTabsScreen` and `TemporaryLocationScreen`. Both will still use `LocationDashboardContent`. Some additional renaming in this area. Apologies, this is why there are many files with changes.

### Still to do for the "sync from server" work:
- How to test this?
- Do the right thing on screens when the data exists, but you shouldn't be allowed to see the current screen because permissions changed (this could happen if you were a manager editing a site note, then became a viewer).
- Audit all screens to specify requirements
- Identify when push response has an error, and then do a pull

### Related Issues
Implements #2232


dev-client/src/screens/ScreenScaffold.tsx
```diff
@@ -63,6 +65,7 @@ export const ScreenScaffold = ({
       />
       <Column backgroundColor="primary.contrast" flex={1}>
         <View onLayout={onLayout}>{PropsAppBar}</View>
+        {isFlagEnabled('FF_offline') && <SyncButton />}
```
paulschreiber, line 68: question: Would it make sense to a Flag Restricted component, like we do for permissions? i.e.
```
<RestrictByFlag flag="FF_offline" component={<SyncButton />} />
```

or
```
<RestrictByFlag flag="FF_offline"><SyncButton /></RestrictByFlag>
```

dev-client/src/components/RenderIfDataExistsAndHandleIfNot.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return requirements.every(({data}) => dataExists(data));
+};
+
+type Props = {
+  requirements: RequiredDataAndWhatToDoIfMissing[];
+  // Use "Function as Child" pattern to defer evaluation of children's props, so they may expect required data
+  children: () => React.ReactNode;
+};
+
+export const RenderIfDataExistsAndHandleIfNot = ({
```
paulschreiber, line 62: issue: this is an awkward name. let's come up with something better.

tm-ruxandra: I agree, the name obscures far more than it explains and is difficult to verbally parse. `RestrictByDependencies` would match the convention of our `RestrictByRole`, the behavior of which it is very similar to.

dev-client/src/components/RenderIfDataExistsAndHandleIfNot.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {useEffect} from 'react';
+import {ToastAndroid} from 'react-native';
+
+import {isFlagEnabled} from 'terraso-mobile-client/config/featureFlags';
+import {useNavigation} from 'terraso-mobile-client/navigation/hooks/useNavigation';
+
+type RequiredDataAndWhatToDoIfMissing = {
```
tm-ruxandra, line 24: Similar to the overall type name I feel that this name should be simplified.

dev-client/src/screens/ScreenScaffold.tsx
tm-ruxandra: I think this would be a good idea.

dev-client/src/components/RenderIfDataExistsAndHandleIfNot.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  useEffect(() => {
+    for (let {data, doIfMissing} of requirements) {
+      if (!dataExists(data)) {
+        doIfMissing === undefined
+          ? navigation.navigate('BOTTOM_TABS')
+          : doIfMissing();
+        // TODO: Decide design / Decide how to show toasts?
+        if (isFlagEnabled('FF_offline')) {
+          ToastAndroid.show('Sorry, someone deleted that!', ToastAndroid.SHORT);
```
knipec, line 46: Move specific default behavior elsewhere, this component seems more general

dev-client/src/components/RenderIfDataExistsAndHandleIfNot.tsx
knipec: "Requirements"

dev-client/src/screens/ScreenScaffold.tsx
knipec: Thanks, will do! Any preferences on if it should go within the SyncButton, or if it should be the caller's responsibility to put it around the SyncButton, or both? (I don't expect there to be many callers so shouldn't matter too much)

dev-client/src/components/RenderIfDataExistsAndHandleIfNot.tsx
knipec: Made a hook `useHandleMissingSite` to return a the default `doIfMissing` function for missing sites

dev-client/src/screens/SiteNotesScreen/EditSiteNoteScreen.tsx
```diff
@@ -75,30 +92,35 @@ export const EditSiteNoteScreen = ({note}: Props) => {
...
-              content={formikProps.values.content}
-              editDisabled={!currentUserIsAuthor}
-            />
-          </Box>
-        </Column>
+    <RestrictByRequirements requirements={requirements}>
+      {() => (
+        <ScreenFormWrapper
+          ref={formWrapperRef}
+          initialValues={{content: note.content}}
```
knipec, line 106: I find this ["Function as a child component" pattern](https://reactpatterns.js.org/docs/function-as-child-component/) kind of ugly, but without it, we'd get errors like 
image
because React evaluates it before knowing if its associated component will end up being rendered.

Feel free to give opinions on the options:
- Keep it
- Could instead just change this to `note?.content` and find all the places to put question marks in subsequent screens -- but I feel like this obscures the intent (makes you think that ScreenFormWrapper could render without a note)
- Something else I haven't thought of?

paulschreiber (commented):
Looks go to me. Will Let others comment on the "function as child component" comment.

dev-client/src/screens/SiteNotesScreen/EditSiteNoteScreen.tsx
knipec, line 1: From garo: Would be nice to have an EditSiteNoteScreenContent to split up the logic for checking that everything exists, and then doing the logic of the screen

dev-client/src/components/dataRequirements/RestrictByRequirements.tsx
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+import {useEffect} from 'react';
+
+type Requirement = {
+  data: any;
+  doIfMissing?: () => void;
+};
+
+const dataExists = (data: any) => {
```
shrouxm, line 25: nit: i think you could use `object | undefined | null` to be a bit less permissive here

dev-client/src/components/dataRequirements/handleMissingData.ts
```diff
@@ -0,0 +1,34 @@
...
+import {ToastAndroid} from 'react-native';
+
+import {isFlagEnabled} from 'terraso-mobile-client/config/featureFlags';
+import {useNavigation} from 'terraso-mobile-client/navigation/hooks/useNavigation';
+
+export const useHandleMissingSite = () => {
+  const navigation = useNavigation();
+
+  return useCallback(() => {
+    navigation.navigate('BOTTOM_TABS');
```
shrouxm, line 28: question: why do we use `navigation.navigate('BOTTOM_TABS')` here but `navigation.pop()` elsewhere?

shrouxm (approved):
LGTM aside from the comments!

dev-client/src/screens/SiteNotesScreen/EditSiteNoteScreen.tsx
knipec: We may end up having a pattern where the requirements logic will go in the Screen component, and the rest of the screen's logic will be extracted into a Content component. 
That would at least isolate this to not be mixed in with the screen's content's logic. And maybe we will end up removing RestrictByRequirements if it seems unnecessary after that refactoring.

dev-client/src/components/dataRequirements/handleMissingData.ts
knipec: I was thinking because there are screens at various navigation depths that require sites, so instead of recursively pop()'ing, it'd be more straightforward to just always navigate to the same screen if we're missing a site.

I could change the navigation.pop() to navigate explicitly to the SiteNotesScreen though -- I don't think there's (currently) another screen that it'd pop() to in the case that a site note got deleted

dev-client/src/components/dataRequirements/RestrictByRequirements.tsx
knipec: This may change when we add permissions stuff, but happy to at least start out more permissive and adjust later if needed 👍

dev-client/src/components/dataRequirements/handleMissingData.ts
shrouxm: ohh like in a situation where you pop but the screen you popped to has also been deleted? do you know what would happen in this approach if we were just using `pop`? e.g. if we were in the `EditSiteNoteScreen` and the site got deleted, would it gracefully pop twice or would something weird happen?

i guess there's also a UX question here: if you navigate to a site from a project, and then that site gets deleted, should you be returned to the project or to the bottom tabs. but mainly i was just curious, cuz this is all pretty low stakes as long as it doesn't crash

knipec: I didn't try it cuz I was like "this approach seems fine" 😅 so I'm not sure if we would do multiple pop()s gracefully or if something weird would happen. Since this case is fairly uncommon I lean towards "just do the easiest reasonable thing".

Yeah I think the question of "where should each screen navigate when each piece of data is missing" is a thing we will tackle as part of [#2290](https://github.com/techmatters/terraso-mobile-client/issues/2290) but also here am leaning towards "just do the easiest reasonable thing"

PR #2288, opened 2024-10-09
feat: Unify Chip components

knipec:
## Description
Create and use a new `Chip` component that no longer uses the native-base `Badge` 

### Related Issues
Implements #1877 

### Before & After
(Ignore the weirdness with the Divider lines; I suspect it's just an artifact of my emulator window size)
Screenshot 2024-10-08 at 6 40 59 PM


dev-client/src/components/Chip.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <View style={styles.container}>
+      {iconName && (
+        <MaterialIcon
+          name={iconName}
+          size={convertIconSize('md')}
+          style={styles.icon}
+        />
+      )}
+      <Text variant="chip-text" style={styles.text}>
```
knipec, line 44: I assume it's not worth trying to switch over to the react-native-paper `Text` component yet, but lmk if you disagree

dev-client/src/screens/ProjectTeamScreen/components/RolePill.tsx
```diff
@@ -19,24 +19,13 @@ import {useTranslation} from 'react-i18next';
...
 import {ProjectMembership} from 'terraso-client-shared/project/projectTypes';
 
-import {Badge} from 'terraso-mobile-client/components/NativeBaseAdapters';
+import {Chip} from 'terraso-mobile-client/components/Chip';
 
 type RolePillProps = {
   membership: ProjectMembership;
 };
 
 export const RolePill = ({membership}: RolePillProps) => {
```
tm-ruxandra, line 28: This can probably be renamed to `RoleChip` and live with the others, it seems potentially more reusable than this screen.

dev-client/src/components/Chip.tsx
```diff
@@ -0,0 +1,69 @@
...
+type Props = {
+  label: string;
+  iconName?: IconName;
+};
+
+export const Chip = ({label, iconName}: Props) => {
+  return (
+    <View style={styles.container}>
+      {iconName && (
+        <MaterialIcon
```
tm-ruxandra, line 38: Was there a reason that you didn't use our existing `Icon` component here?

dev-client/src/components/SiteChip.tsx
```diff
@@ -15,12 +15,12 @@
...
-import {ChipBadge} from 'terraso-mobile-client/components/ChipBadge';
+import {Chip} from 'terraso-mobile-client/components/Chip';
 
 type Props = {
   count: number;
 };
 
-export const PeopleBadge = ({count}: Props) => {
-  return <ChipBadge count={count} iconName="people-alt" />;
+export const SiteChip = ({count}: Props) => {
```
tm-ruxandra, line 24: I think that, given 3-4 types of `Chip` (the base, the site one, the people one, and the role one) we should have a `chips` subdirectory under `components`. I would also circle back to Courtney to let her know that we have these dedicated `Chip` variants so she can document their existence in the component library.

tm-ruxandra (changes requested):
Overall looks good, a few minor things to address & a documentation issue that I'd resolve. Don't forget to update the snapshots.

dev-client/src/components/Chip.tsx
knipec: Oops, I'd missed that! Thanks!

dev-client/src/screens/ProjectTeamScreen/components/UserItem.tsx
```diff
@@ -88,7 +88,7 @@ export const UserItem = ({
     <MenuItem
       icon={<ProfilePic user={user} />}
       label={userLabel}
-      pill={<RolePill membership={membership} />}
+      pill={<RoleChip membership={membership} />}
```
tm-ruxandra, line 91: While we're here, can we rename this property to "chip"? Another thing to refer to Courtney wrt. updating the documented property in `MenuItem`.

tm-ruxandra (approved):
LGTM, one more name-consistency comment that might be nice to address before merging.

dev-client/src/components/chips/Chip.tsx
```diff
@@ -0,0 +1,64 @@
...
+import {Icon, IconName} from 'terraso-mobile-client/components/icons/Icon';
+import {Text} from 'terraso-mobile-client/components/NativeBaseAdapters';
+import {convertColorProp} from 'terraso-mobile-client/components/util/nativeBaseAdapters';
+
+type Props = {
+  label: string;
+  iconName?: IconName;
+};
+
+export const Chip = ({label, iconName}: Props) => {
```
paulschreiber, line 29: question: can/should we use the React Native Paper chip as a base object?

paulschreiber (commented):
question: can/should we use the React Native Paper chip as a base object?

knipec:
@paulschreiber 
> question: can/should we use the React Native Paper chip as a base object?

I had looked at it, but it seemed easier to create our own, since the styling of it (re: rounded corners) is different and it's not obvious to me how to change that. It's also just for display, and not interaction, so we don't need any of the press functionality. Though looking at its `ellipsizeMode` prop is making me wonder what should happen if the chip text is long -- I'll ask Courtney about that!

PR #2301, opened 2024-10-11
feat: global redux actions

### LINKED ISSUE #2162: refactor: resolve async redux actions in a way that maintains state consistency
### Context
Currently, many of our async redux actions dispatch multiple further actions upon resolution which need to _all_ resolve to keep the state consistent. This is already a potential source of bugs using redux as an in-memory store, but becomes a source of potential data corruption now that we're thinking of serializing our redux store.

### Task
Refactor existing async redux actions which dispatch further actions during their resolution to perform all state changes within the scope of a single action. I _think_ this is exactly equivalent to "all invocations of `createAsyncThunk` which transitively call `dispatch` or the `dispatchByKeys` helper function", and I _think_ all such invocations are in the mobile client.

We've identified two possible ways to implement this:

#### Extra reducers
the more straightforward way in our current codebase would be to have multiple slices respond to individual thunks via `extraReducers` as described in the docs [here](https://redux-toolkit.js.org/api/createslice/#extrareducers). For example, consider the `addSite` action: https://github.com/techmatters/terraso-mobile-client/blob/76b041f0e7623aa5b2c2fe8393edfb0dcfecb04c/dev-client/src/model/site/siteSlice.ts#L54-L63

Within the site slice, the action is resolved in `extraReducers` like so: https://github.com/techmatters/terraso-mobile-client/blob/76b041f0e7623aa5b2c2fe8393edfb0dcfecb04c/dev-client/src/model/site/siteSlice.ts#L166-L168

But the `addSite` action is also separately dispatching an `addSiteToProject` action in the other slice. We could instead have code in the project slice's `extraReducers` directly doing `builder.addCase(addSite.fulfilled, ...)`

This is more straightforward because it can be directly implemented in the app as is, but it has very poor locality of behavior: the effects of the `addSite` action are already not defined next to it, and now they're not even defined within the same file.

We only want to implement this way if we can't make another way work.

#### Global reducers
This would be implementing [this approach](https://redux.js.org/usage/structuring-reducers/beyond-combinereducers#sharing-data-between-slice-reducers) described in the redux docs, copied here for convenience:

> As it turns out, there's a useful utility called [reduce-reducers](https://github.com/acdlite/reduce-reducers) that can make that process easier. It simply takes multiple reducers and runs reduce() on them, passing the intermediate state values to the next reducer in line:
> ```
> // Same as the "manual" rootReducer above
> const rootReducer = reduceReducers(combinedReducers, crossSliceReducer)
> ```
> Note that if you use reduceReducers, you should make sure that the first reducer in the list is able to define the initial state, since the later reducers will generally assume that the entire state already exists and not try to provide defaults.

I'm not sure what exactly defining the `crossSliceReducer` would look like, so the newness/uncertainty here is what is preventing me from just saying "do this", but ideally we will do this!

### Issue size
I put a 7 to be safe because I'm not sure how it's going to go. This could easily be broken into subissues for enabling the global reducer functionality and doing the migration in parts if it's seeming like too big a chunk to do in one go.

shrouxm:
## Description
**Please review one commit at a time.** git really mangled the diff, so i rewrote the commit history and it's way cleaner/more comprehensible if you go one at a time and read the descriptions.

Ready for first round of review/discussion. Explanation (will crystallize into docs in the code once consensus is reached):

The only functional difference this PR introduces is that there are no longer actions which modify the state over the course of several dispatches. In order to make that functional change, I had to do a decent chunk of refactoring. I only want to merge this PR if there's consensus that the new shape of our redux code is an improvement on the old shape.

### Changes to codebase
- creates a helper function `createGlobalReducer` which allows you to create a reducer which can act on the entire redux state at once, rather than just a slice at a time
- rewrites many redux actions which existed for the sole purpose of being called by other redux actions (i.e. were not meant as an interface to UI code) as state mutations, examples: 
```
+export const setSoilData = (
+  state: Draft<SoilState>,
+  soilData: Record<string, SoilData>,
+) => {
+  state.soilData = soilData;
+  state.matches = {};
+};
+
 const soilIdSlice = createSlice({
   name: 'soilId',
   initialState,
   reducers: {
-    setSoilData: (state, action: PayloadAction<Record<string, SoilData>>) => {
-      state.soilData = action.payload;
-      state.matches = {};
-    },
```
```
+export const addSiteToProject = (
+  state: Draft<State>,
+  args: {siteId: string; projectId: string},
+) => {
+  state.projects[args.projectId].sites[args.siteId] = true;
+};
 const projectSlice = createSlice({
     ...
-    builder.addCase(
-      addSiteToProject,
-      (state, {payload: {siteId, projectId}}) => {
-        state.projects[projectId].sites[siteId] = true;
-      },
-    );
     ...
 });
-export const addSiteToProject = createAction<SiteKey>(
-  'project/addSiteToProject',
-);
```
- rewrites UI-facing redux actions which previously dispatched actions to different slices to instead be handled by global reducers which use the new state mutations. example:
```
const soilIdSlice = createSlice({
     ...
-    builder.addCase(fetchSoilDataForUser.pending, state => {
-      state.status = 'loading';
-    });
-
-    builder.addCase(fetchSoilDataForUser.rejected, state => {
-      state.status = 'error';
-    });
-
-    builder.addCase(fetchSoilDataForUser.fulfilled, state => {
-      state.status = 'ready';
-    });
     ...
});
...
-export const fetchSoilDataForUser = createAsyncThunk(
-  'soilId/fetchSoilDataForUser',
-  dispatchByKeys(soilDataService.fetchSoilDataForUser, () => ({
-    projects: setProjects,
-    sites: setSites,
-    projectSoilSettings: setProjectSettings,
-    soilData: setSoilData,
-    users: setUsers,
-  })),
-);

+export const fetchSoilDataForUser = createAsyncThunk(
+  'soilId/fetchSoilDataForUser',
+  soilDataService.fetchSoilDataForUser,
+);
+
+export const soilIdGlobalReducer = createGlobalReducer(builder => {
+  builder.addCase(fetchSoilDataForUser.fulfilled, (state, {payload}) => {
+    setProjects(state.project, payload.projects);
+    setSites(state.site, payload.sites);
+    setUsers(state.account, payload.users);
+    setProjectSettings(state.soilId, payload.projectSoilSettings);
+    setSoilData(state.soilId, payload.soilData);
+    updateSoilIdStatus(state.soilId, 'ready');
+  });
+
+  builder.addCase(fetchSoilDataForUser.pending, state => {
+    updateSoilIdStatus(state.soilId, 'loading');
+  });
+
+  builder.addCase(fetchSoilDataForUser.rejected, state => {
+    updateSoilIdStatus(state.soilId, 'error');
+  });
+});
```

### features of the new shape of the code
- IMO this looks closer to how the code would look if we had a DB, which is a pro in my mind. i wanted to stop working before the PR got too big, but it feels like one possible extension of this refactor is to have a separation between a slice's internal code, responsible for monkeying around with the state object, and a set of slice-neutral external-facing actions which call the slices' mutations. as opposed to the previous situation where we do the monkeying and API defining and such all in one big glob
- there's always been friction in the codebase because of the way dispatched actions are not composable/reusable (you can't dispatch within code resolving an action). the mutations are more composable
- this makes heavier (or more explicit/visible use) of redux toolkit's transitive dependency [immer](https://immerjs.github.io/immer/), which allows you to record mutations against an immutable object without actually mutating the underlying object. this is another thing that makes it seem more DB like, an immer instance around the entire app state looks a lot like a transaction :thinking: 
- from what i have seen, this is the less idiomatic redux pattern
  - more idiomatic would be to have code in each slice affected by an action handling that action, but this has such poor locality of behavior that i wanted to try this alternative use (which is still mentioned as an option by the official docs)
  - carissa also pointed out that this would not even be possible without moving the account slice into the mobile client, which i'm not opposed to doing, but feels like more evidence that this approach creates spaghetti interdependencies that we would like to avoid

### Checklist
- [x] organize the code a better
- [x] post associated client-shared PR
- [ ] write documentation and PR annotations to explain changes
- [ ] create some tests

### Related Issues
Fixes #2162 

### Verification steps

### DIFF
--- a/dev-client/src/model/project/projectGlobalReducer.ts
+++ b/dev-client/src/model/project/projectGlobalReducer.ts
@@ -1,0 +1,79 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {addUser, updateUsers} from 'terraso-client-shared/account/accountSlice';
+import * as projectService from 'terraso-client-shared/project/projectService';
+import {createAsyncThunk} from 'terraso-client-shared/store/utils';
+
+import {
+  updateMembership,
+  updateProjects,
+} from 'terraso-mobile-client/model/project/projectSlice';
+import {updateSites} from 'terraso-mobile-client/model/site/siteSlice';
+import {updateProjectSettings} from 'terraso-mobile-client/model/soilId/soilIdSlice';
+import {createGlobalReducer} from 'terraso-mobile-client/store/reducers';
+
+export const fetchProject = createAsyncThunk(
+  'project/fetchProject',
+  projectService.fetchProject,
+);
+
+export const addProject = createAsyncThunk(
+  'project/addProject',
+  projectService.addProject,
+);
+
+export const updateProject = createAsyncThunk(
+  'project/updateProject',
+  projectService.updateProject,
+);
+
+export const addUserToProject = createAsyncThunk(
+  'project/addUser',
+  projectService.addUserToProject,
+);
+
+export const projectGlobalReducer = createGlobalReducer(builder => {
+  builder.addCase(addUserToProject.fulfilled, (state, {meta, payload}) => {
+    updateMembership(state.project, {
+      projectId: meta.arg.projectId,
+      membership: payload.membership,
+    });
+    addUser(state.account, payload.user);
+  });
+
+  builder.addCase(fetchProject.fulfilled, (state, {payload}) => {
+    updateProjects(state.project, {[payload.project.id]: payload.project});
+    updateSites(state.site, payload.sites);
+    updateUsers(state.account, payload.users);
+    updateProjectSettings(state.soilId, payload.soilSettings);
+  });
+
+  builder.addCase(addProject.fulfilled, (state, {payload}) => {
+    updateProjects(state.project, {[payload.project.id]: payload.project});
+    updateSites(state.site, payload.sites);
+    updateUsers(state.account, payload.users);
+    updateProjectSettings(state.soilId, payload.soilSettings);
+  });
+
+  builder.addCase(updateProject.fulfilled, (state, {payload}) => {
+    updateProjects(state.project, {[payload.project.id]: payload.project});
+    updateSites(state.site, payload.sites);
+    updateUsers(state.account, payload.users);
+    updateProjectSettings(state.soilId, payload.soilSettings);
+  });
+});
--- a/dev-client/src/model/site/siteGlobalReducer.ts
+++ b/dev-client/src/model/site/siteGlobalReducer.ts
@@ -28,6 +28,7 @@ import {
   updateProjectOfSite,
   updateSites,
 } from 'terraso-mobile-client/model/site/siteSlice';
+import {deleteSoilData} from 'terraso-mobile-client/model/soilId/soilIdSlice';
 import {createGlobalReducer} from 'terraso-mobile-client/store/reducers';
 
 export const addSite = createAsyncThunk('site/addSite', siteService.addSite);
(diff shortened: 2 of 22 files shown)


dev-client/src/model/project/projectGlobalReducer.ts
```diff
@@ -0,0 +1,94 @@
...
+  {
+    membership: {
+      id: string;
+      userRole: ProjectMembershipProjectRoleChoices;
+      userId: string;
+    };
+    user: User;
+  },
+  ProjectAddUserMutationInput
+>('project/addUser', projectService.addUserToProject);
```
shrouxm, line 63: note to garo: clean up this jonk

dev-client/src/model/site/siteGlobalReducer.ts
```diff
@@ -0,0 +1,89 @@
...
+        siteId: payload.id,
+        projectId: payload.projectId,
+      });
+    }
+    updateSites(state.site, {[payload.id]: payload});
+  });
+
+  builder.addCase(deleteSite.fulfilled, (state, {payload}) => {
+    removeSiteFromAllProjects(state.project, payload);
+    deleteSites(state.site, [payload]);
```
shrouxm, line 74: note: should also delete associated soil data here

tm-ruxandra (approved):

PR #2333, opened 2024-10-15
feat: offline-mode soil updates

### LINKED ISSUE #2277: [Offline] Implement and test local soil data mutations
In order to support offline soil data collection, we must replicate our backend logic for soil data changes on the frontend. These GraphQL mutations will need to be replicated on the client side:

1) `SoilDataUpdateMutation`
2) `DepthDependentSoilDataUpdateMutation`
3) `SoilDataUpdateDepthIntervalMutation` (only if product decides to include depth intervals in MVP)
4) `SoilDataDeleteDepthIntervalMutation` (see above)

These should be stateless pure functions which take in the corresponding mutation input and a `SoilData` object, and return the resulting updated `SoilData` object, or return an error message if the backend would have returned an error message.

Create a service under `terraso-mobile-client/model/soilId` which implements this logic, and cover it thoroughly with unit tests. These should be drop-in replacements for the GraphQL calls; the offline feature flag should determine whether Redux invokes the GraphQL or local version of the mutation.

tm-ruxandra:
## Description

Implement and unit test client-side versions of the four soil data mutations: `updateSoilData`, `deleteSoilDataDepthInterval`, `updateSoilDataDepthInterval`, and `updateDepthDependentSoilData`. When the offline feature flag is turned on, the client will use the local logic rather than making graphQL calls, and the resulting data will be ingested by the redux state.

### Checklist
- [x] New tests added

### Related Issues
Fixes #2277

### Verification steps

Open the app with the offline feature flag turned on. Verify that all soil data actions (editing site settings, depth intervals, taking readings etc) behave as expected.


dev-client/src/model/soilId/actions/soilDataActionFields.ts
```diff
@@ -0,0 +1,105 @@
...
+import {
+  DepthDependentSoilData,
+  SoilData,
+  SoilDataDepthInterval,
+} from 'terraso-mobile-client/model/soilId/soilIdSlice';
+
+/**
+ * The soil data fields which are covered by the update action.
+ */
+export const SOIL_DATA_UPDATE_FIELDS = [
```
tm-ruxandra, line 33: I am open to a better means of enumerating these fields in the future (via some sort of codegen?), but this was the best option that the other devs and I could come up with.

dev-client/src/model/soilId/actions/soilDataActions.ts
```diff
@@ -0,0 +1,109 @@
...
+} from 'terraso-client-shared/graphqlSchema/graphql';
+import * as remoteSoilData from 'terraso-client-shared/soilId/soilDataService';
+import {SoilData} from 'terraso-client-shared/soilId/soilIdTypes';
+import {ThunkAPI} from 'terraso-client-shared/store/utils';
+
+import {isFlagEnabled} from 'terraso-mobile-client/config/featureFlags';
+import * as localSoilData from 'terraso-mobile-client/model/soilId/actions/localSoilDataActions';
+import {AppState} from 'terraso-mobile-client/store';
+
+export const updateSoilDataThunk = async (
```
tm-ruxandra, line 33: Unsure if it makes more sense to just merge the thunk methods with the non-thunk methods -- I suspect that the separation isn't necessary, but somehow this felt cleaner to me.

dev-client/src/model/soilId/actions/localSoilDataActions.ts
```diff
@@ -0,0 +1,160 @@
...
+} from 'terraso-mobile-client/model/soilId/soilIdFunctions';
+
+export const updateSoilData = (
+  input: SoilDataUpdateMutationInput,
+  data?: SoilData,
+): SoilData => {
+  const result = initializeResult(data);
+  updateFields(input, result, SOIL_DATA_UPDATE_FIELDS);
+
+  /* Specifying a depth interval preset clears all existing depth intervals */
```
knipec, line 47: Nit: Could be helpful to point to where the corresponding backend logic lives

dev-client/src/model/soilId/actions/localSoilDataActions.ts
```diff
@@ -0,0 +1,160 @@
...
+} from 'terraso-mobile-client/model/soilId/actions/soilDataActionFields';
+import {DEFAULT_DEPTH_INTERVAL_PRESET} from 'terraso-mobile-client/model/soilId/soilDataConstants';
+import {
+  compareInterval,
+  sameDepth,
+} from 'terraso-mobile-client/model/soilId/soilIdFunctions';
+
+export const updateSoilData = (
+  input: SoilDataUpdateMutationInput,
+  data?: SoilData,
```
knipec, line 42: Question: When would `data` not be supplied here (and in the other actions)?

knipec (approved):

dev-client/src/model/soilId/actions/localSoilDataActions.ts
tm-ruxandra: Good idea, will do

dev-client/src/model/soilId/actions/localSoilDataActions.ts
tm-ruxandra: It's a theoretically-possible edge case in redux (nothing guarantees that any given site ID will have soil data) but the backend also has support for handling cases where a site is missing soil data, so it seemed like a good idea to replicate that.

Issue #2232, opened 2024-10-03
[Offline] Create sync button & handle SiteNotesScreen's deleted required entities


knipec:
### Acceptance Criteria
- When the offline feature flag is on, a "Sync: pull" button appears on most screens after login
- Pressing the sync button will get the most recent data from the server
- If you are on the "Edit Site Note" screen:
  - If someone deletes the note and you pull, you are navigated to the site's list of notes
  - If someone deletes the site and you pull, you are navigated back the the sites list screen (formerly known as home screen)
  - (On Android you'll also see a snackbar notification -- I expect this to be temporary once the actual design gets decided and implemented)
- If you are on any of the site tabs (Site, Slope, Soil, or Notes) or the "Texture" screen for a site, and someone deletes the site, you are navigated back to the sites list screen

knipec:
Available to QA in v198

DerekCaelin:
@knipec Courtney and I tested this!
- We deleted a note while another person was looking at the note, and we were booted back to the notes tab as expected.
- We deleted a site after navigating to it from a project, and were booted back to the project screen
- We deleted a site after navigating to it from the sites screen, and we were booted back to the sites screen. 
- A time when it didn't behave as expected:
   - We created a project with three sites
   - Device A looked at the sites list on the project screen
   - Device B looked at the soils tab of a site in the project
   - Device A deleted the site from the project site list
   - Device B synced, saw no change (expected to be booted to the project site list)
   - Device A pulled, site reappeared (expected to remain deleted)

We didn't test what happens when a project gets deleted - should we?
We realized that the bottom sheet on the sites list is able to expand to cover the entire map again, which should have been addressed by #1483

knipec:
@DerekCaelin Huh, strange. I just went through the same steps and when Device B pulled, I did get booted to the project site list. And when Device B pulled, the site did not reappear.

knipec:
And no, I haven't implemented any logic for deleted projects yet.

And hmm, not sure about what's going on with #1483 -- maybe a question for Paul when he gets back unless you think it's urgent?

DerekCaelin:
We also tried to reproduce our wonky "undelete" experience and couldn't. Let's call this issue resolved and see if the issue happens again.

PR #2334, opened 2024-10-16
test: Add tests for required data logic

knipec:
## Description
- Add test for RestrictByRequirements
- Add EditSiteNoteScreen to just test that the screen handles its required data as intended

### Related Issues
Tests for #2232

### Verification steps
```
npm run test:integration -t 'RestrictByRequirements'
npm run test:integration -t 'EditSiteNoteScreen'
```


dev-client/__tests__/integration/RestrictByRequirements.test.tsx
knipec, line 1: Lmk if you prefer this to live under unit tests. It feels spiritually kinda like a unit test but uses render, so I feel pretty ambivalent about it

dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
```diff
@@ -0,0 +1,64 @@
+import {testState} from '@testing/integration/data';
```
paulschreiber, line 1: issue: add copyright notice.

dev-client/__tests__/integration/RestrictByRequirements.test.tsx
```diff
@@ -0,0 +1,83 @@
+import {Text} from 'react-native';
```
paulschreiber, line 1: issue: add copyright notice.

dev-client/__tests__/integration/RestrictByRequirements.test.tsx
tm-ruxandra: I think that until we better understand infrastructure for testing React and Redux components in isolation, it's safer to mark this as integration. We can always make it a unit test later if we decide it's fast enough / complies with whatever standards we come up with.

dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    ...actualNav,
+    useNavigation: () => ({navigate: mockedNavigate}),
+  };
+});
+
+afterEach(() => {
+  mockedNavigate.mockClear();
+});
+
+test('renders if data exists', () => {
```
tm-ruxandra, line 38: I think that these tests should be wrapped in a describe(), since afaik jest won't automatically label them with the source file.

tm-ruxandra (changes requested):
LGTM, one request re: how tests are structured.

dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const actualNav = jest.requireActual(
+    'terraso-mobile-client/navigation/hooks/useNavigation',
+  );
+  return {
+    ...actualNav,
+    useNavigation: () => ({navigate: mockedNavigate}),
+  };
+});
+
+afterEach(() => {
```
tm-ruxandra, line 34: I think this also needs to go inside the describe().

knipec: Since there's only one describe block in the test file, it shouldn't make a difference; it'll be called either way. Source: https://jestjs.io/docs/setup-teardown 

Since the `mockedNavigate` was declared outside the describe block (and didn't seem to work inside it for reasons I did not investigate) I put it in the same scope, but I really don't have strong feelings about it. Lmk if you still have a preference.

tm-ruxandra (approved):

Issue #2299, opened 2024-10-10
InfoSheet title can push close button partially offscreen


knipec:
### App Version

1.1 (195)

### Platforms

- [ ] iOS
- [X] Android
- [ ] iOS simulator
- [X] Android emulator

### Description

See title

### Steps to Reproduce

1. Set your phone language to Spanish
2. Go to an unaffiliated site
3. Go to its soil tab
4. Click on the filter/slider icon to the right of "Soil Pit"

-----
Or in English + dev environment: just change the string in 
```
  <InfoSheet
    heading={<TranslatedHeading i18nKey="soil.soil_preset.header" />}
```
from `soil.soil_preset.header` to something like `projects.search.no_matches`

### Expected behavior

X button is fully on the screen 

### Actual behavior

X button may be partially offscreen depending on the title

### Additional context

Noticed because in Spanish the title for the InfoSheet in SoilScreen (that houses EditSiteSoilDepthPreset) is long:
<img src="https://github.com/user-attachments/assets/f40801f8-f4b7-44df-9a6b-c355a32b0468" width=300>

When replacing the string (so not a real title in prod):
![Screenshot 2024-10-09 at 7 46 58 PM](https://github.com/user-attachments/assets/72f4ed34-ea72-41c6-9e97-6244b9d7866b)

knipec:
Triage notes: 
- Similar bug a while ago, we thought we'd fixed it -- @tm-ruxandra worked on it, should be a simple (potentially related: #1656, #1894)

tm-ruxandra:
The fix in this case is to add `flex: 1` to the stylesheet used to contain the heading element.

paulschreiber:
@DerekCaelin Please verify this in build 200 or later.

DerekCaelin:
This looks OK to me, over to @CourtneyLee333
<img width="200px" src="https://github.com/user-attachments/assets/82c7ed2b-9b44-4cf3-b8ac-8a2f78f8242f">

CourtneyLee333:
@DerekCaelin It doesn't look fixed in your screen shot. The button should not be up against or flowing over the edge of the screen. Above, Paul says this should be ready in v200. I only have v199 on Android.

DerekCaelin:
You're right. We don't have v200 yet.

DerekCaelin:
In v202, this appears to be resolved. Over to @CourtneyLee333 

[...]

PR #2342, opened 2024-10-18
fix: flex InfoSheet heading content

paulschreiber:
## Description
flex InfoSheet heading content — avoids pushing close button offscreen

### Related Issues
Addresses #2299.


tm-ruxandra (approved):
lgtm, don't forget to update snapshots

PR #2364, opened 2024-10-22
feat: soil change tracking

### LINKED ISSUE #2278: [Offline] Add basic change-tracking to soil data in Soil ID redux slice
Update the `soilID` Redux slice to include tracking of changes for Soil Data.

This should be architected similarly to the prototype in the `feat/local-storage-sync-prototype` branch created for #1932. Specifically, a `ChangeSet` type (or similar) should be introduced which allows tracking change metadata in a `Record` instance keyed by entity IDs. A field of this type should be added to the Redux slice. The `ChangeSet` type should have fields for tracking the following properties for soil data:
- Whether or not the site is known to have unsynced changes
- Timestamp of last sync (possibly never -- in fact, will never be set in this ticket since this ticket does not include the sync mechanism)
- Timestamp of last change (possibly never)

The `ChangeSet` should have its records updated accordingly any time a soil data change is completed in the Redux slice. If standard utility methods are needed (e.g., to update a `ChangeSet`'s timestamps for a given entity ID), those methods should be unit tested.

There should also be a Redux selector which allows code to select the site IDs which have currently-unsynced soil data changes.

tm-ruxandra:
## Description

Add a data model to track changes relevant to the sync system. Entities have change records containing a revision ID and information about the last change and last successful sync. The Soil ID slice tracks changes for soil data by updating the change records when mutations occur and marking them as synced when user data is loaded from the server. The number of un-synced sites is exposed through the sync button component for diagnostic purposes.

Also adds a basic system for patching data loaded from disk during startup -- ultimately we will want this to work by merging the slices' default initial states with the states saved to mmkv, however circular app dependencies prevented this and so we should revisit it as follow-on work once the main offline feature is done.

### Checklist
- [x] New tests added

### Related Issues
Fixes #2278

### Verification steps

Make changes to soil data and observe that the number of unsynced sites changes appropriately. Reload the user's data and observe that changes are marked as synced.

### DIFF
--- a/dev-client/src/model/sync/sync.ts
+++ b/dev-client/src/model/sync/sync.ts
@@ -1,0 +1,144 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+export type ChangeRecords<T> = Record<string, ChangeRecord<T>>;
+
+export type ChangeTimestamp = number;
+
+export type ChangeRevisionId = number;
+
+export type ChangeRecord<T> = {
+  /**
+   * Unique ID for the current state since the last sync, monotonically increasing for each change.
+   * Can be used to determine whether sync responses reflect stale data.
+   * Cleared for a successful sync.
+   * Default to zero for uninitialized records.
+   */
+  revisionId?: ChangeRevisionId;
+  lastModifiedAt?: ChangeTimestamp;
+
+  lastSyncedRevisionId?: ChangeRevisionId;
+  lastSyncedData?: T;
+  lastSyncedAt?: ChangeTimestamp;
+};
+
+export const INITIAL_REVISION_ID = 0;
+
+export const getRevisionId = <T>(
+  record?: ChangeRecord<T>,
+): ChangeRevisionId => {
+  return record?.revisionId ?? INITIAL_REVISION_ID;
+};
+
+export const nextRevisionId = (
+  revisionId: ChangeRevisionId,
+): ChangeRevisionId => {
+  return revisionId + 1;
+};
+
+export const getChanges = <T>(
+  records: ChangeRecords<T>,
+  ids: string[],
+): ChangeRecords<T> => {
+  return Object.fromEntries(ids.map(id => [id, getChange(records, id)]));
+};
+
+export const getChange = <T>(
+  records: ChangeRecords<T>,
+  id: string,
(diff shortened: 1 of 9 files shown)


dev-client/src/store/persistence.ts
```diff
@@ -38,3 +39,9 @@ export const loadPersistedReduxState = () => {
     );
   }
 };
+
+export const patchPersistedReduxState = (state?: Partial<AppState>) => {
+  if (state) {
+    merge(state, {soilId: {soilChanges: {}}});
```
tm-ruxandra, line 45: @shrouxm suggested pulling in the initial soil state here -- I agree that this is the better solution, however it is not possible to do this at the moment because of dependency issues in the soil slice. We should create a tech-debt ticket to design a system to handle this more generally for all redux slices and solve the underlying dependency problems.

dev-client/src/model/sync/sync.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export const getChange = <T>(
+  records: ChangeRecords<T>,
+  id: string,
+): ChangeRecord<T> => {
+  return records[id] ?? {};
+};
+
+export const markAllChanged = <T>(
+  records: ChangeRecords<T>,
```
shrouxm, line 68: question: thoughts on using `Draft<ChangeRecords<T>>`? implicitly that is what is happening when this method gets called via the soil ID slice, so it seems nice to be explicit about it (since it would probably be a bug if we were using this method on something that is _not_ a draft). i'm not sure whether that would make the test code more awkward though, and don't feel too strongly about it

dev-client/src/model/sync/sync.ts
```diff
@@ -0,0 +1,144 @@
...
+  const revisionId = nextRevisionId(prevRevisionId);
+  records[id] = {
+    revisionId: revisionId,
+    lastModifiedAt: at,
+
+    lastSyncedRevisionId: prevRecord?.lastSyncedRevisionId,
+    lastSyncedData: prevRecord?.lastSyncedData,
+    lastSyncedAt: prevRecord?.lastSyncedAt,
+  };
+};
```
shrouxm, line 93: question: am i remembering right that this version is meant to handle cases where the client makes further changes while a push or pull is in flight? i'm having trouble remembering the logic of how/in what order these methods are going to be called to handle that scenario (or am maybe misremembering whether this version is supposed to handle that scenario) and am wondering if you could write it up in a comment somewhere in this file

shrouxm (approved):
LGTM! 

a few take it or leave it comments mostly about documentation/code organization. i feel most strongly about the one asking for a write up how the data flow is supposed to work, but i might be misremembering what scenarios this code is supposed to handle in which case it's less relevant

dev-client/src/store/persistence.ts
```diff
@@ -38,3 +39,9 @@ export const loadPersistedReduxState = () => {
...
+  }
+};
```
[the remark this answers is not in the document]
tm-ruxandra: 1) I don't think that's necessary. I'd rather be able to test the load/save functionality in isolation and also test this in isolation. I don't think the issue of clients "remembering" to call this method is substantial since there is only one client (the app at startup).
2) Good idea, I'll make that change.

dev-client/src/model/sync/sync.ts
tm-ruxandra: I understand the principle of being more explicit about when we're operating on intermediate Redux state vs. immutable data. I'm hesitant to engineer a potentially-complex approach to that right now (as you note it could have consequences for testing) but in a subsequent PR I'd be open to it, once we fully understand how this code is used in a well-tested push implementation.

PR #2425, opened 2024-11-08
fix: rework soil cracks page helper text

paulschreiber:
## Description
Rework soil cracks page helper text.

After:
image

### Related Issues
Addresses https://github.com/techmatters/terraso-product/issues/976.

### DIFF
--- a/dev-client/src/translations/es.json
+++ b/dev-client/src/translations/es.json
@@ -737,6 +737,7 @@
             "soilTexture": "Incluye clase de textura y volumen de fragmentos de roca"
         },
         "vertical_cracking": {
+            "description": "Surface cracks develop in clay-rich soils after they dry out.",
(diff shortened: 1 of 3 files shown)


dev-client/src/translations/es.json
knipec, line 1: Why add english strings in the spanish file?

paulschreiber: We can't merge with checkMissingKeys failing. We will ensure they are translated.

knipec: Doesn't checkMissingKeys compare the keys of the json files? So because the keys exist in both en.json and es.json, wouldn't checkMissingKeys pass, even though the strings in es.json are in english?

paulschreiber: Yes. That's what we want.

knipec (approved):

Issue #2289, opened 2024-10-09
Refactor screens to use id's (not data) as props


knipec:
### Acceptance Criteria
- For each screen, it does not take Redux-selectable data as a prop
- Any Redux-selectable data should be accessed via the Redux selector, with an associated id as a prop to the screen component

### Context / Motivation
This prevents screens from using stale data, and allows us to specify `Requirements` on screens.

PR #2463, opened 2024-11-12
Refactor: make screens take IDs instead of data

knipec:
## Description
A few screens were taking data like User or Project. 
This means that if you were viewing the screen, a pull happened that deleted the data, the screen might not update and might cause errors.
Make them take userId or projectId, and retrieve the corresponding data from redux.

AddUserToProjectRoleScreen: took a User who maybe at that point had not been added to the cached set of users in redux's accountSlice. So this change will add a user to the `users` set after the prior screen first retrieves their info from the backend -- an intermediate part of the "Add user to project" flow -- instead of at the end of the flow. 

### Related Issues
Addresses #2289

### DIFF
--- a/dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
+++ b/dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
@@ -67,9 +67,10 @@ export const AddTeamMemberForm = ({projectId}: FormProps) => {
       // Success
       else {
         const user = userOrError as UserFields;
+        const userId = user.id;
         // Currently there is no need to access any user's preferences besides the current user
         dispatch(addUserToCache({...user, preferences: {}}));
-        navigation.navigate('ADD_USER_PROJECT_ROLE', {projectId, user});
+        navigation.navigate('ADD_USER_PROJECT_ROLE', {projectId, userId});
       }
     } catch (e) {
       console.error(e);
--- a/dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
+++ b/dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
@@ -21,11 +21,8 @@ import {Formik, FormikHelpers} from 'formik';
 import {Button} from 'native-base';
 import * as yup from 'yup';
 
-import {
-  checkUserInProject,
-  UserInProjectError,
-} from 'terraso-client-shared/account/accountService';
-import {addUserToCache} from 'terraso-client-shared/account/accountSlice';
+import {UserInProjectError} from 'terraso-client-shared/account/accountService';
+import {checkUserInProject} from 'terraso-client-shared/account/accountSlice';
 
 import {FormInput} from 'terraso-mobile-client/components/form/FormInput';
 import {Icon} from 'terraso-mobile-client/components/icons/Icon';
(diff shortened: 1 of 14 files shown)


dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
```diff
@@ -64,7 +67,10 @@ export const AddTeamMemberForm = ({projectId}: FormProps) => {
       // Success
       else {
         const user = userOrError as UserFields;
-        navigation.navigate('ADD_USER_PROJECT_ROLE', {projectId, user});
+        const userId = user.id;
+        // Currently there is no need to access any user's preferences besides the current user
```
knipec, line 71: I'm thinking of making a tech debt item to address this weirdness -- I don't think we should need to know any user preferences except for the current user, so perhaps the set of users should not include user preferences, and the current user's preferences should be stored in a different fashion?

dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
```diff
@@ -64,7 +67,10 @@ export const AddTeamMemberForm = ({projectId}: FormProps) => {
...
+        dispatch(addUserToCache({...user, preferences: {}}));
```
knipec, line 72: I'm thinking of making a tech debt item to address this weirdness -- I don't think the app should need to know any user preferences except for the current user... so perhaps the cached set of `users` in accountSlice should not include user preferences, and the current user's preferences should be stored in a different fashion?

dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
```diff
@@ -64,7 +67,10 @@ export const AddTeamMemberForm = ({projectId}: FormProps) => {
...
+        navigation.navigate('ADD_USER_PROJECT_ROLE', {projectId, userId});
```
tm-ruxandra, line 73: I think this should be in a `then` after the dispatch, since our thunks are technically all async calls?

dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
knipec: Talked with garo, will try to:
- change checkUserInProject to an async action so I can add the user there instead of here
- add tech debt items for not getting other users' preferences (and other user data that should potentially be private)

dev-client/src/screens/SiteNotesScreen/components/SiteInstructionsCard.tsx
```diff
@@ -18,25 +18,26 @@
...
 import {Row, Text} from 'terraso-mobile-client/components/NativeBaseAdapters';
 import {useNavigation} from 'terraso-mobile-client/navigation/hooks/useNavigation';
 
 type Props = {
-  siteInstructions: string;
+  project: Project;
 };
 
-export const SiteInstructionsCard = ({siteInstructions}: Props) => {
+export const SiteInstructionsCard = ({project}: Props) => {
```
tm-ruxandra, line 32: Wondering if we should rename this (in another ticket??) to like `PinnedNoteCard`; `SiteInstructions` tripped me up a little since it's a project being passed around.

dev-client/src/screens/SiteNotesScreen/ReadNoteScreen.tsx
```diff
@@ -27,15 +27,18 @@ import {
...
+import {selectProject} from 'terraso-mobile-client/store/selectors';
 
 type Props = {
-  content: string;
-  isSiteInstructions?: boolean;
+  projectId: string;
 };
 
-export const ReadNoteScreen = ({content, isSiteInstructions}: Props) => {
+export const ReadNoteScreen = ({projectId}: Props) => {
```
tm-ruxandra, line 37: Similar comment to below -- wondering if we could give this screen a more accurate name.

tm-ruxandra (changes requested):
Overall good, will approve once the notes re: checkUserInProject are resolved.

dev-client/src/screens/SiteNotesScreen/components/SiteInstructionsCard.tsx
knipec: Yeah, I was thinking about this too. The `SiteInstructions` name goes all the way to the backend, so was debating how much renaming to do and ended up on no renaming 😅  I'll rename a few things here, and make a tech debt ticket for changes that touch the other repos: https://github.com/techmatters/terraso-product/issues/1156. 

Oo I do like `PinnedNote`, and apparently there's already some backend project action that's called "EDIT_PINNED_NOTE".

dev-client/src/screens/AddUserToProjectScreen/components/AddTeamMemberForm.tsx
knipec: Resolving this, as the new changes remove the additional dispatch

Issue #2225, opened 2024-08-29
Input name and description updates: Texture and Slope


DerekCaelin:
Updates to the data inputs screen to improve the accuracy of the text.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=9786-16128&node-type=SECTION&t=a0WnIpL2C9k6IPvN-0)

### Site soil tab
- [x] Change `Soil texture` to `Texture class` (Spanish: `Clase de textura`)

<img width="374" alt="image" src="https://github.com/user-attachments/assets/5fd44b87-6cd8-4cb5-8044-fca85c05b64a">

### Soil Texture Input Screen
- [x] Change `Don't know your soil texture?` to `Don't know your texture class?`

### Guide Screen
- [x] Change `Use this texture` to `Use this texture class`

### Soil Texture Info Screen
- [x] Change all `soil texture` to `texture class`

### Project Inputs Tab
- [x] Add description to "Slope": `Includes slope steepness and shape`
- [x] Add description to "Soil Texture": `Includes texture class and rock fragment volume`

### Depth Options
- [x] Add description to "Soil Texture": `Includes texture class and rock fragment volume`

DerekCaelin:
@paulschreiber We've decided to implement the design [as seen here](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=9786-16128&node-type=section&t=KkvjDwtSJOBs4Qk5-0).
Data Collection requirement: Soil Texture
Soil tab: `Texture Class` and `Rock Fragment`

PR #2475, opened 2024-11-13
fix: correct strings for texture class/soil texture

paulschreiber:
## Description
Correct strings for texture class/soil texture. Courtney and I went over the screens and text together. What's in the build is now correct (and supersedes any discrepencies in #2225 or Figma).

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2225.

### DIFF
--- a/dev-client/src/translations/es.json
+++ b/dev-client/src/translations/es.json
@@ -715,9 +715,12 @@
             "required_data_title": "Requisitos para la recogida de datos"
         },
         "required_method": "{{method}} (obligatorio)",
+        "collection_method_summary": {
+            "soilTexture": "Clase de textura"
+        },
         "collection_method": {
             "slope": "Incluye la inclinación, forma, aspecto y posición del pendiente del suelo.",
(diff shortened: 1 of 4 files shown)


dev-client/src/translations/es.json
```diff
@@ -715,9 +715,12 @@
             "required_data_title": "Requisitos para la recogida de datos"
         },
         "required_method": "{{method}} (obligatorio)",
+        "collection_method_summary": {
+            "soilTexture": "Clase de textura"
+        },
         "collection_method": {
             "slope": "Incluye la inclinación, forma, aspecto y posición del pendiente del suelo.",
-            "soilTexture": "Incluye clase de textura, porcentaje de arcilla y fragmentos de roca",
+            "soilTexture": "Textura del suelo",
```
shrouxm, line 723: question: is there work tracked somewhere to also update slope in a similar way? or is it not a problem because of how they're displayed?

shrouxm (approved):
LGTM, just a quick question

Issue #2376, opened 2024-10-24
Offline UI: prevent logging out/deleting account


DerekCaelin:
When a user is offline, disable the option to sign out and delete account.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=9314-5129&node-type=canvas&t=aNOkjDj8zfXau8az-0)
<img width="354" alt="image" src="https://github.com/user-attachments/assets/54054e02-cf75-42e6-a820-ddc980c83e13">

## Acceptance criteria
- [ ] If offline, on settings screen
   - [x] disable delete account button
   - [ ] disable sign out button

shrouxm:
should be QA-able in versions 210 and above

DerekCaelin:
@shrouxm Both buttons appear to be disabled, but I can interact with "sign out"

shrouxm:
🤔 must be because it opens a modal. will have to look into it when i get back

DerekCaelin:
Unassigning Garo while they are out - any engineer can pick this up.

DerekCaelin:
@paulschreiber Both buttons appear to be disabled, but I can interact with "sign out"

paulschreiber:
@DerekCaelin this is pending #2594. Should be in 219 or later.

DerekCaelin:
This meets my expectation

PR #2484, opened 2024-11-14
feat: disable user settings buttons when offline

shrouxm:
## Description
Disables the sign out and delete account buttons on the user settings screen when offline.

@knipec `useIsOffline` returns `boolean | null`, should it just return `boolean`? or is the intention for `null` to mean "unknown" and different components might want to handle that case differently?

### Checklist

### Related Issues
Related to #2376 

### Verification steps
Buttons should be disabled when offline.

### DIFF
--- a/dev-client/src/screens/UserSettingsScreen/components/menu/SignOutItem.tsx
+++ b/dev-client/src/screens/UserSettingsScreen/components/menu/SignOutItem.tsx
@@ -24,7 +24,7 @@ import {useIsOffline} from 'terraso-mobile-client/hooks/connectivityHooks';
 export function SignOutItem() {
   const {t} = useTranslation();
 
-  const isDisabled = useIsOffline() === true;
+  const isDisabled = useIsOffline();
 
   return (
     <SignOutModal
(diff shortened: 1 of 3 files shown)


dev-client/src/screens/UserSettingsScreen/components/menu/SignOutItem.tsx
```diff
@@ -19,10 +19,13 @@ import {useTranslation} from 'react-i18next';
 
 import {MenuItem} from 'terraso-mobile-client/components/menus/MenuItem';
 import {SignOutModal} from 'terraso-mobile-client/components/modals/SignOutModal';
+import {useIsOffline} from 'terraso-mobile-client/hooks/connectivityHooks';
 
 export function SignOutItem() {
   const {t} = useTranslation();
 
+  const isDisabled = useIsOffline() === true;
```
paulschreiber, line 27:
```suggestion
  const isDisabled = useIsOffline();
```

knipec (approved):
Yeah, it returns null because netinfo can return null. I believe I've only seen it happen for a brief period of time at the start, presumably because it doesn't yet have the info, presumably because it's not awaiting the network call it uses to check for reachability.

EDIT: Most of the time I imagine the behavior as similar to "offline" so maybe actually disable it in the null case as well. Hmm maybe it should have been isOnline so it's more intuitive to handle null and false similarly

Issue #1776, opened 2024-07-08
Make tooltip component consistent with design


CourtneyLee333:
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7281-34294&t=FWzt402zPDCk2duY-1)

Ensure that tooltips can be associated with components besides just the help icon (e.g., design has example future usage of tooltip on a pill.)

tm-ruxandra:
Ensure that tooltips can be associated with components besides just the help icon (e.g., design has example future usage of tooltip on a pill.)

shrouxm:
@DerekCaelin this issue is marked as blocked, but the figma link says "ready for dev", do you know if there's any reason I shouldn't start working on this?

tm-ruxandra:
@shrouxm I originally marked it blocked because I had issues with implementing it to match the specified design and we decided to defer it until we had the time to investigate. It's fine for you to take this ticket.

shrouxm:
can be QA'd in versions 212 and above!

DerekCaelin:
@CourtneyLee333 your thoughts?

CourtneyLee333:
We're only in v211, so I'll test this at next release.

CourtneyLee333:
@shrouxm I can't tell if this has proper padding, but it looks too tight. Does it have 8px left and right, and 4px top and bottom?

DerekCaelin:
Marking as "blocked" while Garo is OoO.

shrouxm:
@CourtneyLee333 just went back and double checked the code, definitely specified 8px horizontal padding and 4px vertical padding. i can bump it up a bit if it's seeming a bit tight on a real screen?

shrouxm:
we agreed this looks good as is in this thread in Figma; https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN?node-id=7281-34294#1106025027

PR #2499, opened 2024-11-19
feat: tooltip component design updates

shrouxm:
## Description
Updates the `Tooltip` component to match the new design spec. Waiting on confirmation of a couple things in [this Figma thread](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN?node-id=7281-34294&m=dev#1032530124) before fully ready to merge.

### Checklist

### Related Issues
Related to #1776 

### Verification steps
Tooltip component should look like the design. I believe the only example right now is on the `?` on the create site screen.

### DIFF
--- a/dev-client/src/components/tooltips/Tooltip.tsx
+++ b/dev-client/src/components/tooltips/Tooltip.tsx
@@ -28,7 +28,11 @@ type TooltipProps = React.PropsWithChildren<{
 export const Tooltip = memo(({trigger, children}: TooltipProps) => {
   return (
     <Popover trigger={trigger}>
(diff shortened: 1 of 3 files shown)


shrouxm:
`react-native-paper` doesn't offer a tooltip component which is able to match the spec, so it would increase the effort required to find and vet another library and/or implement it ourselves. this doesn't further entrench us in nativebase and it at least gets us compliant with the spec, so i don't see any reason to couple that work together.

tm-ruxandra (approved):
My only comment would have been to get us off of NativeBase if possible, but maybe that just isn't a practical goal right now.

Issue #2517, opened 2024-10-11
Change "Sites List" to "+ Create Sites" for users without sites, add updated guidance content.


DerekCaelin:
In usability tests, several first-time user testers expressed confusion about how to create a site. They eventually figured it out, but we decided to reduce friction by adding some guidance in the bottom sheet. This will replace the "Sites List" with "+ Create Sites" until the user has a site to list, and update the guide text.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10731-13831&t=tODMlRhz2JF7lXl3-0)

<img src="https://github.com/user-attachments/assets/5f442c9b-1576-422b-bd42-71cd74306b1e" width="300px">

## Acceptance Criteria
- [ ] For users without sites, set title of bottom sheet to "+ Create Site"
- [ ] Update guidance text.
- [ ] This experience would be replaced with the default experience once the user has a site

paulschreiber:
Please verify in 217 or later.

DerekCaelin:
@CourtneyLee333 reviewed and approved.

PR #2518, opened 2024-11-22
fix: redesign empty sites sheet

paulschreiber:
## Description
Update empty sites sheet with new design.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2517

### DIFF
--- a/dev-client/src/screens/SitesScreen/components/EmptySiteMessage.tsx
+++ b/dev-client/src/screens/SitesScreen/components/EmptySiteMessage.tsx
@@ -14,35 +14,51 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
-import {Trans} from 'react-i18next';
 
-import {BulletList} from 'terraso-mobile-client/components/BulletList';
-import {LocationIcon} from 'terraso-mobile-client/components/icons/LocationIcon';
+import {StyleSheet} from 'react-native';
+
+import {TranslatedParagraph} from 'terraso-mobile-client/components/content/typography/TranslatedParagraph';
+import {Icon} from 'terraso-mobile-client/components/icons/Icon';
 import {Text, View} from 'terraso-mobile-client/components/NativeBaseAdapters';
 
 export const EmptySiteMessage = () => {
   return (
     <View px="17px">
-      <Text variant="body1">
-        <Trans
-          i18nKey="site.empty.info"
-          components={{
-            bold: <Text bold />,
-          }}
-        />
+      <Text bold>
+        <TranslatedParagraph i18nKey="site.empty.info" />
       </Text>
 
-      <BulletList
-        data={[1, 2, 3]}
-        renderItem={i => (
-          <Text variant="body1" color="text.primary">
-            <Trans
-              i18nKey={`site.empty.bullet_${i}`}
-              components={{icon: <LocationIcon />}}
-            />
-          </Text>
-        )}
-      />
+      <View style={styles.view}>
+        <Icon name="tap-and-play" color="text.icon" size={14} pr={5} />
+        <TranslatedParagraph i18nKey="site.empty.map" />
+      </View>
+
+      <View style={styles.view}>
+        <Icon name="my-location" color="text.icon" size={14} pr={5} />
+        <TranslatedParagraph i18nKey="site.empty.icon" />
+      </View>
+
+      <View style={styles.enter}>
+        <Icon name="search" color="text.icon" size={14} pr={5} marginTop={1} />
+        <TranslatedParagraph i18nKey="site.empty.coords" />
+      </View>
+
+      <View style={styles.enter}>
+        <TranslatedParagraph i18nKey="site.empty.summary" />
+      </View>
     </View>
   );
 };
+
+const styles = StyleSheet.create({
+  view: {
+    flexDirection: 'row',
+    marginTop: 10,
+    alignItems: 'center',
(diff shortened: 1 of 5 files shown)


dev-client/src/screens/SitesScreen/components/EmptySiteMessage.tsx
```diff
@@ -14,35 +14,51 @@
...
-          <Text variant="body1" color="text.primary">
-            <Trans
-              i18nKey={`site.empty.bullet_${i}`}
-              components={{icon: <LocationIcon />}}
-            />
-          </Text>
-        )}
-      />
+      <View style={styles.view}>
+        <Icon name="tap-and-play" color="text.icon" size={14} pr={5} />
```
paulschreiber, line 32: issue: this should be `gesture-tap`, but that name is not valid.

dev-client/src/screens/SitesScreen/components/EmptySiteMessage.tsx
knipec, line 1: Is this intended to complete ticket #2517? 
If so, seems like it's missing the "+", missing the offline message, and the icons are smaller than in the Figma design. 
If not, for the future when you're implementing something in multiple pieces could you clarify that in the PR description?

Screenshot 2024-12-03 at 4 47 39 PM

dev-client/src/screens/SitesScreen/components/EmptySiteMessage.tsx
```diff
@@ -14,35 +14,68 @@
...
+      </View>
+
+      <View style={styles.view}>
+        <Icon name="my-location" color="text.icon" size={24} pr={5} />
+        <TranslatedParagraph i18nKey="site.empty.icon" />
+      </View>
+
+      <View style={styles.enter}>
+        <Icon name="search" color="text.icon" size={24} pr={5} marginTop={1} />
+        <TranslatedParagraph i18nKey="site.empty.coords" />
```
knipec, line 59: Issue: Hmm, looks like the text wrapping point is further to the right than we want ("the" should probably wrap):
Screenshot 2024-12-04 at 12 22 34 PM

dev-client/src/screens/SitesScreen/components/EmptySiteMessage.tsx
```diff
@@ -14,35 +14,68 @@
...
-          <Text variant="body1" color="text.primary">
-            <Trans
-              i18nKey={`site.empty.bullet_${i}`}
-              components={{icon: <LocationIcon />}}
-            />
-          </Text>
-        )}
-      />
+      <View style={styles.view}>
+        <MaterialCommunityIcons
```
knipec, line 43: Question: Any chance MaterialCommunityIcons is a superset of MaterialIcon? If so, could we use that in our custom `Icon` component? It just feels nicer to have one unified way to refer to icons if possible.

dev-client/src/translations/en.json
```diff
@@ -49,14 +49,16 @@
...
-            "bullet_3": "Enter coordinates, address, or other search term in the search bar."
+            "info": "<bold>Three ways to create a site:</bold>\n",
+            "map": "Tap anywhere on the map.",
+            "icon": "Tap the My Location icon.",
+            "coords": "Enter coordinates, address, or search term in the map search bar",
+            "summary": "A site lets you collect and save observations, learn more about the land and its potential, and share with others."
         },
         "create": {
-            "title": "Create Site",
+            "title": "+ Create Site",
```
knipec, line 61: Issue: Why did a plus get added here? As far as I can tell this string is only used for the "Create site" button, which shouldn't get a plus next to it.

dev-client/src/screens/SitesScreen/components/SiteListBottomSheet.tsx
knipec, line 1: Issue: Wait, doesn't this change the title even when the site list isn't empty? The [figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10463-16911&node-type=frame&t=WMKu4AOkqw4SpNBZ-0) looks like it should only say "+ Create Site" when the site list is empty, and "Sites List" otherwise, right?

dev-client/src/screens/SitesScreen/components/EmptySiteMessage.tsx
paulschreiber: They appear to be separate. @expo/vector-icons includes a few different icons sets.

dev-client/src/components/content/typography/TranslatedParagraph.tsx
```diff
@@ -31,8 +33,13 @@ export function TranslatedParagraph({
...
   return (
-    <Text variant={variant}>
+    <Text variant={variant} style={styles.text}>
       <TranslatedContent i18nKey={i18nKey} values={values} />
     </Text>
   );
 }
+const styles = StyleSheet.create({
+  text: {
+    flex: 1,
```
knipec, line 43: Ah, this again 🙃 
I don't understand this enough to know if/how it would affect anywhere else that TranslatedParagraph is used -- do you or @tm-ruxandra?

tm-ruxandra: This is not an appropriate solution. We should not be putting such a specific fix on such a general component. @paulschreiber please rework to wrap affected paragraphs in a `View` with the necessary styling.

tm-ruxandra (approved):

PR #2550, opened 2024-11-25
fix: handle additional elevation error

paulschreiber:
## Description
handle additional elevation error.


dev-client/src/model/elevation/elevationService.ts
```diff
@@ -39,7 +42,10 @@ export const getElevation = async (
       `https://epqs.nationalmap.gov/v1/json/?${queryString}`,
     );
     const result = await response.text();
-    if ([OUT_OF_RANGE_MESSAGE, UNABLE_TO_PROCESS_MESSAGE].includes(result)) {
+    if (
+      [OUT_OF_RANGE_MESSAGE, UNABLE_TO_PROCESS_MESSAGE].includes(result) ||
```
tm-ruxandra, line 46: The increasing complexity here suggests to me that this should become a method called like `isErrorResult` or similar.

tm-ruxandra (approved):

Issue #2292, opened 2024-10-09
[Offline] When app starts or user comes back online, I see updated server data


knipec:
### Acceptance criteria
- When:
  - app starts and is online, or 
  - user returns from offline to online
- Then:
  - pull (after push has completed, if there was a push)
- If you have a spotty connection, don't do this repeatedly over and over (I think the debouncing for push would apply here too)

knipec:
This and #2291 should be in future version 216

DerekCaelin:
@knipec has something changed here? When Courtney and I tested offline data collection on 11/14 we found a push and pull when coming from offline to online.

knipec:
@DerekCaelin Hmm, do you know if it did a full pull of all the data? The way the push system works I'd expect the app to receive updated data from server, but only for the sites that had changes to push. So if on 11/14 you didn't also check sites that didn't have pushed changes, I suspect they weren't getting updated data when coming back online yet.

DerekCaelin:
@knipec does this test sound right?
- User A and User B are in a project with sites 1 and 2
- User A goes offline
- User A adds observation to site 1
- User A does nothing to site 2
- User B adds observation to site 2
- User A comes back online

Expected outcome
- User A should see their change reflected in site 1
- User A should see User B's edits to site 2

knipec:
@DerekCaelin Yeah, that sounds right to me!

DerekCaelin:
Courtney and I tested, and this works.

Issue #2291, opened 2024-10-09
[Offline: Pull] When online, I see up-to-date data pulled from server every [?] seconds


knipec:
### Acceptance Criteria
- Decide frequency (Derek says 5 minutes :))
- When offline, no pulls are attempted
- When online, user sees data updated at that frequency 
  - Don't pull when the user has unsaved changes in the syncManager
  - Check if this will re-trigger soil ID matches (maybe this is fine either way, but would be nice to do so if and only if data has changed)
- When app is active (in the foreground), pulls happen. Otherwise, no pulls happen

- When verifying, do a quick check of data usage to make sure I wasn't totally off in my estimates

Non-goals:
- Any indicator that a pull is happening or has happened (besides that your data may update)

### Notes
- Very rough data usage ballpark: Looks like a site's data is around 1-2 KB, project's data around: 1-2 KB. Pulling every minute if you have 10 sites would take estimated ~50 minutes to hit 1MB data usage.

DerekCaelin:
every 5 mins while app is in foreground?

knipec:
This and #2292 should be in future v216

PR #2557, opened 2024-11-26
feat: Add automatic pull mechanism

knipec:
## Description
- `PullRequester` will request (but not execute) a pull. It will update redux state in a new `sync` slice to note that a pull is now requested.
- `PullDispatcher` will decide to do a pull when a pull has been requested, and if it detects all the other conditions are right.

### Related Issues
Implements #2291 and #2292 (which, apologies, should really be one issue)

### Verification steps
- Pull should happen on open, when app goes from offline to online (with 500ms debounce), following a push that returned errors, and every 5 minutes.
- Pull should not happen before login, when offline, when there are local changes that have not yet been pushed, or when the app is not active in the foreground.


dev-client/src/store/reducers.ts
```diff
@@ -39,6 +40,7 @@ const sliceReducers = {
   site: siteReducer,
   project: projectReducer,
   soilId: soilIdReducer,
+  sync: syncReducer,
```
tm-ruxandra, line 43: Don't forget to add the sync slice to the load-from-storage routine in `persistence`.

dev-client/src/store/sync/hooks/syncHooks.tsx
```diff
@@ -71,3 +77,21 @@ export const useRetryInterval = (interval: number, action: () => void) => {
 
   return {beginRetry, endRetry};
 };
+
+export const useSyncErrorSiteIds = () => {
+  return Object.keys(useSelector(selectSyncErrorSites));
```
tm-ruxandra, line 82: This will return a new array for every render; better to just create a new memoized selector with `createSelector` in the soil selectors file.

dev-client/src/store/sync/hooks/syncHooks.tsx
```diff
@@ -71,3 +77,21 @@ export const useRetryInterval = (interval: number, action: () => void) => {
...
+};
+
+export const usePullDispatch = () => {
+  const dispatch = useDispatch();
+
+  const currentUserID = useSelector(selectCurrentUserID);
+
+  return useCallback(() => {
+    if (currentUserID !== undefined) {
```
tm-ruxandra, line 91: I think it would be better to have the logic around `currentUserId` exist outside of the hook here. Similarly I'm a little uncomfortable with the coupling of dispatching `setPullRequested` and `fetchSoilDataForUser`; I think it'd be more appropriate to have each `dispatch` in its own mini hook and still use the logic in the component to coordinate them + branch on the user ID. This is because these hooks files aren't covered by tests, and we are using them to make the actual logic of these processes easier to test. (Happy to discuss further if this is a confusing comment).

tm-ruxandra (changes requested):
Overall looked good, had two comments that I think we should resolve before merging.

dev-client/src/store/sync/hooks/syncHooks.tsx
knipec: Ah, thanks for catching that!
The [Redux docs](https://redux.js.org/usage/writing-tests#:~:text=Do%20not%20try%20to%20mock,actual%20app%20code%20is%20working.) advise not mocking selector functions in tests (though mocking this hook that just wraps a redux selector may also not be in the spirit of the advice...?)
I'll plan to add the new memoized selector, and use it in `useSyncErrorSiteIds` if that sounds good?

dev-client/src/store/sync/hooks/syncHooks.tsx
knipec: That makes sense, thanks!

knipec: FYI my last 3 commits should address your 3 comments, lmk if I misunderstood anything though!

dev-client/src/hooks/appStateHooks.ts
```diff
@@ -0,0 +1,22 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {AppState} from 'react-native';
+
+export const useAppState = () => {
+  return AppState.currentState;
```
knipec, line 21: Issue: Make this update when the current state updates

shrouxm (commented):
i'm sorry, i didn't end up having the brain to finish reviewing this. leaving my one comment from when i had brain

tm-ruxandra (approved):

Issue #2560, opened 2024-10-23
Offline UI: Temporary Site Popover - Soil ID data not available offline, disable buttons


DerekCaelin:
Changes to the popover for temporary sites.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10368-5034&node-type=instance&t=aNOkjDj8zfXau8az-0)
<img width="324" alt="image" src="https://github.com/user-attachments/assets/7d316daa-6daf-4d0e-921e-c16ff3c90cc0">

<img width="397" alt="image" src="https://github.com/user-attachments/assets/76782daa-baad-418e-a430-bdda6cace61e">

## Acceptance criteria
- [ ] When offline, and the user taps the map, display `"NOT AVAILABLE OFFLINE"` for Top Matches and Elevation. 
- [ ] Disable "Create Site" and "Learn More" buttons

paulschreiber:
Please verify in 216 or later.

DerekCaelin:
This meets my expectations.

CourtneyLee333:
Niiiiice.

Issue #2372, opened 2024-10-23
Offline UI: Alert: Soil Properties Score not avillable offline


DerekCaelin:
When offline, make an alert saying the soil properties score will be used offline.

References:
- https://github.com/techmatters/terraso-product/issues/1117

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=9314-5377&node-type=frame&t=aNOkjDj8zfXau8az-0)
<img width="405" alt="image" src="https://github.com/user-attachments/assets/52463ea5-d5c1-43cc-b591-3eff64cffb3f">

## Acceptance Criteria
- [ ] If the user is offline and views the Soil ID screen, show the most recent soil property score calculated (do not try to update the soil property score) and display this error.

DerekCaelin:
I added more information in the acceptance criteria.

paulschreiber:
@DerekCaelin the UI part is done and you can verify that in 217. @tm-ruxandra will implement the score data separately. (Don't close out the issue until both parts are done.)

DerekCaelin:
OK, I will review when both parts are finished

CourtneyLee333:
Looks good!

PR #2561, opened 2024-11-27
fix: temporary site popover: disable when offline

paulschreiber:
## Description
On the temporary site popover, disable data fetching and site creation when offline.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2560

### DIFF
--- a/dev-client/locales/po/en.po
+++ b/dev-client/locales/po/en.po
@@ -127,6 +127,9 @@ msgstr "{{value}} {{units}}"
 msgid "site##elevation_unknown"
 msgstr "unknown"
 
+msgid "site##not_available_offine"
+msgstr "Not available offline"
+
 msgid "site##empty##info"
 msgstr ""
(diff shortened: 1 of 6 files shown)


knipec (commented):
Issue: it appears there are still backend queries being attempted when the user clicks a point on the map to bring up the temporary site popover (and therefore console errors are visible in dev builds). I feel like we should make sure to avoid any queries associated with the temporary site popover when offline.

tm-ruxandra (changes requested):
PO files need to be updated, otherwise seems reasonable

paulschreiber:
@tm-ruxandra I updated the PO files. Do I need to do additional work to ensure I'm not making calls to the soil id service when offline?

tm-ruxandra:
@paulschreiber I think I lost track of which ticket is which - this ticket is fine standalone since it didn't cover disabling soil ID loading offline, but the other ticket (#2372) probably needs to be reworked.

Issue #2373, opened 2024-10-24
Offline UI: Projects not available offline


DerekCaelin:
While offline, hide everything on the Projects screen and display an alert

References:
- https://github.com/techmatters/terraso-product/issues/1117

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10346-10425&node-type=section&t=aNOkjDj8zfXau8az-0)
<img width="428" alt="image" src="https://github.com/user-attachments/assets/65503873-0cf0-44e7-86ba-cb478f5c22c4">

paulschreiber:
Please verify in 217 or later.

DerekCaelin:
Looks great

CourtneyLee333:
It looks like the paragraph text isn't using the color specified in the design. 
The paragraph is missing a period at the end.

paulschreiber:
The period and color are fixed in 218.

CourtneyLee333:
Closing this, with the expectation that the details will get fixed in the component implementation issue.

PR #2566, opened 2024-11-27
fix: add offline alert to soil properties and project screens

paulschreiber:
## Description
Add offline alert to soil properties screen.
Move `<ConnectivityContextProvider>` higher up in hierarchy to support this.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2372 and https://github.com/techmatters/terraso-mobile-client/issues/2373


tm-ruxandra (changes requested):
Overall looks good, don't forget to update PO files.

tm-ruxandra (changes requested):
After re-reviewing the main ticket: this does not implement the acceptance criteria. The AC are that:
1) The existing soil properties scores are retained while offline
2) The alert is displayed in addition to those scores/table data etc
3) Entering new soil data does not discard old soil ID data (this ticket doesn't address this at all)

I don't have a succinct way of explaining how to properly implement this right now; the ticket was marked as 5 and probably should have been assigned to me because of its complexity.

PR #2569, opened 2024-11-29
feat: list/rank soil inconsistency workaround

shrouxm:
## Description
Changes the `fetchLocationBasedSoilMatches` redux action to secretly call the `fetchDataBasedSoilMatches` endpoint with an empty set of data. See discussion on https://github.com/techmatters/soil-id-algorithm/issues/126 for why this is desirable behaviorally. This is in draft because it puts the codebase in a kinda hacky state where the method names aren't really honest anymore, but I wanted to record my progress before I take time off.

### Checklist

### Related Issues
Related to https://github.com/techmatters/soil-id-algorithm/issues/126. Depends on https://github.com/techmatters/terraso-client-shared/pull/1112.

### Verification steps
There should be no difference in the soil ID results now before or after creating a site.

### DIFF
--- a/dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
+++ b/dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
@@ -37,7 +37,9 @@ export const fetchLocationBasedSoilMatchesThunk = async (
 ) => fetchLocationBasedSoilMatches(coords);
 
 export const fetchLocationBasedSoilMatches = async (coords: Coords) => {
-  const result = await soilIdService.fetchLocationBasedSoilMatches(coords);
+  const result = await soilIdService.fetchDataBasedSoilMatches(coords, {
+    depthDependentData: [],
+  });
 
   if (result.__typename === 'SoilIdFailure') {
     return locationEntryForStatus(coords, result.reason);
--- a/dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
+++ b/dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
@@ -37,6 +37,10 @@ export const fetchLocationBasedSoilMatchesThunk = async (
 ) => fetchLocationBasedSoilMatches(coords);
 
 export const fetchLocationBasedSoilMatches = async (coords: Coords) => {
+  // NOTE: we call the dataBasedSoilMatches endpoint here as a temporary workaround
+  //       to make the pre and post site creation soil lists consistent. The ideal
+  //       upstream fix would be to actually make the results of these endpoints
+  //       consistent for a fresh site
   const result = await soilIdService.fetchDataBasedSoilMatches(coords, {
     depthDependentData: [],
   });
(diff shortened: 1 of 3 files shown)


dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
knipec, line 1: This means that we're not ever calling the location-based endpoint anymore, right?

dev-client/src/model/soilIdMatch/soilIdMatches.test.tsx
```diff
@@ -64,12 +66,17 @@ describe('locationEntryForStatus', () => {
...
-      locationEntryForMatches({latitude: 1, longitude: 2}, ['match'] as any),
+      locationEntryForMatches(
+        {latitude: 1, longitude: 2},
+        inputMatches as DataBasedSoilMatch[],
+      ),
     ).toEqual({
       input: {latitude: 1, longitude: 2},
       status: 'ready',
-      matches: ['match'],
+      matches: [{match, dataSource: 'SSURGO'}],
```
knipec, line 79: Question: If the `matches` are of type `LocationBasedSoilMatch[]`, and `distanceToNearestMapUnitM` and `soilInfo` properties are not listed as optional on `LocationBasedSoilMatch`, why wouldn't the type system complain when the matches here don't have them?

dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
```diff
@@ -37,7 +37,9 @@ export const fetchLocationBasedSoilMatchesThunk = async (
 ) => fetchLocationBasedSoilMatches(coords);
 
 export const fetchLocationBasedSoilMatches = async (coords: Coords) => {
-  const result = await soilIdService.fetchLocationBasedSoilMatches(coords);
+  const result = await soilIdService.fetchDataBasedSoilMatches(coords, {
```
knipec, line 40: Since the PR description refers to this as a "kinda hacky state where the method names aren't really honest anymore", are you hoping to update names or keep them like this going forward?
If you plan to keep the names like this, I'd maybe put a comment here like "Use data-based endpoint so that soil id results are consistent before and after creating a site". And if there's a task to update the soil-id-algorithm so that the location-based one gives the same result as this, maybe link to it?

dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
shrouxm: yes, if this wasn't a temporary workaround i would suggest removing it, and we might end up doing so in the future once we see how things actually shake out

dev-client/src/model/soilIdMatch/soilIdMatches.test.tsx
shrouxm: i was sort of unthinkingly emulating the previous code, which wast just casting `['match']` to `any`. but i think the argument to `toEqual` can just be any old type

dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
shrouxm: that makes sense, will do!

shrouxm: done, also documented on the backend for good measure: https://github.com/techmatters/terraso-backend/pull/1614

knipec (approved):

Issue #2324, opened 2024-10-14
Sync Conflict Notification (in a few cases)


DerekCaelin:
In the case that there is an issue while syncing, display a notification. 

Sync errors can occur when the user client `pushes` data and 
- A site was deleted or removed from a project
- The user was removed from a project
- Depth requirements for a project changed

We will also display this when a `pull` occurs and a user could be looking at content to which they no longer have access.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10659-8623&node-type=frame&t=GgmGoTssg8eIF1h9-0)

<img width="392" alt="image" src="https://github.com/user-attachments/assets/9146beee-fd50-488c-9b46-c1dde7c34439">

## Acceptance Criteria
- [x] For this story: If the user client **pulls** and the user is looking at something they can no longer access, display the dialog in the cases that are already implemented: removing a site or site note on the "edit site note" screen, removing a site on the texture screen or one of the site tabs
- [x] Link to: https://landpks.terraso.org/help/project-conflict/

### Non-goals
- Display the dialog for all **pull** cases when user is looking at a site or project they can no longer access (will be covered in [#2290](https://github.com/techmatters/terraso-mobile-client/issues/2290))
- If the user client **pushes** data and 
   - A site was deleted or removed from a project
   - The user was removed from a project
   - Depth requirements for a project changed
....then display the dialog (will be covered in [#2293](https://github.com/techmatters/terraso-mobile-client/issues/2293))

knipec:
Should be done after #2279 
Could be done after or alongside #2293

tm-ruxandra:
This will be testable when #2293 is out in a build as well.

CourtneyLee333:
I'm assuming this will be in v217.

DerekCaelin:
@tm-ruxandra this notification appears when expected and passes the design review. Please update the "support" link to go here: https://landpks.terraso.org/help/synchronization-conflict/

tm-ruxandra:
@DerekCaelin This fix is now in review, will notify you when it can be re-tested.

DerekCaelin:
@tm-ruxandra I assume this is testable in 221?

tm-ruxandra:
@DerekCaelin yes

DerekCaelin:
Beautiful.

PR #2572, opened 2024-12-02
Feat: sync conflict notification

tm-ruxandra:
## Description

Add a new dialog which implements the sync error notification requirements. The dialog is exposed to the application through a new "sync notification context", which the missing site pull handler now uses instead of the generic Android toast. This context can also be reused for handling push errors.

### Checklist

### Related Issues
#2324 

### Verification steps

Trigger a sync error condition (for example, navigate to a site and then delete that site). Trigger a pull and observe the error dialog.

### DIFF
--- a/dev-client/src/components/NativeBaseAdapters.tsx
+++ b/dev-client/src/components/NativeBaseAdapters.tsx
@@ -109,18 +109,28 @@ type HeadingProps = NativeBaseTextProps &
     variant?: keyof (typeof theme.components)['Heading']['variants'];
   };
 
-export const Heading = (props: React.PropsWithChildren<HeadingProps>) => (
-  <RN.Text
-    {...convertNBStyles(
-      {
-        variant: 'h6',
-        color: 'text.primary',
-        ...props,
-      },
-      'Heading',
-    )}
-  />
-);
+export const Heading = (props: React.PropsWithChildren<HeadingProps>) => {
+  const hasTextAncestor = useContext(TextAncestorContext);
+  return (
+    <TextAncestorContext.Provider value={true}>
+      <RN.Text
+        {...convertNBStyles(
+          {
+            ...(hasTextAncestor
+              ? {variant: 'h6'}
+              : {
+                  variant: 'h6',
+                  color: 'text.primary',
+                  ...props,
+                }),
+            ...props,
+          },
+          'Heading',
+        )}
+      />
+    </TextAncestorContext.Provider>
+  );
+};
 
 type BadgeProps = NativeBaseProps &
   RN.ViewProps & {
--- a/dev-client/src/components/buttons/DismissButton.tsx
+++ b/dev-client/src/components/buttons/DismissButton.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2023 Technology Matters
+ * Copyright © 2024 Technology Matters
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as published
--- a/dev-client/src/components/buttons/DismissButton.tsx
+++ b/dev-client/src/components/buttons/DismissButton.tsx
@@ -29,7 +29,7 @@ export type DismissButtonProps = {
 export const DismissButton = ({onPress}: DismissButtonProps) => {
   return (
     <Button
-      borderColor="alert.errorContent"
+      borderColor="error.content"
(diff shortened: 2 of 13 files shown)


dev-client/src/theme.ts
```diff
@@ -62,6 +62,10 @@ export const theme = extendTheme({
       main: '#ED6C02',
       content: '#663C00',
     },
+    alert: {
+      errorContent: '#5F2120',
+      errorFill: '#FDEDED',
+    },
```
paulschreiber, line 68: issue: this duplicates error.content and error.background.

knipec (commented):
I'm assuming I don't need to review this as it looks to be a subset of what's in https://github.com/techmatters/terraso-mobile-client/pull/2574 but let me know if I'm misunderstanding!

dev-client/src/translations/en.json
```diff
@@ -392,6 +392,7 @@
         "users": "users",
         "save": "Save",
         "save_fab": "Save",
+        "dismiss": "Dismiss",
```
knipec, line 395: This doesn't seem to correspond to an entry in the es.json or es.po -- is that an issue?

dev-client/src/components/buttons/DismissButton.tsx
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright © 2023 Technology Matters
```
knipec, line 2: 2024

dev-client/src/components/dialogs/ErrorDialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+import DismissButton from 'terraso-mobile-client/components/buttons/DismissButton';
+import {Icon} from 'terraso-mobile-client/components/icons/Icon';
+import {ExternalLink} from 'terraso-mobile-client/components/links/ExternalLink';
+import {ModalHandle} from 'terraso-mobile-client/components/modals/Modal';
+import {Text, View} from 'terraso-mobile-client/components/NativeBaseAdapters';
+import {convertColorProp} from 'terraso-mobile-client/components/util/nativeBaseAdapters';
+
+export type ErrorDialogProps = React.PropsWithChildren<{
+  headline?: React.ReactNode;
```
knipec, line 37: Is there a reason to use ReactNode here with TranslatedHeading as a prop to the ErrorDialog, as opposed to using a string here and TranslatedHeading as a component within the ErrorDialog?

dev-client/src/components/NativeBaseAdapters.tsx
knipec, line 1: Confirming: Is this just so that we don't override the color or other props when we put the under a component?

dev-client/src/components/buttons/DismissButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {Button} from 'native-base';
+
+import {TranslatedContent} from 'terraso-mobile-client/components/content/typography/TranslatedContent';
+import {convertColorProp} from 'terraso-mobile-client/components/util/nativeBaseAdapters';
+
+export default function DismissButton() {
+  return (
+    <Button
+      borderColor="alert.errorContent"
+      _pressed={{backgroundColor: 'transparent'}} // TODO: get actual color
```
knipec, line 29: Do we have the actual color now? Is that what `action.selected` in theme.ts is?

dev-client/src/translations/en.json
tm-ruxandra: Yes, let me update es.json

dev-client/src/components/buttons/DismissButton.tsx
tm-ruxandra: I confirmed with Courtney that the button should remain transparent when selected. I'll remove the TODO.

dev-client/src/components/NativeBaseAdapters.tsx
tm-ruxandra: Yes

dev-client/src/components/dialogs/ErrorDialog.tsx
tm-ruxandra: I thought it was easier to pass in a node rather than proliferate the "i18n key + optional bundle of values" parameters that the `typography` components currently take, which feels clunkier to me.

knipec (approved):

Issue #2374, opened 2024-10-23
Offline UI: Notes: Disable add site notes, disable manipulates


DerekCaelin:
While offline, change the Notes screen experience.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10643-15334&node-type=section&t=aNOkjDj8zfXau8az-0)
<img width="252" alt="image" src="https://github.com/user-attachments/assets/69e69426-9fd1-4458-b3f9-3f8d057dd4ad">

## Acceptance Criteria
- [x] add alert
- [x] remove "Add note" button
- [x] disable edit icon button

paulschreiber:
Please verify in 217 or later.

DerekCaelin:
this meets my expectations

CourtneyLee333:
Looks good to me.

PR #2573, opened 2024-12-02
fix: disable adding and editing of notes when offline

paulschreiber:
## Description
Disable adding and editing of notes when offline.

Note: this builds on https://github.com/techmatters/terraso-mobile-client/pull/2566 and should be rebased after `fix/soil-properties` is merged..

### Related Issues
Addresses #2374.


knipec (approved):
Also might just be worth noting to Derek and Courtney that the Edit button on the site notes won't be disabled, it just won't show up

PR #2574, opened 2024-12-02
feat: push error notification

### LINKED ISSUE #2293: [Offline] If a push has errors, do a pull and show dialog
### Acceptance Criteria
- If a push fully succeeds, no need to pull right after.
- If a push response has errors (like an entity was deleted or permissions were restricted), pull data from server. This will cause the user's incompatible data to be dropped and clobbered with server data. But at least the data in the app will be in a consistent state. And if the push attempt stays in a history log we think that's ok. We suspect this case would happen infrequently anyway.
- If this happens when app starts or when user comes back online, we will already always pull after a push (from [#2292](https://github.com/techmatters/terraso-mobile-client/issues/2292)), so no need to do it an extra time if the push has errors
- Do this for soil data pushes through the new mechanism

Also:
- Notify the user of data conflicts per #2324, or create a new task for this:
If the user client pushes data and
  - A site was deleted or removed from a project
  - The user was removed from a project
  - Depth requirements for a project changed
....then display the dialog
- Consider if any mutation's error response should trigger a pull, including things that don't go through the new mechanism (like site notes, project edits, etc). If not, what are the consequences? And do we care if this case shows an error message?

tm-ruxandra:
## Description

(Depends on https://github.com/techmatters/terraso-mobile-client/pull/2572)

Show a sync error notification when a push completes with sync errors.

### Checklist
- [x] New tests added

### Related Issues
Fixes #2293

### Verification steps

Push with an error condition (e.g., to a deleted site). Observe the error notification.

### DIFF
--- a/dev-client/src/store/sync/PushDispatcher.tsx
+++ b/dev-client/src/store/sync/PushDispatcher.tsx
@@ -93,7 +93,7 @@ export const PushDispatcher = () => {
 
     /* Cancel any pending retries when push input changes or component unmounts */
     return endRetry;
-  }, [needsPush, dispatchPush, beginRetry, endRetry, syncNotifications]);
+  }, [needsPush, dispatchPush, beginRetry, endRetry]);
 
   return <></>;
 };
--- a/dev-client/src/store/sync/PushDispatcher.tsx
+++ b/dev-client/src/store/sync/PushDispatcher.tsx
@@ -17,6 +17,7 @@
 
 import {useEffect} from 'react';
 
+import {useSyncNotificationContext} from 'terraso-mobile-client/context/SyncNotificationContext';
 import {
   useDebouncedIsOffline,
   useDebouncedUnsyncedSiteIds,
@@ -40,6 +41,9 @@ export const PushDispatcher = () => {
   /* Determine whether the user is logged in before doing anything. */
   const isLoggedIn = useIsLoggedIn();
 
+  /* Use notifications to show errors to the user */
+  const syncNotifications = useSyncNotificationContext();
+
   /* Debounce offline state so we know when it's safe to attempt a push. */
   const isOffline = useDebouncedIsOffline(PUSH_DEBOUNCE_MS);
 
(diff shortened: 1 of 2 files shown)


dev-client/src/components/buttons/DismissButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {Button} from 'native-base';
+
+import {TranslatedContent} from 'terraso-mobile-client/components/content/typography/TranslatedContent';
+import {convertColorProp} from 'terraso-mobile-client/components/util/nativeBaseAdapters';
+
+export default function DismissButton() {
+  return (
+    <Button
+      borderColor="alert.errorContent"
+      _pressed={{backgroundColor: 'transparent'}} // TODO: get actual color
```
knipec, line 29: Do we have the actual color now? Is that what `action.selected` in theme.ts is?

dev-client/src/store/sync/PushDispatcher.tsx
```diff
@@ -62,18 +66,21 @@ export const PushDispatcher = () => {
...
     if (needsPush) {
       dispatchPush()
-        /* If the initial push failed, begin a retry cycle */
         .then(result => {
           if (!result.payload || 'error' in result.payload) {
+            /* If the initial push failed, begin a retry cycle */
             beginRetry();
+          } else if (Object.keys(result.payload.errors).length > 0) {
+            /* If the push yielded sync errors, notify the user */
+            syncNotifications.showError();
```
knipec, line 75: Confirming: the difference here is that `payload.error` indicates a generic error communicating with the backend, but `payload.errors` indicate an error like "does not exist" or "not allowed" (as set by `mutationResponseToResults`), right?

knipec: If that's the case, if we receive sync errors in the response from a retry cycle, the error wouldn't show since this is not in the retry interval's `dispatchPush`, but it should, right?

dev-client/src/translations/en.json
```diff
@@ -392,6 +392,7 @@
         "users": "users",
         "save": "Save",
         "save_fab": "Save",
+        "dismiss": "Dismiss",
```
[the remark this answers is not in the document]
knipec: Migrating comments to #2572 and resolving here

dev-client/src/store/sync/PushDispatcher.tsx
tm-ruxandra: Discussed on a call -- now I understand the problem, will update the retry mechanism to include the error handler correctly.

PR #2575, opened 2024-12-02
Fix/signin offline

paulschreiber:
## Description
Disable signing in while offline.

Note: this builds on https://github.com/techmatters/terraso-mobile-client/pull/2566 and should be rebased after `fix/soil-properties` is merged..

### Related Issues
Addresses #2376.


dev-client/locales/po/en.po
```diff
@@ -1018,6 +1021,9 @@ msgstr "Next"
 msgid "general##proceed"
 msgstr "Proceed"
 
+msgid "general##offline_title"
```
knipec, line 1024: Question: Where did this come from? I don't see the corresponding string in en.json
Or should the changes to this file just be dropped per your slack message?

dev-client/src/screens/LocationScreens/components/soilId/messageBoxes/OfflineMessageBox.tsx
```diff
@@ -20,13 +20,17 @@ import {useTranslation} from 'react-i18next';
...
+type OfflineMessageBoxProps = {
+  message: string;
+};
+
+export const OfflineMessageBox = ({message}: OfflineMessageBoxProps) => {
   const {t} = useTranslation();
 
   return (
-    <AlertMessageBox title={t('site.soil_id.matches.offline_title')}>
+    <AlertMessageBox title={t('general.offline_title')}>
```
knipec, line 31: Issue: `general.offline_title` does not appear to be in `en.json` -- does this work?

knipec (commented):
Could you also quickly update the associated #2376 to mention that it should include this too? Since right now it doesn't mention it, but I imagine we want Derek or Courtney to verify it.

dev-client/__tests__/snapshot/__snapshots__/LoginScreen-test.tsx.snap
knipec, line 1: Were you planning on updating the test for this?

dev-client/src/screens/LocationScreens/components/soilInfo/PropertiesScoreDisplay.tsx
```diff
@@ -50,6 +52,9 @@ export function PropertiesScoreDisplay({
     () => rowsFromSoilIdData(match.soilInfo.soilData),
     [match.soilInfo.soilData],
   );
+
+  const isOffline = useIsOffline();
```
knipec, line 56: Thought: Hmm this can be considered later, but I wonder if we want to use `useDebouncedIsOffline` or something so that if someone's connection is spotty, the UI isn't flickering? (But I think it's not as easy as "just use the debounced hook", as the UI would then allow users to make a change that they shouldn't be able to make offline)

dev-client/src/screens/LocationScreens/components/soilId/messageBoxes/OfflineMessageBox.tsx
knipec: Ah, I see it's in PR #2573 -- I assume you'll figure out the right order to merge things for the strings to work out :)

knipec (approved):
Approving, given necessary string changes get merged in first, and responses to my comments. :) 
Lmk if you want me to re-review after you've rebased or anything though!

dev-client/__tests__/snapshot/__snapshots__/LoginScreen-test.tsx.snap
paulschreiber: How do I simulate isOffline being false?

knipec: Let me know if you'd like more detail than [what I'd mentioned here](https://tech-matters.slack.com/archives/C02K02L3Q48/p1733245216813129?thread_ts=1733239497.386329&cid=C02K02L3Q48)?

Issue #2516, opened 2024-11-22
Offline UI: Disable checkbox for selected soil when offline


knipec:
When offline:
- If there is a selected soil for a site, I can click on it. Scrolling down on the resulting screen, I see a checkbox for selecting the correct soil. The checkbox should be disabled, because setting it when offline is not yet supported by our local changes mechanism.
- If there is any other way to view soil match tiles while offline, make sure clicking on them brings up a screen with the checkbox disabled as well.

This should be the state after https://github.com/techmatters/terraso-mobile-client/issues/2506 but before https://github.com/techmatters/terraso-product/issues/1055

CourtneyLee333:
I don't know what release this will be in, but I can't test on 216 because Top Match list returns errors for every site when offline.

CourtneyLee333:
Looks good!

PR #2576, opened 2024-12-02
feat: disable soil id match selector when offline

tm-ruxandra:
## Description

Disable soil ID match selector when offline. Add a new `DisableableText` typography component to help implement this type of feature in the future.

### Checklist

### Related Issues
#2516

### Verification steps

Enter offline mode and navigate to a soil ID match for a site. Observe that the selection checkbox is still visible but it and its text appear disabled. Interacting with it does not change the user's selection.

### DIFF
--- a/dev-client/src/components/content/typography/DisableableText.tsx
+++ b/dev-client/src/components/content/typography/DisableableText.tsx
@@ -1,0 +1,29 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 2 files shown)


dev-client/src/components/content/typography/DisableableText.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+type DisableableTextProps = {
+  disabled?: boolean;
+} & React.PropsWithChildren;
+
+export const DisableableText = ({
+  disabled = false,
+  children,
+}: DisableableTextProps) => {
+  return <Text color={disabled ? 'text.disabled' : undefined}>{children}</Text>;
```
paulschreiber, line 28: suggestion: style disabled text in theme.ts instead of with an inline color here.

tm-ruxandra: I don't know if it's correct to make a full variant for disabled text; it's an attribute we're adding to other text, not a first-class type like a heading, paragraph etc. The notion of "disabled" text is fully encapsulated by this component and it already uses the theme for the color key itself.

paulschreiber: You don't need to make a variant for that. React Native already supports `_disabled` as a sub-style. We've done this in other places in theme.ts.

tm-ruxandra: The `disabled` property doesn't work for text the way it does for e.g. buttons; if a `_disabled` section is specified in the theme for text as a base style or in variants it doesn't get triggered by setting its `disabled` property. I really think that this is the simplest solution. I will also note that `theme.ts` as a whole is a NativeBase construct which we are trying to move away from.

knipec (approved):

Issue #2568, opened 2024-11-27
Public/Private radio buttons should be in row, not column layout


paulschreiber:
## Description
Public/Private radio buttons for sites should be in row, not column layout
```
( ) Public ( ) Private
````

PR #2577, opened 2024-12-02
fix: make public/private radio buttons horizontal

paulschreiber:
## Description
Make public/private radio buttons horizontal instead of vertically stacked.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2568


knipec:
Actually you may not have to do this much work -- looking at other usages of RadioBlock, it looks like there's a variant called `'oneLine'` -- see `lightingGroupProps` in PhotoConditions.tsx

Issue #2371, opened 2024-10-23
Offline UI: Disable site settings


DerekCaelin:
While offline, disable options on the site settings screen.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10604-16804&node-type=frame&t=aNOkjDj8zfXau8az-0)
<img width="323" alt="image" src="https://github.com/user-attachments/assets/13cca913-91f8-4584-9ac7-2bae3514440b">

paulschreiber:
Please verify in 219 or later.

DerekCaelin:
this meets my expectation

PR #2594, opened 2024-12-06
fix: disable sign out button action, site settings when offline

paulschreiber:
## Description
Disable sign out button action when offline.

### Related Issues
Addresses #2376, #2371.

### DIFF
--- a/dev-client/src/components/buttons/DeleteButton.tsx
+++ b/dev-client/src/components/buttons/DeleteButton.tsx
@@ -27,7 +27,6 @@ type Props = {
 
 export default function DeleteButton({label, ...props}: Props) {
   const {t} = useTranslation();
-  const color = props?.disabled ? 'action.disabled' : 'error.main';
 
   label = label ?? t('general.delete_fab');
 
@@ -36,10 +35,17 @@ export default function DeleteButton({label, ...props}: Props) {
       background="background.default"
       p={0}
       pt="10px"
-      startIcon={<Icon name="delete" color={color} size="md" mr={2} />}
+      variant="error"
+      startIcon={
+        <Icon
+          name="delete"
+          color={props?.disabled ? 'action.disabled' : 'error.main'}
+          size="md"
+          mr={2}
+        />
+      }
       isDisabled={props?.disabled ? props.disabled : undefined}
       _text={{
-        color: color,
         fontWeight: 500,
         fontSize: 'md',
         textTransform: 'uppercase',
--- a/dev-client/src/components/buttons/DeleteButton.tsx
+++ b/dev-client/src/components/buttons/DeleteButton.tsx
@@ -30,14 +30,16 @@ export default function DeleteButton({label, ...props}: Props) {
 
   label = label ?? t('general.delete_fab');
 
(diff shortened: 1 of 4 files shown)


dev-client/src/components/buttons/DeleteButton.tsx
```diff
@@ -36,6 +36,7 @@ export default function DeleteButton({label, ...props}: Props) {
       p={0}
       pt="10px"
       startIcon={<Icon name="delete" color="error.main" size="md" mr={2} />}
+      isDisabled={props?.disabled ? props.disabled : undefined}
```
knipec, line 39: Suggestion/question: This is equivalent to just `props?.disabled`, right?

dev-client/src/components/buttons/DeleteButton.tsx
knipec, line 1: Issue: It looks like the button is still red, should be grayed out
Screenshot 2024-12-06 at 3 40 03 PM

dev-client/src/screens/UserSettingsScreen/components/menu/SignOutItem.tsx
```diff
@@ -16,28 +16,38 @@
...
-}
+};
+
+export const SignOutItem = () => {
+  const isDisabled = useIsOffline();
+
+  return isDisabled ? (
+    <SignOutMenuItem disabled={isDisabled} />
+  ) : (
+    <SignOutModal trigger={onOpen => <SignOutMenuItem onPress={onOpen} />} />
```
knipec, line 51: Note: Hmm the modal/trigger hierarchy here makes this read particularly awkward. We may want to rethink our pattern here if we're going to have to end up doing this in a lot of different places.

paulschreiber: I couldn't find a way to disable the trigger itself. I tried disabling onPress, but that didn't work.

dev-client/src/components/buttons/DeleteButton.tsx
paulschreiber: That results in a TypeScript error:
```
Type 'boolean | null | undefined' is not assignable to type 'boolean | undefined'.
  Type 'null' is not assignable to type 'boolean | undefined'.ts(2322)
```

dev-client/src/components/buttons/DeleteButton.tsx
paulschreiber: Fixed. Had to manually add disabled styles, even with `isDisabled={true}`.

dev-client/src/screens/UserSettingsScreen/components/menu/SignOutItem.tsx
paulschreiber: @tm-ruxandra any suggestions?

tm-ruxandra: @paulschreiber There's probably a better solution here but it requires re-working how our "trigger" system works. This is fine for now.

tm-ruxandra (approved):

Issue #2227, opened 2024-10-02
First time Android user doesn't see their blue location marker on the map


DerekCaelin:
### App Version

194

### Platforms

- [ ] iOS
- [X] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

As a first-time android user who has provided permissions, I can see the area around me but not the blue marker signifying my location. 

Discovered after a similar issue was resolved: https://github.com/techmatters/terraso-mobile-client/issues/1780

### Steps to Reproduce

On Android:

1. Uninstall, reinstall and open app
2. Pass through interstitial
3. Sign in with any identity
4. Open home screen and accept permissions request
5. See issue 

### Expected behavior

There should be a blue marker signifying my phone's location

### Actual behavior

No marker is visible. 

### Additional context

Marker returns after user closes and reopens the app.

knipec:
Spent some time on this, but am dropping this for now to prioritize other things. Here's where I left off:

## What's going on?
- Not entirely sure, but I hypothesize it's something about the sequencing of permissions being granted versus rendering mapbox's UserLocation component within the CustomUserLocation component.
- Perhaps related to https://github.com/techmatters/terraso-mobile-client/issues/1893 
- Makes some sense that it only repros on Android, because I hear rnmapbox implementation is different for Android and iOS

## Potential approaches for the fix:
- **Approach 1**: Continue using mapbox's UserLocation component, and newly use `useUpdatedLocationPermissions()` to prevent rendering UserLocation when permissions aren't granted (to fix [#1893](https://github.com/techmatters/terraso-mobile-client/issues/1893)), and to re-render when permissions are updated
  - On the Android emulator, if you move the emulator's location in the "..." menu, the blue dot would appear at the emulator's location, but the map would still zoom to the default location. [[I wonder if we are not getting the location correctly for the mapSlice's userLocation?]]
  - [[TODO: Confirm behavior using APK]]
- **Approach 2**: Stop using mapbox's UserLocation component. Instead, use Mapbox.ShapeSource and a redux selector to get state.map.userLocation.coords.
  - Work so far in [fix/map-location](https://github.com/techmatters/terraso-mobile-client/compare/main...fix/map-location) branch 
  - This would briefly generate two markers while it was zooming (see image below)
- Regardless of approach:  
  - The userLocation didn't get cleared when permissions were revoked, so probably do that too. (Relevant for where we zoom to.)
[...]

PR #2596, opened 2024-12-06
fix: Don't show user location dot when permissions aren't granted + re-render when permissions change.

knipec:
### Description 
- I expect re-rendering when permissions change should allow the user to see the blue location marker on the map when they first use the app. 
- However, the issue only reproes on a physical Android device, and tbh I don't want to go through the hassle of making a tagged release just to test it for myself. I suspect this change is at least not bad, and I will test it in the next beta release unless someone thinks I shouldn't do this. 😅 
- Also, I think we shouldn't show the location marker when they have revoked location permissions. (Side note: The map will still zoom, as that's determined by the userLocation in redux, which does not get cleared when permissions are revoked).
- I think it's also not ideal that we don't just use the userLocation in redux to determine where to draw the location marker. But when I tried that, I was having issues where two dots were rendering and I didn't understand why, and this bug doesn't seem high enough priority to take more time on it.

Also remove an unused prop

### Related Issues
Hopefully addresses #2227
Also #1893

### DIFF
--- a/dev-client/src/screens/SitesScreen/components/CustomUserLocation.tsx
+++ b/dev-client/src/screens/SitesScreen/components/CustomUserLocation.tsx
@@ -14,7 +14,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
-import {CircleLayer, Location, UserLocation} from '@rnmapbox/maps';
+import {CircleLayer, UserLocation} from '@rnmapbox/maps';
 
 import {USER_DISPLACEMENT_MIN_DISTANCE_M} from 'terraso-mobile-client/constants';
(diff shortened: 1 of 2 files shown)


dev-client/src/screens/SitesScreen/components/CustomUserLocation.tsx
```diff
@@ -41,20 +42,20 @@ const layerStyles = {
...
   onUserLocationPress: (event?: GeoJSON.GeoJsonProperties) => void;
-  updateUserLocation?: (location: Location) => void;
 };
 
 export const CustomUserLocation = ({
   onUserLocationPress,
-  updateUserLocation,
 }: CustomUserLocationProps) => {
+  const {permissions} = useUpdatedForegroundPermissions();
+  console.log('Permissions: ', permissions?.granted);
```
paulschreiber, line 51:
```suggestion
```

PR #2635, opened 2024-12-13
feat: offline soil id errors

tm-ruxandra:
## Description

Flush any data-based matches which have error statuses when the app first enters online mode.

### Checklist
- [x] New tests added

### Related Issues
[#1179](https://github.com/techmatters/terraso-product/issues/1179)

### Verification steps

Simulate a soil ID error when loading data-based matches (a site at a coordinate with no soilweb data, for example). Restart the application and verify that it attempts to reload the data when the user views that site again.

### DIFF
--- a/dev-client/src/context/SoilIdMatchContext.tsx
+++ b/dev-client/src/context/SoilIdMatchContext.tsx
@@ -43,6 +43,7 @@ import {
 import {
   fetchLocationBasedSoilMatches,
   fetchSiteDataBasedSoilMatches,
+  flushDataCacheErrors,
   flushLocationCache,
 } from 'terraso-mobile-client/model/soilIdMatch/soilIdMatchSlice';
 import {useDispatch, useSelector} from 'terraso-mobile-client/store';
@@ -82,7 +83,7 @@ export const SoilIdMatchContextProvider = ({
   const activeSites = useValueSet<string>();
 
   /* One-time cache flush for when the user first comes online */
-  useLocationCacheFlushing();
+  useCacheFlushing();
 
   /* Dispatch fetches as needed for the active inputs */
   useCoordFetching(useCoordsToFetch(activeCoords.values));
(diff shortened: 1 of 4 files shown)


dev-client/src/context/SoilIdMatchContext.tsx
```diff
@@ -114,6 +115,7 @@ const useLocationCacheFlushing = () => {
   useEffect(() => {
     if (!isOffline && !isFlushed) {
       dispatch(flushLocationCache());
+      dispatch(flushDataCacheErrors());
```
knipec, line 118: Actually, should we be caching matches for errored locations at all? Our messaging to the user is "Try again later" if soil id returns an error -- so wouldn't this only allow them to try again if they happened to go offline and then back online?

knipec (approved):
Approving so that if you decide the comments require no action you can submit it on Monday -- otherwise I'll revisit Tuesday

dev-client/src/context/SoilIdMatchContext.tsx
tm-ruxandra: This executes if the user starts the app in online mode, not just if they go offline -> online. We do need to cache error entries since we don't want to be continuously reloading them.

dev-client/src/model/soilIdMatch/soilIdMatches.ts
```diff
@@ -89,3 +93,15 @@ export const dataEntryForMatches = (
...
 };
+
+export const flushErrorEntries = (
+  entries:
+    | Record<string, SoilIdLocationEntry>
+    | Record<string, SoilIdDataEntry>,
+) => {
+  for (const id of Object.keys(entries)) {
+    if (isErrorStatus(entries[id].status)) {
+      delete entries[id];
```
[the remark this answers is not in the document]
tm-ruxandra: Yes, because this method is invoked in the `deleteSiteMatches` action which operates on a draft of new state. It will never modify existing state. (Redux generally is structured to disallow unintended modification of state.)

Issue #2290, opened 2024-10-09
[Offline: Pull] Screens specify data requirements & what to do if missing


knipec:
### Context
Unlikely to happen, but don't want to crash if it does.

### Acceptance criteria:
- For each screen: 
  - It specifies data that it requires
  - It specifies what to do if that data was deleted (generally, navigate somewhere and show the message as specified in #2324)
  - If viewing it requires certain permissions, confirm something reasonable happens when permissions are revoked & user syncs
  - If subcomponents require certain permissions, confirm they update when permissions change & user syncs 
- Automated tests
- Perhaps clarify for QA what to validate
  
### Notes:
- Consider breaking out separate tasks for:
  - Things that require special effort 
  - Permissions work could also be broken out if this ends up being too unwieldy.
  - Error messages for "sorry, you can't be on this screen anymore": ([Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10659-6337&t=GvPFAvFZqGQN6KO8-1), [GitHub Issue #1064](https://github.com/techmatters/terraso-product/issues/1064)) 
  - Refactoring to make selectors for things like Site and SiteNote return types that clearly include undefined or null (and use this to drive where requirements logic needs to go)

### Examples: 
- EditSiteNoteScreen
  - If site deleted, navigate to SitesScreen of BottomTabsScreen
  - If note deleted, navigate to SiteNotesScreen tab of SiteTabsScreen
  - If permissions changed manager-->viewer, ideally return to SiteNotesScreen -- which would not show notes as editable. (But really just do something reasonable.)
  - If permissions revoked entirely (removed from project), should be equivalent to site deleted
- ProjectSettingsScreen
  - If project deleted, return to ProjectsScreen of BottomTabsScreen
  - If permissions changed manager-->viewer, navigate to the Inputs tab of that project (this already happens!)

knipec:
Blocked on design

knipec:
Issues noticed after first checkin: 
- Engineering: Sigh, it looks like when you manually delete a site, you get the "Project conflict" dialog. This is not desirable behavior.
- Design: If an unaffiliated site has a conflict, you'd still get the "project conflict" dialog. Currently I don't think this is an active issue because sites can't be shared so won't have a conflict (unless you're logged in as the same user on multiple devices and delete the site on one while looking at it on the other)

DerekCaelin:
[...]

PR #2636, opened 2024-12-13
feat: Make site/project/note data required for relevant screens

knipec:
## Description
- Relevant screens specify what data is required to show the screen, and what to do if the data is missing
- Some minor refactoring

**NOTES**:
- There will be another PR for screens that require depth dependent data (to support a "deleted depth interval" scenario) -- just figured I could split up the PRs
- It may be helpful to go through individual commits to review

### Related Issues
Implements part of #2290

### DIFF
--- a/dev-client/src/screens/LocationScreens/LocationSoilIdScreen.tsx
+++ b/dev-client/src/screens/LocationScreens/LocationSoilIdScreen.tsx
@@ -20,6 +20,8 @@ import {ScrollView} from 'react-native-gesture-handler';
 
 import {Coords} from 'terraso-client-shared/types';
 
+import {useHandleMissingSiteOrProject} from 'terraso-mobile-client/components/dataRequirements/handleMissingData';
+import {RestrictByRequirements} from 'terraso-mobile-client/components/dataRequirements/RestrictByRequirements';
 import {Box} from 'terraso-mobile-client/components/NativeBaseAdapters';
 import {SiteRoleContextProvider} from 'terraso-mobile-client/context/SiteRoleContext';
 import {AppBar} from 'terraso-mobile-client/navigation/components/AppBar';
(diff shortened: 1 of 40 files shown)


dev-client/src/screens/SoilScreen/ColorScreen/ColorScreen.tsx
```diff
@@ -97,57 +100,76 @@ export const ColorScreen = (props: SoilPitInputScreenProps) => {
     [data],
   );
 
+  const site = useSelector(selectSite(props.siteId));
+  const handleMissingSite = useHandleMissingSiteOrProject();
+  // TODO-cknipe: Require the depth interval to exist for the site/project
```
tm-ruxandra, line 105: Is this TODO still relevant?

tm-ruxandra: Answered: it will be resolved in a subsequent PR

dev-client/src/screens/LocationScreens/LocationSoilIdScreen.tsx
```diff
@@ -45,29 +47,42 @@ export const LocationSoilIdScreen = ({siteId, coords}: Props) => {
     isSite ? selectSite(siteId)(state) : undefined,
   );
 
+  const siteIsRequiredButMissing = isSite && !site;
+  const handleMissingSite = useHandleMissingSiteOrProject();
+  const requirements = [
```
tm-ruxandra, line 52: Let's add a comment here / find some other way to make this logic easier to understand.

dev-client/src/screens/SiteNotesScreen/ReadPinnedNoteScreen.tsx
```diff
@@ -37,33 +42,51 @@ type Props = {
...
+
   const project = useSelector(selectProject(projectId));
   const content = project?.siteInstructions;
 
   const handleClose = () => {
     navigation.pop();
   };
 
+  const handleMissingProject = useHandleMissingSiteOrProject();
+  const handleMissingPinnedNote = useCallback(() => {
```
tm-ruxandra, line 55: Suggestion: turn this into a reusable handler hook like `useHandleMissingSiteOrProject()`

dev-client/src/screens/ManageTeamMemberScreen.tsx
```diff
@@ -81,53 +86,69 @@ export const ManageTeamMemberScreen = ({
...
   }, [dispatch, projectId, userId, selectedRole, navigation]);
 
+  const handleMissingProject = useHandleMissingSiteOrProject();
+  const handleMissingUser = useCallback(() => {
+    navigation.pop();
+    if (isFlagEnabled('FF_offline')) {
+      syncNotifications.showError();
+    }
+  }, [navigation, syncNotifications]);
+  const requirements = [
```
tm-ruxandra, line 96: I think we should find a way to avoid recomputing a new value every render cycle for requirements, since the consequences for re-rendering the Restrict component + potentially re-evaluating its child function seem unclear to me.

tm-ruxandra (changes requested):
Overall looks good; let's resolve the questions about memoizing requirements values + extracting the pop + sync notification data handler.

dev-client/src/screens/ManageTeamMemberScreen.tsx
knipec: Thinking more about it, I actually don't think this is really an issue currently. If the Screen is re-rendering, the `ScreenDataRequirements` component will also re-render (unless we memoize `ScreenDataRequirements` component, but I don't think that's necessary (and side note: empirically it looks like if a memoized component has children it'll still re-render when the parent does, maybe unless its children are also memoized? Not sure)) 

But the solution we talked through works and I don't think it's any harm, just adds complexity is probably not needed currently.

tm-ruxandra: Fine by me -- entirely possible I was being over-vigilant here. I'll let you make the call on whether to keep this logic before merging.

Issue #2367, opened 2024-10-23
Offline UI: Disable search bar


DerekCaelin:
Disable search bar while the user is offline.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=9314-5280&node-type=frame&t=1cPOpawF7xV05phQ-0)
<img width="247" alt="image" src="https://github.com/user-attachments/assets/b771af6c-0d60-4333-8e9b-b13348fa1907">

## Acceptance criteria
- [ ] When the user is offline disable the search bar

tm-ruxandra:
Per discussion with @DerekCaelin and @CourtneyLee333 , we are revising the requirements to leave the search bar enabled and only allow coordinate-based searches while offline. The autocomplete suggestions bar will be replaced with a text notification that search is not available offline.

CourtneyLee333:
It looks like the offline message color is using the disabled text color, which has legibility issues in this instance. Let's instead use the secondary text color, which is #001923 at 80% opacity. Otherwise, looks good to me!

tm-ruxandra:
@CourtneyLee333 Understood, I'll submit a change for that shortly.

DerekCaelin:
Looks good to me.

CourtneyLee333:
Closing, as I expect Ruxandra's fix will be fine.

CourtneyLee333:
@tm-ruxandra Just flagging that I don't see the text color update yet.

tm-ruxandra:
@CourtneyLee333 not sure how this got past me -- will have a fix up shortly.

PR #2638, opened 2024-12-13
feat: offline map search

tm-ruxandra:
## Description

Disable suggestion API calls when offline. Instead, display a message box styled similar to the Autocomplete component explaining to the user that only exact-coordinate matches are supported.

### Checklist
- [ ] 

### Related Issues
#2367

### Verification steps

Use search bar while offline and verify that expected message box is visible in place of autocomplete suggestions.

### DIFF
--- a/dev-client/src/screens/SitesScreen/components/search/MapSearch.tsx
+++ b/dev-client/src/screens/SitesScreen/components/search/MapSearch.tsx
@@ -37,7 +37,7 @@ import {useSitesScreenContext} from 'terraso-mobile-client/context/SitesScreenCo
 import {useUpdatedForegroundPermissions} from 'terraso-mobile-client/hooks/appPermissionsHooks';
 import {useIsOffline} from 'terraso-mobile-client/hooks/connectivityHooks';
 import {useMapSuggestions} from 'terraso-mobile-client/hooks/useMapSuggestions';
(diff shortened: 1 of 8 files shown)


dev-client/src/screens/SitesScreen/components/search/MapSearch.tsx
```diff
@@ -129,26 +107,29 @@ export default function MapSearch({zoomTo, zoomToUser, toggleMapLayer}: Props) {
...
+                  }}
+                  value={query}
+                  placeholder={t('search.placeholder')}
+                  style={{
+                    ...searchBarStyles.search,
+                    ...searchBarStyles.wideSearchOverride,
+                  }}
+                  inputStyle={searchBarStyles.input}
+                />
+                {isOffline && !hideResults ? <MapSearchOfflineBox /> : <></>}
```
knipec, line 131: Why does it matter what `hideResults` is / why not always show the search box when offline?

knipec (approved):

dev-client/src/screens/SitesScreen/components/search/MapSearch.tsx
tm-ruxandra: The `OfflineBox` is the alert that shows up underneath the search bar, and it should only show up when the search bar has focus. I can rename some things to clarify this relationship.

Issue #2415, opened 2024-11-07
Offline: Don't fetch Soil ID when offline


DerekCaelin:
Currently, the site popover and the soil ID tile displays "Error" when  viewed and the app is in Offline. The app should store soil ID information and present the most recent version of truth if it is not able to fetch new information.

#### Current & Expected temporary location Popover
<img width="150px" src="https://github.com/user-attachments/assets/d1a24731-3f57-4c10-9a61-3943a3c55174">

<img width="150px" alt="image" src="https://github.com/user-attachments/assets/71d3746e-ca72-43e9-9524-cbaa59d0cdd0">

#### Current & Expected Offline Soil ID Tile 
<img width="150px" src="https://github.com/user-attachments/assets/ea7b40f4-67a2-4cda-9f6e-476ef95767e2">

<img width="150px" alt="image" src="https://github.com/user-attachments/assets/87219b46-a264-4433-ab97-f01ac3ad2530">

### Acceptance Criteria
- [ ] If offline,
   - [x] Site popover shows last determined soil ID and Ecological Site ID
   - [x] Soil ID tile shows last determined soil ID and Ecological Site ID
   - [x] Soil ID page shows information last fetched from the soil ID algorithm

DerekCaelin:
I think this was assigned to me for a cleaning up... I should have asked an engineer to review. @tm-ruxandra it sounds like you are already doing work to satisfy this user story, is that correct?

tm-ruxandra:
@DerekCaelin correct, I was handling this as part of #2372

CourtneyLee333:
What I notice is that if the temp site popover remains open when I go offline:
- Soil ID Top Match and ES Top Match remain in the popover
- Elevation displays "Not Available Offline"

But if I close the temp popover and go offline (or go offline first and then close the popover), enter the exact same coordinates:
- All 3 fields on the popover display "Not Available Offline"

Soil ID works as expected for sites.

tm-ruxandra:
@CourtneyLee333 I can't reproduce that behavior -- for `37.43,-122` I'm able to repeatedly exit the popover and re-enter coordinates to see previously-loaded soil ID data. I'm curious which coordinates you're using to test?

DerekCaelin:
@tm-ruxandra I can reproduce courtney's experience.

1. Be online, search for coordinates `38,-122`
2. In temporary popover, see top match for soil ID, ESD, elevation
3. Go offline
[...]

PR #2662, opened 2024-12-16
fix: offline soil id displays

tm-ruxandra:
## Description

Correct various offline Soil ID displays to reflect intended behavior for soil ID data that was loaded before going offline. Clean up some nearby offline behavior while we're there.

### Checklist

### Related Issues
#2415

### Verification steps
 Confirm offline appearance for affected displays for sites that both do and do not have already-loaded soil ID data.

### DIFF
--- a/dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
+++ b/dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
@@ -69,12 +69,12 @@ export const SoilIdMatchesSection = ({
       <RestrictByConnectivity offline={true}>
         <OfflineMessageBox message={t('site.soil_id.matches.offline')} />
       </RestrictByConnectivity>
-      <MatchTilesOrMessage siteId={siteId} coords={coords} />
+      <MatchTiles siteId={siteId} coords={coords} />
     </ScreenContentSection>
   );
 };
 
-const MatchTilesOrMessage = ({siteId, coords}: SoilIdMatchesSectionProps) => {
+const MatchTiles = ({siteId, coords}: SoilIdMatchesSectionProps) => {
   const isOffline = useIsOffline();
   const soilIdData = useSoilIdData(coords, siteId);
   const status = soilIdData.status;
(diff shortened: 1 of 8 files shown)


dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -65,26 +66,23 @@ export const SoilIdMatchesSection = ({
           <TopSoilMatchesInfoContent isSite={isSite} />
         </InfoButton>
       </Row>
+      <RestrictByConnectivity offline={true}>
+        <OfflineMessageBox message={t('site.soil_id.matches.offline')} />
+      </RestrictByConnectivity>
       <MatchTilesOrMessage siteId={siteId} coords={coords} />
```
knipec, line 72: Could rename from `MatchTilesOrMessage` to just `MatchTiles` or `MatchResults` or `MatchesArea` or something, since the message is no longer part of this component

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -65,26 +66,23 @@ export const SoilIdMatchesSection = ({
...
   const status = soilIdData.status;
   const isSite = !!siteId;
 
-  if (isOffline) {
-    return <OfflineMessageBox message={t('site.soil_id.matches.offline')} />;
-  }
-
   switch (status) {
     case 'loading':
-      return <ActivityIndicator size="small" />;
```
knipec, line 87: Would we ever be in 'loading' state and offline? If so would we want to avoid showing the `ActivityIndicator`?

knipec (commented):
Thanks for fixing this behavior!

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
tm-ruxandra: Good idea.

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
tm-ruxandra: This is what the below does -- the indicator is hidden when offline.

knipec: Ohh yup, I'd misread it, thanks!

knipec (approved):

Issue #2366, opened 2024-10-23
Implement alert box component


CourtneyLee333:
Implement styles and behaviors of the alert box component per the design:
```[tasklist]
### Design:
- [ ] https://github.com/techmatters/terraso-product/issues/1117
```

paulschreiber:
This appears to be complete.

CourtneyLee333:
This does not pass my review. 

Feedback for both warning and error alerts:
- Text does not appear to be the color specified in the design. It should be #5F2120.
- Is the icon 18 x 18 px?

paulschreiber:
Color is fixed in 218. Icon will be 18x18 instead of 24x24 in 219.

CourtneyLee333:
@paulschreiber I don't see the text color fix.

paulschreiber:
Please verify in 219 or later.

CourtneyLee333:
@paulschreiber The text color looks good, but now the icon is super tiny and not legible. I'm not sure what's going on, as it certainly doesn't look 18px, but it was better before you downsized it. (My concern previously was that it looked a bit small, NOT that it needed to be smaller, which is why I asked what size it was.)

CourtneyLee333:
Looks good!

PR #2669, opened 2024-12-19
fix: make alert icon 24px

paulschreiber:
## Description
Make alert icon 24px.

Courtney approved this size.

Addresses #2366.

### DIFF
--- a/dev-client/src/theme.ts
+++ b/dev-client/src/theme.ts
@@ -300,6 +300,7 @@ export const theme = extendTheme({
     },
     Icon: {
       sizes: {
+        xs: 18,
         sm: 20,
         md: 24,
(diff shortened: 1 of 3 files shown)


dev-client/src/theme.ts
```diff
@@ -300,7 +300,6 @@ export const theme = extendTheme({
     },
     Icon: {
       sizes: {
-        xs: 18,
```
tm-ruxandra, line 303: We actually will need `xs` for the button component work, so this needs to be restored.

tm-ruxandra (changes requested):
See comment re: the xs icon size.

tm-ruxandra (approved):

PR #2705, opened 2024-12-26
fix: avoid displaying the sync error message when user deletes a site

knipec:
### Related Issues
Addresses issue after first checkin for #2290


knipec:
A

PR #2741, opened 2024-12-30
feat: require depth intervals and permissions on relevant screens

knipec:
## Description
Screens that depend on the existence of a depth interval now require the depth interval. 
Screens that depend on a user having certain permissions to view, now require those permissions.
To support permissions in particular,`ScreenDataRequirements` changed from requiring the data's existence, to requiring the data to be truthy. 

### Related Issues
Part 2 of #2290 
(There will be another PR for deletion of site/project not triggering the sync error)

### DIFF
--- a/dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
+++ b/dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
@@ -32,15 +32,15 @@ jest.mock('terraso-mobile-client/navigation/hooks/useNavigation', () => {
   };
 });
 
-const mockedUserMayEditSiteNote = jest.spyOn(
+const mockedUserCanEditSiteNote = jest.spyOn(
   permissionHooks,
-  'useUserMayEditSiteNote',
+  'useUserCanEditSiteNote',
 );
-mockedUserMayEditSiteNote.mockReturnValue(true);
+mockedUserCanEditSiteNote.mockReturnValue(true);
 
 afterEach(() => {
   mockedNavigate.mockClear();
-  mockedUserMayEditSiteNote.mockReset();
+  mockedUserCanEditSiteNote.mockReset();
 });
 
 describe('EditSiteNoteScreen', () => {
@@ -90,7 +90,7 @@ describe('EditSiteNoteScreen', () => {
   });
 
   test('renders null if user does not have permissions to edit the note', () => {
-    mockedUserMayEditSiteNote.mockReturnValue(false);
+    mockedUserCanEditSiteNote.mockReturnValue(false);
 
     const screen = render(
       <EditSiteNoteScreen noteId="nonexistent-note" siteId="1" />,
--- a/dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
+++ b/dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
@@ -18,6 +18,7 @@
 import {testState} from '@testing/integration/data';
 import {render} from '@testing/integration/utils';
 
+import * as permissionHooks from 'terraso-mobile-client/hooks/permissionHooks';
 import {EditSiteNoteScreen} from 'terraso-mobile-client/screens/SiteNotesScreen/EditSiteNoteScreen';
 
 const mockedNavigate = jest.fn();
@@ -27,12 +28,19 @@ jest.mock('terraso-mobile-client/navigation/hooks/useNavigation', () => {
   );
   return {
     ...actualNav,
-    useNavigation: () => ({navigate: mockedNavigate}),
+    useNavigation: () => ({navigate: mockedNavigate, pop: mockedNavigate}),
   };
 });
 
+const mockedUserMayEditSiteNote = jest.spyOn(
+  permissionHooks,
+  'useUserMayEditSiteNote',
+);
+mockedUserMayEditSiteNote.mockReturnValue(true);
+
 afterEach(() => {
   mockedNavigate.mockClear();
+  mockedUserMayEditSiteNote.mockReset();
 });
 
 describe('EditSiteNoteScreen', () => {
@@ -64,6 +72,26 @@ describe('EditSiteNoteScreen', () => {
   });
 
   test('renders null if note missing', () => {
+    const screen = render(
+      <EditSiteNoteScreen noteId="nonexistent-note" siteId="1" />,
+      {
+        route: 'EDIT_SITE_NOTE',
+        initialState: testState,
+      },
+    );
+
+    expect(screen.queryByText('Site Note')).toBeNull();
+    expect(screen.queryByText('note 1 contents')).toBeNull();
+    expect(mockedNavigate).toHaveBeenCalledTimes(1);
+    expect(mockedNavigate).toHaveBeenCalledWith('SITE_TABS', {
+      initialTab: 'NOTES',
+      siteId: '1',
+    });
+  });
+
+  test('renders null if user does not have permissions to edit the note', () => {
+    mockedUserMayEditSiteNote.mockReturnValue(false);
+
     const screen = render(
       <EditSiteNoteScreen noteId="nonexistent-note" siteId="1" />,
       {
--- a/dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
+++ b/dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
@@ -28,7 +28,7 @@ jest.mock('terraso-mobile-client/navigation/hooks/useNavigation', () => {
   );
   return {
     ...actualNav,
-    useNavigation: () => ({navigate: mockedNavigate, pop: mockedNavigate}),
+    useNavigation: () => ({navigate: mockedNavigate}),
   };
 });
 
--- a/dev-client/__tests__/integration/ScreenDataRequirements.test.tsx
+++ b/dev-client/__tests__/integration/ScreenDataRequirements.test.tsx
@@ -100,7 +100,7 @@ describe('ScreenDataRequirements', () => {
     expect(thingsDone).toEqual('null');
   });
 
-  test('renders children and triggers no action when required data exists, even if it is a boolean equal to false', () => {
+  test('does not render children and triggers action when required data is falsy', () => {
     let thingsDone = '';
     const requirements = [
       {
@@ -115,7 +115,7 @@ describe('ScreenDataRequirements', () => {
       </ScreenDataRequirements>,
     );
 
-    expect(queryByText('Hello world')).toBeTruthy();
-    expect(thingsDone).toEqual('');
+    expect(queryByText('Hello world')).toBeNull();
+    expect(thingsDone).toEqual('false');
   });
 });
--- a/dev-client/__tests__/integration/ScreenDataRequirements.test.tsx
+++ b/dev-client/__tests__/integration/ScreenDataRequirements.test.tsx
@@ -27,12 +27,27 @@ describe('ScreenDataRequirements', () => {
     const requirements = [
       {
         data: {dummyData: 1},
-        doIfMissing: () => (thingsDone += 'object1'),
+        doIfMissing: () => (thingsDone += 'object'),
       },
       {
         // This is an object, so it exists despite having only undefined properties
         data: {dummyData: undefined},
-        doIfMissing: () => (thingsDone += 'object2'),
+        doIfMissing: () => (thingsDone += 'object with undefined'),
+      },
+      {
+        // Some falsy things are still considered valid
+        data: 0,
+        doIfMissing: () => (thingsDone += '0'),
+      },
+      {
+        // Some falsy things are still considered valid
+        data: '',
+        doIfMissing: () => (thingsDone += 'empty string'),
+      },
+      {
+        // Some falsy things are still considered valid
+        data: NaN,
+        doIfMissing: () => (thingsDone += 'NaN'),
       },
     ];
 
--- a/dev-client/src/components/dataRequirements/ScreenDataRequirements.tsx
+++ b/dev-client/src/components/dataRequirements/ScreenDataRequirements.tsx
@@ -23,7 +23,7 @@ type Requirement = {
 };
 
 const dataExists = (data: object | undefined | null) => {
-  return data !== null && data !== undefined;
+  return !!data;
 };
 // First item should be the entity with the largest scope
 // Example: if EditSiteNoteScreen is missing the site and the site note,
@@ -50,9 +50,9 @@ type Props = {
 };
 
 /*
- * This is intended to wrap components (mostly screens as of 2024-12) so they only render
- * if required data is truthy. This prevents screens from breaking if, for example, a pull
- * happens that deletes data that is required to view the screen.
+ * This is intended to wrap Screen components so they only render if required data is truthy, and do
+ * an action if not. This prevents screens from breaking if, for example, a pull happens that
+ * deletes data that is required to view the screen.
  */
 export const ScreenDataRequirements = ({requirements, children}: Props) => {
   const requiredDataExists = useRequiredData(requirements);
--- a/dev-client/src/components/dataRequirements/ScreenDataRequirements.tsx
+++ b/dev-client/src/components/dataRequirements/ScreenDataRequirements.tsx
@@ -22,8 +22,8 @@ type Requirement = {
   doIfMissing?: () => void;
 };
 
-const dataExists = (data: object | undefined | null) => {
-  return Boolean(data);
+const dataExists = (data: object | undefined | null | boolean) => {
+  return data !== undefined && data !== null && data !== false;
 };
 // First item should be the entity with the largest scope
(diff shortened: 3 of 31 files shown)


dev-client/src/components/dataRequirements/ScreenDataRequirements.tsx
```diff
@@ -23,7 +23,7 @@ type Requirement = {
 };
 
 const dataExists = (data: object | undefined | null) => {
-  return data !== null && data !== undefined;
+  return !!data;
```
paulschreiber, line 26:
```suggestion
  return Boolean(data);
```

dev-client/src/screens/SoilScreen/ColorScreen/ColorGuideScreen.tsx
```diff
@@ -144,21 +146,33 @@ export const ColorGuideScreen = (props: SoilPitInputScreenProps) => {
...
-      <ScrollView>
-        <Column backgroundColor="grey.300" space="md">
-          {stepContent.map((content, index) => (
-            <Column backgroundColor="primary.contrast" key={index} padding="md">
-              <Text variant="body1-strong">
-                {t(`soil.color.guide.step${index + 1}.title`)}
-              </Text>
-              {content}
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 156: question: why is there an anonymous function here?

dev-client/src/screens/SoilScreen/ConductivityScreen.tsx
```diff
@@ -14,16 +14,27 @@
...
+    props.siteId,
+    props.depthInterval.depthInterval,
+  );
+
   return (
-    <SoilPitInputScreenScaffold {...props}>
-      <Text>Unimplemented Conductivity Screen</Text>
-    </SoilPitInputScreenScaffold>
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 33: question: (same) why is there an anonymous function here?

dev-client/src/screens/ColorAnalysisScreen/ColorAnalysisScreen.tsx
```diff
@@ -46,13 +48,22 @@ export const ColorAnalysisScreen = ({photo, pitProps}: ColorAnalysisProps) => {
...
   return (
-    <ColorAnalysisContext.Provider value={contextValue}>
-      <Stack.Navigator
-        initialRouteName="COLOR_ANALYSIS_HOME"
-        screenOptions={DEFAULT_STACK_NAVIGATOR_OPTIONS}>
-        {screens}
-      </Stack.Navigator>
-    </ColorAnalysisContext.Provider>
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 58: question: (same) why is there an anonymous function here?

dev-client/src/screens/SoilScreen/PhScreen.tsx
```diff
@@ -14,16 +14,27 @@
...
+    props.siteId,
+    props.depthInterval.depthInterval,
+  );
+
   return (
-    <SoilPitInputScreenScaffold {...props}>
-      <Text>Unimplemented Ph Screen</Text>
-    </SoilPitInputScreenScaffold>
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 33: question: (same) why is there an anonymous function here?

dev-client/src/screens/SoilScreen/SARScreen.tsx
```diff
@@ -14,16 +14,27 @@
...
+    props.siteId,
+    props.depthInterval.depthInterval,
+  );
+
   return (
-    <SoilPitInputScreenScaffold {...props}>
-      <Text>Unimplemented SAR Screen</Text>
-    </SoilPitInputScreenScaffold>
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 33: question: (same) why is there an anonymous function here?

dev-client/src/screens/SoilScreen/SOCSOMScreen.tsx
```diff
@@ -14,16 +14,27 @@
...
+    props.siteId,
+    props.depthInterval.depthInterval,
+  );
+
   return (
-    <SoilPitInputScreenScaffold {...props}>
-      <Text>Unimplemented SOCSOM Screen</Text>
-    </SoilPitInputScreenScaffold>
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 33: question: (same) why is there an anonymous function here?

dev-client/src/screens/SoilScreen/StructureScreen.tsx
```diff
@@ -14,16 +14,27 @@
...
+    props.siteId,
+    props.depthInterval.depthInterval,
+  );
+
   return (
-    <SoilPitInputScreenScaffold {...props}>
-      <Text>Unimplemented Structure Screen</Text>
-    </SoilPitInputScreenScaffold>
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 33: question: (same) why is there an anonymous function here?

dev-client/src/store/selectors.ts
```diff
@@ -338,6 +338,24 @@ export const useSiteSoilIntervals = (siteId: string): AggregatedInterval[] => {
   );
 };
 
+// This confirms that a given DepthInterval is actually part of the site's
+// displayed depth intervals. (There may be a mismatch if, for example, a
+// depth interval got deleted in redux, but a sheet takes depth as a prop)
+export const useSiteSoilInterval = (
```
paulschreiber, line 344: issue: the function name is a bit confusing.

does this return a boolean? if so, could it be name `usesSiteSoilInterval`?

dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
```diff
@@ -93,190 +92,213 @@ export const TextureGuideScreen = (props?: SoilPitInputScreenProps) => {
...
+              {result !== undefined && (
+                <Column p="15px" bg="primary.contrast" alignItems="flex-start">
+                  <Text variant="body1-strong" textTransform="uppercase">
+                    {t('soil.texture.guide.result', {
+                      result: t(`soil.texture.class.${result}`),
+                    })}
+                  </Text>
+                  <Box height="10px" />
+
+                  <Button
```
paulschreiber, line 289: question: does this need to be updated for the button component work?

dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
```diff
@@ -93,190 +92,213 @@ export const TextureGuideScreen = (props?: SoilPitInputScreenProps) => {
...
-                YES: {text: t('general.yes')},
-                NO: {text: t('general.no')},
-              }}
-              groupProps={{name: 'ball', onChange: setBall, value: ball}}
-            />
-          </Column>
-          {ball === 'YES' && (
-            <>
+    <ScreenDataRequirements requirements={requirements}>
+      {() => (
```
paulschreiber, line 102: question: (same) why is there an anonymous function here?

dev-client/src/hooks/permissionHooks.ts
```diff
@@ -33,3 +42,38 @@ export const useProjectUserRolesFilter = (
...
+  const note = site?.notes[noteId];
+  const currentUser = useSelector(state => state.account.currentUser.data);
+  const currentUserIsAuthor = !!note && note.authorId === currentUser?.id;
+
+  return (
+    userRole &&
+    ((userRole.kind === 'site' && canEditSite(userRole)) ||
+      (userRole.kind === 'project' &&
+        canEditProject(userRole.role) &&
+        currentUserIsAuthor))
```
paulschreiber, line 77: question: what are the scenarios in which `currentUserIsAuthor` but the user can't edit the note? is it if they get removed from the project? does that mean the new project owner can never edit the note?

dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
```diff
@@ -93,190 +92,213 @@ export const TextureGuideScreen = (props?: SoilPitInputScreenProps) => {
...
-              </Text>
-              <Box height="10px" />
-              {onUseResult !== undefined && (
-                <Button
-                  leftIcon={<Icon name="check" />}
-                  _text={{textTransform: 'uppercase'}}
-                  onPress={onUseResult}>
-                  {t('soil.texture.guide.use_label')}
-                </Button>
+              {result !== undefined && (
```
paulschreiber, line 280: question: do we want to act on any other falsy values of result? or does TypeScript prevent them from occuring?

dev-client/src/screens/SoilScreen/ColorScreen/ColorGuideScreen.tsx
knipec: The `ScreenDataRequirements` mentions this in its Props definition -- it defers evaluation of the children's props, so that the children can take required data

dev-client/src/hooks/permissionHooks.ts
knipec: It can happen if the site author became a viewer on a project. 
And yes, if the owner is removed from the project, the new project owner cannot edit the note.

dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
knipec: Could you explain more here? I didn't change this code; it only appears changed because wrapping everything in the `ScreenDataRequirements` component indented everything and GitHub is bad at diffs 🙃

dev-client/src/store/selectors.ts
knipec: No, it doesn't return a boolean. It is a hook that, given a site and DepthInterval (which is just start & end depth values), returns the existing AggregatedInterval for that site at that depth, if it exists. 

An `AggregatedInterval` is primarily a `SoilDataDepthIntervalNode`, which stores settings about the interval like `carbonatesEnabled`, `phEnabled`, `label`, etc (but does not include the depth-dependent data).

I'm not sure how to improve the name; the naming around depth intervals is rather confusing, and I tried to name it consistently with the `useSiteSoilIntervals` hook above it. I'll update the comment though, and happy to brainstorm together if that might help.

dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
knipec: Done; had to do this to resolve merge conflicts

dev-client/src/screens/AddUserToProjectScreen/AddUserToProjectRoleScreen.tsx
```diff
@@ -73,10 +74,13 @@ export const AddUserToProjectRoleScreen = ({projectId, userId}: Props) => {
     navigation.dispatch(TabActions.jumpTo(TabRoutes.TEAM));
   }, [dispatch, projectId, user, selectedRole, navigation]);
 
+  const roleIsEditor = useRoleMayEditProject(projectId);
```
shrouxm, line 77: nit: we changed from `isEditor` to `canEdit` elsewhere, would be nice to be consistent here

dev-client/src/components/dataRequirements/handleMissingData.ts
```diff
@@ -44,3 +52,30 @@ export const usePopNavigationAndShowSyncError = () => {
...
+  const site = useSelector(selectSite(siteId));
+  const realDepthInterval = useSiteSoilInterval(siteId, depthIntervalSpec);
+  const handleMissingSite = useNavToBottomTabsAndShowSyncError();
+  const handleMissingDepth = useNavToSiteAndShowSyncError(siteId);
+  const requirements = useMemoizedRequirements([
+    {data: site, doIfMissing: handleMissingSite},
+    {data: realDepthInterval, doIfMissing: handleMissingDepth},
+  ]);
+  return requirements;
+};
```
shrouxm, line 81: nit: i'm not sure if this exactly fits in `handleMissingData.ts`. maybe there should be another file specifically for factoring out common requirement lists like this?

dev-client/src/screens/AddUserToProjectScreen/AddUserToProjectRoleScreen.tsx
shrouxm: sry just now realizing this in like 10 different files :see_no_evil:

shrouxm (approved):
LGTM! some naming/code organization nits, but nothing blocking

dev-client/__tests__/integration/ScreenDataRequirements.test.tsx
```diff
@@ -100,7 +100,7 @@ describe('ScreenDataRequirements', () => {
     expect(thingsDone).toEqual('null');
   });
 
-  test('renders children and triggers no action when required data exists, even if it is a boolean equal to false', () => {
+  test('does not render children and triggers action when required data is falsy', () => {
```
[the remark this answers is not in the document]
shrouxm: ok, with the full context of the PR, and seeing how all the screen requirements shtuff has shaken out in practice, a couple more thoughts:
- `doIfMissing` is optional on `Requirement` but i think we always use it, so might be better to make it required?
- `ScreenDataRequirements` has kinda now grown in scope to be more like `ScreenRequirements`
- what if instead of it being `Requirement = {data: any; doIfMissing: ...}`, it was `Requirement = {requirement: boolean; doIfMissing: ...}`, and we called it like `{requirement: dataExists(site), doIfMissing: ...}`?

other than maybe the `doIfMissing` optional one, i wouldn't advocate for making these changes on this PR because it's already quite big and has another PR depending on it. just noodling on some things that confused me a bit now that we can actually see what this looks like implemented across the codebase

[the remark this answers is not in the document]
knipec: Yeah, I was also trying to figure out how to rename/restructure for a future PR. Thoughts:
- I found I wanted to use a boolean as well as existence for requiring permissions. Maybe it should just consider the requirement unmet if the value is `null | undefined | false`?
- Yeah I'd been thinking of something like `requirement` and `doIfUnmet` but I like the `dataExists` suggestion
- Filing https://github.com/techmatters/terraso-mobile-client/issues/2825 for this

dev-client/src/screens/AddUserToProjectScreen/AddUserToProjectRoleScreen.tsx
knipec: Will rename from `roleIsEditor` to `userCanEditSite` or `userCanEditProject` -- lmk if you meant something else though!

dev-client/__tests__/integration/EditSiteNoteScreen.test.tsx
```diff
@@ -27,12 +28,19 @@ jest.mock('terraso-mobile-client/navigation/hooks/useNavigation', () => {
   );
   return {
     ...actualNav,
-    useNavigation: () => ({navigate: mockedNavigate}),
+    useNavigation: () => ({navigate: mockedNavigate, pop: mockedNavigate}),
```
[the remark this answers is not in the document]
knipec: Turns out nothing is expecting pop here, so I'll just remove it 😅

dev-client/src/screens/SoilScreen/ColorScreen/ColorGuideScreen.tsx
paulschreiber: My question is: why do we return a function that returns a component instead of returning a component?

dev-client/src/screens/SoilScreen/TextureGuideScreen.tsx
paulschreiber: i.e. can we say `{result && (` instead of `{result !== undefined && (`?

shrouxm (approved):

dev-client/src/screens/SoilScreen/ColorScreen/ColorGuideScreen.tsx
knipec: I understood the question, lmk if this makes sense and I'll make a comment in the code about it. The main points are:
- We want the children to be able to expect that the required entities exist.
- But, React seems to evaluate props of children earlier than we'd want here, before the moment that ScreenDataRequirement decides if the children should actually render or not. 

So we want to allow children components to do things like
```
    <ScreenDataRequirements requirements={requirements}>
        <Text>{site!.name}</Text>
    </ScreenDataRequirements>
```
but this would give an error at runtime if the site was deleted on the backend, because it seems to be trying to evaluate `site!.name` when site is actually null (even though `ScreenDataRequirements` will later decide not to render the Text child at all). If we use this "function as child" pattern, the evaluation of `site!.name` happens only once `ScreenDataRequirements` decides to render its children, and we do not get the error.

PR #2799, opened 2025-01-09
docs: offline primer

garobrik:
## Description
Adds a primer on our offline architecture to the documentation.


knipec (approved):
I feel like we can submit this even if it's not 100% filled in. I'll put it on my todo list to add my part 🙃

PR #2822, opened 2025-01-14
(old) fix: Don't show error when deleting a site or project

knipec:
## Note:
This branch is in some messed-up merge state, so I'm abandoning it and just cherry-picked the commits into https://github.com/techmatters/terraso-mobile-client/pull/2839 -- please see that PR instead of this.

### DIFF
--- a/dev-client/src/screens/ProjectViewScreen.tsx
+++ b/dev-client/src/screens/ProjectViewScreen.tsx
@@ -66,6 +66,8 @@ export const ProjectViewScreen = ({projectId}: Props) => {
     projectPurposelyDeleted ? navigation.pop() : popNavAndShowSyncError();
   }, [projectPurposelyDeleted, navigation, popNavAndShowSyncError]);
 
+  // Child tabs don't need to duplicate the requirements here as long as their updates
+  // cause a rerender here
   const requirements = useMemoizedRequirements([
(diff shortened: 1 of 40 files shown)


dev-client/src/screens/ProjectViewScreen.tsx
```diff
@@ -15,37 +15,76 @@
...
 import {ScreenScaffold} from 'terraso-mobile-client/screens/ScreenScaffold';
 import {useSelector} from 'terraso-mobile-client/store';
 
+type ProjectDeletedState = [boolean, Dispatch<SetStateAction<boolean>>];
+
+export const ProjectDeletedContext = createContext<ProjectDeletedState>([
+  false,
+  () => {},
+]);
+
```
shrouxm, line 46: nit: i think we only ever use the set state action, so we could simplify the project deleted state by leaving it out

shrouxm (approved):
LGTM! very satisfying how much simpler all the project screens got :drooling_face:

dev-client/src/screens/ProjectViewScreen.tsx
knipec: Ah, yes -- we use the `projectPurposelyDeleted` state in the ProjectViewScreen but the child component doesn't need it so it doesn't need to be in the context 👍

Issue #2637, opened 2024-12-13
App settings screen: add support link, ToS, Privacy Policy


DerekCaelin:
Implement updates to app setting screen. 

Dependent on:
- https://github.com/techmatters/terraso-product/issues/1182

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=11353-43201&t=6odhXJdV9YbHwU7o-0)
![image](https://github.com/user-attachments/assets/5cd923a0-3060-4367-b4fb-9e8a8689716b)

## Acceptance criteria
- [x] Add external links
   - [x] Support: https://landpks.terraso.org/help
   - [x] Privacy Policy: https://techmatters.org/privacy-policy/
   - [x] Terms of Service: https://terraso.org/terms-of-service/
- [ ] Change sign out link to mixed case

CourtneyLee333:
One small edit needed: "Sign Out" should be "Sign out" (lower case "o").

DerekCaelin:
This meets my expectation aside from the note Courtney made

DerekCaelin:
@paulschreiber The "O" in "Sign out" is still capitalized.

CourtneyLee333:
Just noticed the "A" in "Delete account" should also be lowercase.

CourtneyLee333:
Looks good to me!

Issue #2302, opened 2024-10-11
Launch milestone interstitial


DerekCaelin:
Update the interstitial with content to reflect features at launch

## References
Note: add Tos and PP language

## Acceptance Criteria
Update interstitial language:

```
**This version includes:**
- Soil identification for US locations
-  Slope and soil observations entry
- Site notes
- Project management
- Offline capabilities

**What’s next?  Future versions will include:**
- Data visualization enhancements
- Additional soil properties, such as soil structure and soil organic carbon
- Site photos

**Learn more and stay updated:**
[LandPKS from Terraso](landpks.terraso.org)
By using LandPKS you accept our [Terms of Service](https://terraso.org/terms-of-service/) and [Privacy Policy](https://techmatters.org/privacy-policy/).
```

paulschreiber:
Please verify in 227 or later.

DerekCaelin:
Looks great. Over to @CourtneyLee333

CourtneyLee333:
Looks good!

PR #2827, opened 2025-01-15
fix: update interstitial for 1.2

paulschreiber:
## Description
Update interstitial for 1.2
• new bullets
• refactor/use a constant for key / flag to show/hide window
• add ToS and PP

### Related Issues
Addresses #2302. Also fixes capitalization of sign out on user settings screen (#2637).

### DIFF
--- a/dev-client/src/constants.ts
+++ b/dev-client/src/constants.ts
@@ -41,3 +41,5 @@ export const GEOSPATIAL_CONTEXT_USER_DISTANCE_CACHE = 5;
 export const LOCALE = 'en-US';
 export const MAP_QUERY_MIN_LENGTH = 2;
 export const SEARCH_DEBOUNCE_TIME_MS = 500;
+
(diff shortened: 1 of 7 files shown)


shrouxm (approved):
The code looks good! Have you run it on an emulator and confirmed it looks accurate in practice?

paulschreiber:
@shrouxm Yes — tested on the iOS emulator.

dev-client/src/constants.ts
```diff
@@ -41,3 +41,5 @@ export const GEOSPATIAL_CONTEXT_USER_DISTANCE_CACHE = 5;
 export const LOCALE = 'en-US';
 export const MAP_QUERY_MIN_LENGTH = 2;
 export const SEARCH_DEBOUNCE_TIME_MS = 500;
+
+export const WELCOME_SCREEN_SEEN_KEY = 'welcomeScreenAlreadySeen00';
```
knipec, line 45: Suggestion: The idea here is that with each release where we update the WelcomeScreen content, we'd increment the number here, right? If so, could make a comment noting this.

PR #2829, opened 2025-01-15
chore: disable non-security dependency updates for now

shrouxm:
## Description
Disable non-security dependency updates for now.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,8 @@ updates:
     schedule:
       interval: 'weekly'
       day: 'sunday'
-    open-pull-requests-limit: 20
+    # The following line will still allow security updates.
+    open-pull-requests-limit: 0
     allow:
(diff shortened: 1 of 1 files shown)


knipec (approved):
I'd propose something like this, is that what you're thinking?

In main: 
- I feel kinda uneasy with turning version updates off on main, but I could be convinced if we schedule a task to turn it back on once we make a release branch
- Once we turn version updates back on, I'd be down for having dependabot make update PRs monthly or quarterly instead of weekly (looks like quarterly intervals [just got implemented](https://github.com/dependabot/dependabot-core/issues/7552), hehe). We can choose to not take the update if we don’t think we’d have time to do basic testing or risk is high for some reason (though I realize that decision point may be the challenge we're currently facing)

When we make a release branch:
- In that branch, turn version updates off like this PR does (but dependabot will still security update PRs)

PR #2831, opened 2025-01-16
chore: disable ios build cache

shrouxm:
## Description
Disables the iOS build cache, per/pending discussion in https://github.com/techmatters/terraso-product/issues/1211

### Checklist

### Related Issues
Closes https://github.com/techmatters/terraso-product/issues/1211

### DIFF
--- a/.github/workflows/build-shared-ios.yml
+++ b/.github/workflows/build-shared-ios.yml
@@ -74,15 +74,15 @@ jobs:
           cache: 'npm'
           cache-dependency-path: dev-client/package-lock.json
 
-      - name: Install and configure ccache
-        run: |
-          brew install ccache
-          ln -s $(which ccache) /usr/local/bin/gcc
-          ln -s $(which ccache) /usr/local/bin/g++
-          ln -s $(which ccache) /usr/local/bin/cc
-          ln -s $(which ccache) /usr/local/bin/c++
-          ln -s $(which ccache) /usr/local/bin/clang
-          ln -s $(which ccache) /usr/local/bin/clang++
+      # - name: Install and configure ccache
+      #   run: |
+      #     brew install ccache
+      #     ln -s $(which ccache) /usr/local/bin/gcc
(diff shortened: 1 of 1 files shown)


.github/workflows/build-shared-ios.yml
```diff
@@ -88,14 +88,14 @@ jobs:
         working-directory: dev-client
         run: npm ci
 
-      - name: Restore iOS Expo cache
-        uses: actions/cache@v4
-        with:
-          path: |
-            dev-client/ios
-            dev-client/node_modules
```
paulschreiber, line 96: issue: you only want to remove `dev-client/ios`. We need to keep node_modules cached -- prebuild makes changes to `node_modules` we need to preserve.

shrouxm: can you explain further? the build on this PR succeeded, and i'm not understanding why an `npm ci` followed by `npm run prebuild` would need anything to be in cache for a successful build

paulschreiber: I'm suggesting we cache the results of npm run prebuild to save us time. this is separate from caching the iOS build step.

note there's another iOS build step you may want to disable: the ccache step.

shrouxm: ah good point, i disabled the `ccache` as well. i'm suggesting we _not_ cache the results of `npm run prebuild` because it doesn't save enough time to justify the not infrequent build issues we experience

paulschreiber: What about caching `node_modules` to save time on `npm ci`?

shrouxm: as far as i understand, the `setup-node` action already does all of the caching that would be helpful for `npm ci`:
- `npm ci` blows away the `node_modules` folder
- so the only thing that would be helpful for caching would be to have the dependency blobs cached somewhere
- `setup-node` does this caching

paulschreiber: Got it.

PR #2839, opened 2025-01-22
fix: Don't show error when deleting a site or project

knipec:
## Note
See https://github.com/techmatters/terraso-mobile-client/pull/2822 for additional PR history. The original branch got into some messed up state, so I'm giving up and making a new branch for this PR 🙃 

## Description
The main changes are in the Settings screens and the ProjectViewScreen.

### Site deletion
- Before: when you manually delete a site, you get the "Project conflict" dialog
- Now: you just go back to the site list screen, and don't get a conflict dialog
- Add tests for SiteSettingsScreen missing its required site and deleting a site 

### Project deletion
- Before: Project deletion had a more subtle issue where because the navigation was happening before the project actually got deleted, a) if the deletion took a long time it could still show the deleted project on the screen it just navigated to, b) if the deletion failed it would have navigated you away and that's annoying, and c) I didn't investigate this fully, but at least with Sites there was a concurrency issue where navigating before the dispatch would sometimes prevent the site from getting deleted in redux, and you'd have to wait til the next pull for the site to get out of redux.
- Now: Navigation as a result of a missing project happens as a side-effect in the requirements system. Different things happen depending on if the project is missing due to the user "purposely" deleting it just then, or if it was due to a pull from the backend. 
- This more closely matches the changes for sites. Though it's more complicated because the ProjectSettingsScreen is a tab of the ProjectViewScreen. 
- I would like to add tests for Projects, but should move on from this task. I'll create a tech-debt task for the tests I'd like to write.
- Also deleted redundant requirements from the child tabs of ProjectViewScreen.

### Related Issues
Fixes an issue with a previous commit for #2290 

### Verification steps
- Delete a site, make sure we navigate to the right place and don't show the error
- Delete a project, same
- Make sure that if the site or project is deleted in the backend, we do show the error
- FYI I found bugs along the way deleting a project with a site, and deleting a project that I just added members to; I'll file them soon. They are due to existing buggy behavior, but exacerbated by the changes in this PR

### DIFF
--- a/dev-client/__tests__/integration/screens/SiteSettingsScreen.test.tsx
+++ b/dev-client/__tests__/integration/screens/SiteSettingsScreen.test.tsx
@@ -1,0 +1,112 @@
+/*
+ * Copyright © 2024 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {render} from '@testing/integration/utils';
+
+import {FeatureFlagName} from 'terraso-mobile-client/config/featureFlags';
+import {SyncNotificationContext} from 'terraso-mobile-client/context/SyncNotificationContext';
+import {SiteSettingsScreen} from 'terraso-mobile-client/screens/SiteSettingsScreen/SiteSettingsScreen';
+import {AppState as ReduxAppState} from 'terraso-mobile-client/store';
+
+// TODO-cknipe:
+// - Mock syncNotification show
+// - Turn the feature flag on
+
+// -------- Too hard ---------
+// - Set up redux store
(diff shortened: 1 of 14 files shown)


dev-client/__tests__/integration/screens/SiteSettingsScreen.test.tsx
```diff
@@ -0,0 +1,199 @@
+/*
+ * Copyright © 2024 Technology Matters
```
paulschreiber, line 2:
```suggestion
 * Copyright © 2025 Technology Matters
```

dev-client/src/screens/ProjectViewScreen/ProjectViewScreen.tsx
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright © 2023 Technology Matters
```
paulschreiber, line 2:
```suggestion
 * Copyright © 2025 Technology Matters
```

dev-client/src/components/dialogs/ErrorDialog.tsx
```diff
@@ -54,7 +54,7 @@ export const ErrorDialog = forwardRef<
 
   return (
     <Portal>
-      <Modal visible={isOpen} onDismiss={onClose}>
+      <Modal visible={isOpen} onDismiss={onClose} testID="error-dialog">
```
paulschreiber, line 57: question: what does `testID` do here? should it be in a regular/non-test component?

paulschreiber (approved):

dev-client/src/components/dialogs/ErrorDialog.tsx
knipec: It just allows you to refer to the component by ID in a test with `getByTestId`. I find this to be a reliable way to get a component in a test (rather than by text, or by role, etc.)

Issue #2826, opened 2025-01-15
Map search field won't stay in focus


CourtneyLee333:
The map search field will not stay in focus, meaning it is not usable. When I tap into the search field, the keyboard briefly opens and then closes again, and I am no longer in the search field. 

- [x] Android
- [x] iOS

https://github.com/user-attachments/assets/d95468a7-35d8-4903-8570-dbcdbf35e21f

shrouxm:
this starting happening when paul upgraded the autocomplete dependency from version 5.4.0 to 5.5.0 a week ago. other people report experiencing the same thing here: https://github.com/byteburgers/react-native-autocomplete-input/issues/338. we can work around by either:
- rewriting our code to accommodate the new behavior of the library (this will also lean to our code being cleaner IMO), which is like a 2 or a 3 effort
- downgrading the library for now, which is a 1 effort

DerekCaelin:
This meets my expectations - @CourtneyLee333's turn

CourtneyLee333:
This works for me on Android and iOS.

PR #2844, opened 2025-01-24
fix: map search

shrouxm:
## Description
Reorganizes the map search code so that we are memoizing the autocomplete props properly. It'll be easier to review if you do it one commit at a time, as there's a fix commit followed by a refactor commit.

I struggled to write a test for this:
- i got the following error just from trying to render the component in a test:
    ```
    Cannot load null or undefined font source: { "material-community":  }. 
    Expected asset of type `FontSource` for fontFamily of name: "material-community"
    ```
- even if i can get the component to render, i'm not sure how to test focus state :thinking: 

### Checklist
- [ ] New tests added

### Related Issues
Related to #2826 

### Verification steps
Map search should be focusable again.

### DIFF
--- a/dev-client/src/screens/SitesScreen/components/search/MapSearch.tsx
+++ b/dev-client/src/screens/SitesScreen/components/search/MapSearch.tsx
@@ -15,10 +15,19 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import {useCallback, useEffect, useState} from 'react';
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import {useTranslation} from 'react-i18next';
-import {Keyboard} from 'react-native';
-import Autocomplete from 'react-native-autocomplete-input';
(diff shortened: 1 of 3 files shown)


dev-client/src/screens/SitesScreen/components/search/MapSearch.tsx
```diff
@@ -90,52 +156,20 @@ export const MapSearch = ({zoomTo, zoomToUser, toggleMapLayer}: Props) => {
...
-                <MapSearchSuggestionBox
-                  name={item.name}
-                  address={item.place_formatted}
-                  mapboxId={item.mapbox_id}
-                  onPress={selectQuery}
-                />
-              ),
-            }}
-            inputContainerStyle={{borderWidth: 0}} // eslint-disable-line react-native/no-inline-styles
-            renderTextInput={() => (
```
knipec, line 109: Question: No reason that you can tell that this needed to be a function instead of just the components, right?

knipec (approved):
Sounds like the library bug got fixed, but this fixes our bug now, and once we update that dependency the behavior should still be fine. And nice re-organization! 🎉

PR #2845, opened 2025-01-24
refactor: change mmkv library

### LINKED ISSUE #2842: debt: need to change KV storage library
We are currently using https://github.com/ammarahm-ed/react-native-mmkv-storage. We've experienced a lack of responsiveness to bug reports with the maintainer:
- https://github.com/ammarahm-ed/react-native-mmkv-storage/issues/360
- https://github.com/ammarahm-ed/react-native-mmkv-storage/issues/369

and are now blocked from upgrading to expo 52 and react native 0.76 because of this.

I see two potential options for other libraries to use:
- https://github.com/mrousavy/react-native-mmkv
- https://docs.expo.dev/versions/latest/sdk/sqlite/ (using the kv-storage module)

Switching to the Expo SQLite library seems more desirable to me:
- it reduces the number of organizations we rely on for our dependencies
- we know that the Expo people can be reached via office hours and will take care of critical bugs promptly
- it has some bells and whistles (on-device database browser seems nice)
- it makes a SQLite database available in our app in case we ever want to use one
- it's based on SQLite, the most widely used database in the world

Switching to the other MMKV library would mainly be desirable because:
- it'd be easier
- it has better performance (switching to sqlite might make our KISS strategy of synchronously writing the whole redux store to disk on each redux update less do-able)

But we're less likely to encounter any incompatibilities switching to the other MMKV library. It's possible the only reason we didn't take this approach initially is because the expo integration didn't exist previously.

## Migration implications
This changeover could be done in two ways:
- phased transition, where we try to preserve the contents of users' local databases by having two libraries installed at once and writing migration code. this is more effort (probably makes this a 7+ instead of a 3)
- hard transition. this is easier, but has the following user-visible implications:
  - users will be logged out of the app
  - unsynced soil data stored on users' devices could be lost
  - users will be shown the interstitial again

shrouxm:
## Description
Changes the MMKV library. Validated that all looks good in a release build on iOS and Android.

### Checklist

### Related Issues
Fixes #2842 

### Verification steps


dev-client/src/persistence/kvStorage.ts
```diff
@@ -24,36 +30,61 @@ import {MMKVLoader} from 'react-native-mmkv-storage';
...
+  setString: (key: string, value: string) => mmkvStorage.set(key, value),
+  setNumber: (key: string, value: number) => mmkvStorage.set(key, value),
+  setObject: (key: string, value: object) =>
+    mmkvStorage.set(key, JSON.stringify(value)),
+  getBool: (key: string) => mmkvStorage.getBoolean(key),
+  getString: (key: string) => mmkvStorage.getString(key),
+  getNumber: (key: string) => mmkvStorage.getNumber(key),
+  getObject: <T extends object>(key: string) => {
+    const str = mmkvStorage.getString(key);
+    if (str === undefined) return str;
```
paulschreiber, line 69: question: why do we need setXXX/getXXX instead of the generic set/get the library provides?

paulschreiber (approved):

PR #2853, opened 2025-01-29
feat: Offline snackbar for projects

knipec:
## Description
- Display project list screen content, with an offline message box instead of the button to create a project.
- Snackbar saying "You are offline. Projects can't be edited" will show up when you're on a screen related to projects and go offline, or if you are already offline and navigate to that screen
- Snackbar will also show up if you press a button offline that results in a failed backend call 

Not adding tests due to launch deadline 😢 -- will think about it as a follow-on task

### Related Issues
Implements https://github.com/techmatters/terraso-product/issues/1178


shrouxm (approved):
all LGTM! just need to add the `TK`s to `en.json` (/ i don't remember what the conclusion was re: turnaround time for translations?)

PR #2855, opened 2025-01-30
fix: Only show dev-only SyncContent if FF_testing flag is on

knipec:
## Description
So we don't show this dev-team-only stuff when we turn on the Offline feature flag:
Screenshot 2025-01-29 at 4 44 05 PM

### Related Issues
N/A 

### Verification steps
When the Offline feature flag is ON, but the Testing feature flag is OFF, make sure we don't see the sync content at the top of the app

### DIFF
--- a/dev-client/src/components/SyncContent.tsx
+++ b/dev-client/src/components/SyncContent.tsx
@@ -32,9 +32,11 @@ export const SyncContent = () => {
   return (
     <>
       <RestrictByFlag flag="FF_offline">
-        <SyncButton />
-        <PullInfo />
-        <PushInfo />
+        <RestrictByFlag flag="FF_testing">
+          <SyncButton />
+          <PullInfo />
+          <PushInfo />
+        </RestrictByFlag>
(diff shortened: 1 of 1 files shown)


dev-client/src/components/SyncContent.tsx
```diff
@@ -32,9 +32,11 @@ export const SyncContent = () => {
   return (
     <>
       <RestrictByFlag flag="FF_offline">
-        <SyncButton />
-        <PullInfo />
-        <PushInfo />
+        <RestrictByFlag flag="FF_testing">
```
knipec, line 35: We could also add a check that the environment is not prod if we wanna be doubly safe 🤷

shrouxm (approved):

dev-client/src/components/SyncContent.tsx
shrouxm: i think this is ok, we should be manually QA'ing any builds that are going to prod anyway and it'll be prettttty obvious if those controls are visible

Issue #2848, opened 2025-01-28
"Error saving soil ID" snackbar should not appear when opening app offline


knipec:
### Version observed
v232 

### Steps to Reproduce
1. Make sure offline feature flag is on 🙃 
2. Close the app & go offline
3. Open the app
**Expected**: No error message
**Actual**: See "Error saving soil ID information" temporarily show up in a toast at the bottom

shrouxm:
i looked into this briefly, it seems like just a crufty artifact that we haven't revisited in a long time. as far as i can tell:
- the "Error saving soil ID information" message shows up whenever the `status` field of the `soilData` slice is set to `error`
- the only time we set that `status` field is to `error` is when `fetchSoilDataForUser` fails, which isn't even a "save" action
- if you open the app offline, we dispatch a `fetchSoilDataForUser` from the `SitesScreen`, which will always fail, so you will always see "Error saving soil ID information" which is very inaccurate

i suggest we just remove the `Toasts` component before launch. it isn't serving any useful function right now, it is tied to native base, and carissa has already started work on a better snack bar solution anyway.

knipec:
That sounds reasonable, thanks! 
Also oh! We should probably not dispatch a `fetchSoilDataForUser` when opening the app offline, as once my snackbar change is in, I suspect that will cause the snackbar to show up when the user opens the app offline

DerekCaelin:
Related to: https://github.com/techmatters/terraso-mobile-client/issues/1657

DerekCaelin:
@shrouxm should this be in QA?

shrouxm:
yes, sorry!

DerekCaelin:
Is this QA'able in 236?

knipec:
Yes, this should be in 236! And yep, the feature flag should be on. The "invisible" thing is that in a production build, you should not be able to see the toggle controls for the feature flags (whereas in staging and dev builds you can toggle the feature flag controls off/on from the User Settings screen). So in prod, the feature flags are just set to the static hard-coded value of "on" or "off".

PR #2856, opened 2025-01-30
fix: don't do `fetchSoilDataForUser` requests when offline and don't show inaccurate toast message when that request fails

shrouxm:
## Description
In draft as I need to do my own QA for it first. Removes the old `Toasts` component, and changes the `fetchSoilDataForUser` dispatch on the `SitesScreen` into a `setPullRequested` dispatch inside the `PullRequester`. This was the last place `fetchSoilDataForUser` was called outside of the `PullDispatcher`.

### Checklist

### Related Issues
Related to #2848 

### Verification steps
- App should no longer display "Error saving soil ID information" when user opens app offline.
- App should not behave strangely when user logs out.
- App should successfully sync new user data when user opens app online.
- App should successfully load user data when user logs into app.


knipec:
I expect this should also address https://github.com/techmatters/terraso-product/issues/1235

knipec (approved):
Discussed in person: 
- Seems fine to no longer do `dispatch(setSoilIdStatus('ready'))` that the Toasts component was doing, since it's only used to show a loading indicator, which only happens if the status is 'loading'.
- Could keep the "request a pull" logic in SitesScreen instead of moving it to PullDispatcher, but we think it should result in the expected behavior either way (to make sure we pull if the current user changes).

PR #2857, opened 2025-01-30
fix: Reorganize the Feature Flag Control Panel

knipec:
## Description
- Not urgent, but it was bugging me
- Move the feature flag control panel up and into a Scrollview so it doesn't get cut off on differently sized devices
- Follows the same pattern as the UI components section

Before:

After (on a different device, to be fair): 
Collapsed / Expanded
 / 

### Related Issues
N/A

### DIFF
--- a/dev-client/src/components/FeatureFlagControl.tsx
+++ b/dev-client/src/components/FeatureFlagControl.tsx
@@ -20,6 +20,7 @@ import {ScrollView} from 'react-native-gesture-handler';
 import {Divider, Switch} from 'react-native-paper';
 
 import {ContainedButton} from 'terraso-mobile-client/components/buttons/ContainedButton';
+import {ScreenContentSection} from 'terraso-mobile-client/components/content/ScreenContentSection';
 import {
   Heading,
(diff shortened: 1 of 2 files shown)


dev-client/src/components/FeatureFlagControl.tsx
```diff
@@ -32,17 +34,29 @@ import {
...
-        </>
+        <View>
+          <ContainedButton
+            label="Feature Flags: Show/Hide"
+            stretchToFit={true}
+            onPress={toggle}
+          />
+          {show && (
+            <>
+              <ScrollView>
```
paulschreiber, line 50: suggestion: add margins/padding to the sides of the feature flag section so it matches the rest of the screen.

knipec: Now looks like this: Screenshot 2025-01-30 at 10 04 21 AM

knipec:
I'll merge this after we cut the release for launch

paulschreiber (approved):

Issue #2858, opened 2025-01-30
Soil ID: New sites indefinitely show "Loading" until close and restart of app


DerekCaelin:
### App Version

234

### Platforms

- [ ] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

When I create a new site, the soil ID information does not display until I close and restart the app. 

### Steps to Reproduce

1. Create temporary pin
3. Create site
4. Open site dashboard
5. See issue

### Expected behavior

See soil ID loading

### Actual behavior

See "loading..." indefinitely in soil ID tile and soil ID screen.

<img src="https://github.com/user-attachments/assets/ff56b912-ea84-4ff1-8a3d-859fe01cbf87" width="200px">
<img src="https://github.com/user-attachments/assets/1dfc74bd-b989-46d8-833c-7b0cafb18e84" width="200px">

### Additional context

Upon quitting the app and returning Soil Id algorithm runs as expected and the same soils list and percentages appears instantly.

PR #2865, opened 2025-01-30
fix: properly load soil ID when creating new sites

shrouxm:
## Description
Yet another bug due to our house of cards of redux selector logic. Forgot to include the default soil data in a selector. This creates a bug specifically for newly created sites because when we pull, the backend supplies default soil data for all sites. So it's only when we've created a site but haven't pulled yet that we need to supply default data ourselves on the client. There's definitely a better way to architect this but it seems out of scope for this pre-launch fix, but I think I can write a localized test for it.

Not sure if there's a straightforward better way to architect this, I'm a bit scared of all the nested selectors :(

### Checklist

### Related Issues
Related to #2858

### Verification steps
Soil ID should now load for newly created sites.


knipec (approved):
Looks reasonable to me! Still happy to talk thru it though

PR #2901, opened 2025-02-11
fix: android build by running `npx expo install --fix`

shrouxm:
## Description
Updates our dependencies according to expo's preferences by running `npx expo install --fix`.

### Verification steps
Android build should succeed.


paulschreiber:
Expo's preferences are behind — it's even picking an old version of expo.

shrouxm:
@paulschreiber if it's picking old versions of non-expo packages, doesn't that mean expo isn't tested to work with the newer versions of those packages yet and we shouldn't be upgrading them yet?

if this fixes the android build, we can revert some of the other downgrades one by one to make sure everything is still working

shrouxm:
but i'm inclined to use expo's guidance on what the appropriate version of other dependencies should be

paulschreiber (approved):

Issue #1233, opened 2024-05-02
Horizontal Icon Button doesn't have consistent hover/tap styling behavior


tm-ruxandra:
### App Version

118

### Platforms

- [X] iOS
- [X] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

The "Horizontal Icon Button" is used in several locations. When the user "hovers" (taps but does not release) on the icon, it displays a background color that does not extend to the text. This causes a confusing user experience, because the text is also clickable.

![Screenshot_1714690457](https://github.com/techmatters/terraso-mobile-client/assets/161085305/d73a0884-6ff3-426f-a1c9-1f950a2b7a19)

### Steps to Reproduce

1) Navigate to an occurrence of this button (e.g.: the "delete" prompt for pinned notes, or the privacy links in a project's privacy info page).
2) Press down on the icon without releasing, and observe that the background color only extends to the icon.
3) Click on the text and observe that it is clickable despite the affordance implied by the icon's background color.

### Expected behavior

The clickable area is consistent with the area that receives a background color when the user presses on the button.

### Actual behavior

The background area does not reflect the actual clickable area.

knipec:
Going thru bugs and wondering, do you consider this part of the components library work? @tm-ruxandra @CourtneyLee333

tm-ruxandra:
@knipec I think this ticket will be obviated once we have finalized designs for the button components.

knipec:
Closing, as Ruxandra and Courtney will be addressing this with the components library work

PR #2915, opened 2025-02-25
test: add some tests for offline snackbar

knipec:
## Description
Write tests for when the OfflineSnackbar component appears and disappears.

Also, allow files in `__tests__` and `__mocks__` to be affected by prettier

Thanks to @wseveraldojunior for the tips for getting this to work!

### Related Issues
Addresses [#1233](https://github.com/techmatters/terraso-product/issues/1233)

### DIFF
--- a/dev-client/__tests__/integration/OfflineSnackbar.test.tsx
+++ b/dev-client/__tests__/integration/OfflineSnackbar.test.tsx
@@ -188,12 +188,14 @@ describe('Offline snackbar', () => {
       </PaperProvider>,
     );
 
+    // FYI: Because the snackbar has an animation transition before hiding and returning null,
+    // Need to run the jest timers, and  wrap it in act() to make sureresulting state updates 
+    // are fully applied
     act(() => {
       jest.runAllTimers();
     })
 
-    const snackbar = screen.queryByTestId(snackbarTestId);
-    expect(snackbar).not.toBeOnTheScreen();
+    expect(screen.queryByTestId(snackbarTestId)).not.toBeOnTheScreen();
   });
 
 
--- a/dev-client/__tests__/integration/OfflineSnackbar.test.tsx
+++ b/dev-client/__tests__/integration/OfflineSnackbar.test.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Technology Matters
+ * Copyright © 2025 Technology Matters
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as published
@@ -18,16 +18,22 @@
 import 'terraso-mobile-client/translations';
 import 'terraso-mobile-client/config';
 
-import {PaperProvider, Portal} from 'react-native-paper';
+import { Button } from 'react-native';
+import { PaperProvider, Portal } from 'react-native-paper';
 
-import {testState} from '@testing/integration/data';
-import {render} from '@testing/integration/utils';
+import { fireEvent, waitForElementToBeRemoved } from '@testing-library/react-native';
+import { testState } from '@testing/integration/data';
+import { render } from '@testing/integration/utils';
 
-import {OfflineSnackbar} from 'terraso-mobile-client/components/messages/OfflineSnackbar';
-import {FeatureFlagName} from 'terraso-mobile-client/config/featureFlags';
+import * as projectService from 'terraso-client-shared/project/projectService';
+
+import { OfflineSnackbar } from 'terraso-mobile-client/components/messages/OfflineSnackbar';
+import { FeatureFlagName } from 'terraso-mobile-client/config/featureFlags';
 import * as connectivityHooks from 'terraso-mobile-client/hooks/connectivityHooks';
-import {ProjectListScreen} from 'terraso-mobile-client/screens/ProjectListScreen/ProjectListScreen';
-import {ProjectViewScreen} from 'terraso-mobile-client/screens/ProjectViewScreen/ProjectViewScreen';
+import { deleteProject } from 'terraso-mobile-client/model/project/projectSlice';
+import { ProjectListScreen } from 'terraso-mobile-client/screens/ProjectListScreen/ProjectListScreen';
+import { ProjectViewScreen } from 'terraso-mobile-client/screens/ProjectViewScreen/ProjectViewScreen';
+import { AppState, useDispatch } from 'terraso-mobile-client/store';
 
 jest.mock('terraso-mobile-client/config/featureFlags', () => {
   const actual = jest.requireActual(
@@ -55,12 +61,37 @@ jest.mock('terraso-client-shared/terrasoApi/api', () => {
   };
 });
 
+jest.mock('terraso-client-shared/project/projectService', () => {
+  const actual = jest.requireActual(
+    'terraso-client-shared/project/projectService',
+  );
+  return {
+    ...actual,
+    deleteProject: jest.fn(),
+  };
+});
+
 describe('Offline snackbar', () => {
   const useIsOfflineMock = jest.mocked(connectivityHooks.useIsOffline);
+
+  // TODO-cknipe: Uhh maybe remove this? Or look at the debugger to find what to retun
+  // const requestGraphQLMock = jest.mocked(api.requestGraphQL);
+  // type GraphQLResponseType = ReturnType<typeof api.requestGraphQL>;
+  // const graphQLError = Promise.reject([
+  //   'terraso_api.error_unexpected',
+  // ]) as GraphQLResponseType;
+  // const graphQLSuccess = Promise.resolve() as GraphQLResponseType;
+
+  const deleteProjectMock = jest.mocked(projectService.deleteProject);
+
+  // TODO-cknipe: Use only one of these
   const snackbarText = "You are offline. Projects can't be edited.";
+  const snackbarTestId = 'offline-snackbar';
 
   beforeEach(() => {
     useIsOfflineMock.mockReset().mockReturnValue(false);
+    // requestGraphQLMock.mockReset();
+    deleteProjectMock.mockReset();
   });
 
   test('appears when offline and first rendering project screen', () => {
@@ -79,7 +110,7 @@ describe('Offline snackbar', () => {
       },
     );
 
-    expect(screen.getByText(snackbarText)).toBeTruthy();
+    expect(screen.queryByTestId(snackbarTestId)).toBeOnTheScreen();
   });
 
   test('does not appear when online and first rendering project screen', () => {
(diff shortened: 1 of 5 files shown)


dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+    // This fails, and doesn't seem like an ideal way to test this anyway
+    // expect(screen.queryByTestId(snackbarTestId).props.visible).toBe(true);
+
+    console.log('Starting 2nd render');
+    useIsOfflineMock.mockReturnValue(false);
+    screen.rerender(
+      <PaperProvider>
+        <Portal.Host>
+          <OfflineSnackbar />
```
wseveraldojunior, line 207: It seems that the `SnackBar` component from react-native-paper uses transition animations. Once these transitions complete, an internal state `isHidden` is set to true.

wseveraldojunior: This means you need to some memoments before checking if the snackbar is still on the screen. Otherwise, the check may return true while the animation is still in progress.

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+      })
+      .catch(() => {
+        console.log('Nope'); // This happens
+      });
+
+    // These all fail
+    expect(screen.queryByText(snackbarText)).toBeFalsy();
+    expect(screen.queryByText(snackbarText)).not.toBeOnTheScreen();
+    expect(screen.queryByTestId(snackbarTestId)).toBeFalsy();
+    expect(screen.queryByTestId(snackbarTestId)).not.toBeOnTheScreen();
```
wseveraldojunior, line 225: So, in order to make it work you should do something like this
```suggestion
  // Wait for the animation to complete
  act(() => {
    jest.advanceTimersByTime(1500);
  });

  // These assertions should now pass
  expect(screen.queryByText(snackbarText)).toBeFalsy();
  expect(screen.queryByText(snackbarText)).not.toBeOnTheScreen();
  expect(screen.queryByTestId(snackbarTestId)).toBeFalsy();
  expect(screen.queryByTestId(snackbarTestId)).not.toBeOnTheScreen();
```

wseveraldojunior (commented):
Waiting for animation before asserting.

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+    //
+    // Output from `npm run test OfflineSnackbar`: 
+    //   Test suite failed to run
+    //   TypeError: Cannot set properties of undefined (setting 'message')
+    //     at node_modules/jest-circus/build/formatNodeAssertErrors.js:47:25
+    //         at Array.map (<anonymous>)
+    //   Test Suites: 1 failed, 1 total
+    //   Tests:       0 total
+    // I suspect this is the erroring line in the test library: 
+    // https://github.com/jestjs/jest/blob/main/packages/jest-circus/src/formatNodeAssertErrors.ts#L58
```
wseveraldojunior, line 277: What is causing your entire test suite to fail (for the same reason as in your Question 0) is not this line:
`expect(undefined).toBeTruthy();`

This should make a single test fail for sure. 

The real issue causing the whole suite to fail is this mocked return without proper async handling:

`deleteProjectMock.mockReturnValue(Promise.reject(['rejected']));`

(try to delete this line and call `expect(undefined).toBeTruthy();` or `await act(...) )`

So, how to have the proper async handling here?

Use `await act(async () => {})`. By using it, you ensure that Jest waits for all asynchronous operations inside the callback to complete before continuing with the test.

(using async here is important because any Promises created inside, like the rejected one from delete projects will be captured within the scope of the async function)

You could do something like this: (next comment)

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+    expect(screen.queryByTestId(snackbarTestId)).not.toBeOnTheScreen();
+
+
+    // Fire the test button event to make a server request. The request is mocked to fail, 
+    // which should add a message to the notificationsSlice and trigger the snackbar
+    fireEvent.press(screen.queryByTestId('test-delete-project-btn'));
+    await act(() => {
+      jest.runAllTimers();
+    });
```
wseveraldojunior, line 286:
```suggestion
await act(async () => {
   fireEvent.press(screen.queryByTestId('test-delete-project-btn'));
});
```

After that, you don’t need jest.runAllTimers() to assert whether the snackbar is on the screen, since it is first mounted (i.e., it appears in the DOM) before the fade-in animation starts. This is different from the fade-out animation, where the component first fades out and only after some time returns null.

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+    console.log("4")
+
+    // Question 2: When should an act call be awaited or not?
```
wseveraldojunior, line 289: `await act(async () => {...})` it when you need to properly handle an async call, otherwise just act will be prob fine.

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
knipec: Ah thank you, that's very helpful! It looks like if I use 
`deleteProjectMock.mockImplementation(async () => Promise.reject(['rejected']));`
it stops the issue of the whole test failing.

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
knipec, line 1: @wseveraldojunior Thank you for your help! (And apologies for the slow response, my work schedule is weird this month and next due to hopefully-temporary funding issues.) 

Another question if you have time: The test is working now, but I am still trying to understand when and why some of the fireEvent+act+runAllTimers do or don't work, so I understand better for the future. I updated the PR with examples of code that does/doesn't work in code comments -- if you are able to explain why the Options noted as "Bad" cause the test to fail, that would help me learn! Thanks again!

Also just as an FYI:
- It seems React [recommends](https://react.dev/reference/react/act) using await/async with act, and plans to deprecate the sync version, so I'll use that.
- It seems the whole suite was failing from the mock because I was using mockReturn instead of mockImplementation with an explicitly async function, so changed it to the latter.

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+    expect(screen.queryByTestId(snackbarTestId)).not.toBeOnTheScreen();
+
+    // Fire the test button event to make a server request. The request is mocked to fail,
+    // which should add a message to the notificationsSlice and trigger the snackbar.
+
+    // B: Option 1 - Bad, test fails with snackbar not on the screen
+    // fireEvent.press(screen.queryByTestId('test-delete-project-btn'));
+    // act(() => {
+    //   jest.runAllTimers();
+    // });
```
wseveraldojunior, line 316: Here the test fails because `deleteProject` is an async action triggered by the button press, and you need to properly handle the promise. While `fireEvent` is already wrapped by `act`, it doesn’t wait for async operations caused by after the button press (redux), only the event itself (button press).  That’s why you need to explicitly call `await act(async () => {}),` ensuring Jest waits for all async updates before moving on with the test.

In this case, you could use:

```
await act(async () => {
   fireEvent.press(screen.queryByTestId('test-delete-project-btn'));
});
```

_(Also, I don’t think _jest.runAllTimers();_ is needed here.)_

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+
+    // B: Option 3 - Good, test passes
+    await act(async () => {
+      fireEvent.press(screen.queryByTestId('test-delete-project-btn'));
+    });
+
+    // B: Option 4 - Bad, test fails with snackbar not on the screen
+    // fireEvent.press(screen.queryByTestId('test-delete-project-btn'));
```
wseveraldojunior, line 330: Same issue as above.

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
wseveraldojunior: Without act:
image

With act:
image

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
wseveraldojunior: Thanks for the info!. No need to apologize, I'm keeping an eye on my emails :)

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+    // fireEvent(snackbar, 'onDismiss');
+    // await act(async () => {
+    //   jest.runAllTimers();
+    // });
+
+    // A: Option 3 - Bad, test fails with snackbar still found on the screen
+    // await act(async () => {
+    //   fireEvent(snackbar, 'onDismiss');
+    //   jest.runAllTimers();
+    // });
```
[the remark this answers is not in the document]
knipec: Ah, thanks! So to confirm my understanding: the `fireEvent` here for dismissing the snackbar does not need to be explicitly wrapped in async act (but doesn't hurt), but the subsequent `fireEvent` for pressing the delete project button does. This difference is likely because the only async state updates that result from `onDismiss` are handled later by calling `runAllTimers`, whereas the delete project button press causes async state updates (due to the async action that updates redux state).

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
```diff
@@ -0,0 +1,350 @@
...
+        <Portal.Host>
+          <OfflineSnackbar />
+          <ProjectViewScreen projectId="1" />
+        </Portal.Host>
+      </PaperProvider>,
+      {
+        route: 'PROJECT_VIEW',
+        initialState: testState,
+      },
+    );
```
paulschreiber, line 110: question: since this is repeated, should it be defined outside the test?

paulschreiber (approved):

dev-client/__tests__/integration/OfflineSnackbar.test.tsx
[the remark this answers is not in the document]
wseveraldojunior: Yes, that's right!

PR #2939, opened 2025-05-06
build: bump terraso-client-shared and UUID

knipec:
## Description
- Update the commit hash for the client-shared dependency, after updating its commit hash for its backend dependency

### Related Issues
Required for https://github.com/techmatters/terraso-backend/issues/1618

### DIFF
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -62,7 +62,7 @@
         "react-native-svg": "^15.11.2",
         "react-native-tab-view": "^4.0.12",
         "reduce-reducers": "^1.0.4",
-        "terraso-client-shared": "github:techmatters/terraso-client-shared#9a50e858fd16e7025b742cc57cb6105537892940",
+        "terraso-client-shared": "github:techmatters/terraso-client-shared#9c6941f",
         "use-debounce": "^10.0.4",
         "uuid": "^10.0.0",
         "yup": "^1.6.1"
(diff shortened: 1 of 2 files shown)


dev-client/package-lock.json
```diff
@@ -24578,20 +24578,34 @@
...
-        "terraso-backend": "github:techmatters/terraso-backend#fac222b",
-        "uuid": "^10.0.0"
+        "terraso-backend": "github:techmatters/terraso-backend#43c94f4",
+        "uuid": "^11.1.0"
       },
       "engines": {
         "node": ">=18"
       }
     },
+    "node_modules/terraso-client-shared/node_modules/uuid": {
```
knipec, line 24596: This uuid section in `package-lock.json` appeared when I ran `npm install` -- I'm assuming it's fine, but let me know if you suspect anything weird.

dev-client/package-lock.json
```diff
@@ -24578,20 +24578,34 @@
...
+      "integrity": "sha512-hXwpBYIFH0igAid7nVHHAN5dt+XA7FPuVXhz+q/3utL83bbsBbhG//ai+uP04XZSRY7jYJiHjSi4kqTyJdvpAA==",
       "dependencies": {
         "@reduxjs/toolkit": "^1.9.7",
         "jwt-decode": "^4.0.0",
         "lodash": "^4.17.21",
         "react": "^18.3.1",
         "react-redux": "^8.1.3",
-        "terraso-backend": "github:techmatters/terraso-backend#fac222b",
-        "uuid": "^10.0.0"
+        "terraso-backend": "github:techmatters/terraso-backend#43c94f4",
```
knipec, line 24589: Is it expected that the terraso-backend entry (listed above on line 14577) still lists its commit hash as `fac222bd10eeac0176a2040e9b09449e84d0a693` whereas this reference to terraso-backend has this updated commit hash? Or do I need to do something to make them consistent?

paulschreiber (changes requested):
Hashes should match.

dev-client/package-lock.json
knipec: Ah, thank you -- I think I only ran with `install` but failed to run with `update` both in mobile-client and client-shared. Will run it in client-shared, then here.

knipec:
> @knipec note also we don't use `chore` as a branch prefix. and this would be more of a `build` than a chore.

Noted for next time 👍

paulschreiber (approved):

Issue #2940, opened 2025-05-06
Build is broken on main, preventing merging PRs


knipec:
### Description

Between Build #6026 and #6027 it appears the iOS build step consistently fails. See https://github.com/techmatters/terraso-mobile-client/actions/workflows/build.yml 

FYI I had to update my personal access token in Github to include the "workflow" scope, so that I could push this change. Maybe you do too?

### Steps to Reproduce

Issue a PR to kick off a Build workflow

### Expected behavior

From [last successful build output](https://github.com/techmatters/terraso-mobile-client/actions/runs/14230107821/job/39878738575): 
Seems to use Xcode 16.2
```
[21:40:08]: ▸ Command line invocation:
[21:40:08]: ▸     /Applications/Xcode_16.2.app/Contents/Developer/usr/bin/xcodebuild -resolvePackageDependencies -workspace dev-client/ios/LandPKSSoilID.xcworkspace -scheme LandPKSSoilID -configuration Release
```

### Actual behavior

From [first failing build output](https://github.com/techmatters/terraso-mobile-client/actions/runs/14573043310/job/40873630152):
Seems to use Xcode 16.3
```
[12:02:00]: Maybe the error shown is caused by using the wrong version of Xcode
[12:02:00]: Found multiple versions of Xcode in '/Applications/'
[12:02:00]: Make sure you selected the right version for your project
[12:02:00]: This build process was executed using '/Applications/Xcode_16.3.app'
[12:02:00]: If you want to update your Xcode path, either
[12:02:00]: 
[12:02:00]: - Specify the Xcode version in your Fastfile
[12:02:00]: ▸ xcversion(version: "8.1") # Selects Xcode 8.1.0
[12:02:00]: 
[12:02:00]: - Specify an absolute path to your Xcode installation in your Fastfile
[12:02:00]: ▸ xcode_select "/Applications/Xcode8.app"
[12:02:00]: 
[12:02:00]: - Manually update the path using
[12:02:00]: ▸ sudo xcode-select -s /Applications/Xcode.app
```

paulschreiber:
Here are the build errors:
```
[...]

PR #2941, opened 2025-05-06
fix: Fix the iOS build workflow xcode version to 16.2

knipec:
## Description
The iOS build is broken on main. Will this fix it? Let's see... (Update: yes!)

### Related Issues
Addresses #2940

Filed https://github.com/techmatters/terraso-mobile-client/issues/2943 to follow up in a couple months to see if we can change it back to using the latest Xcode version

### Verification steps
If the build associated with this PR succeeds, that's a very positive sign


johannesparty (approved):
looks good for now.  hopefully apple or someone else fixes before we are forced to upgrade xcode-version.

Issue #2797, opened 2025-01-09
Terminology updates - Soil properties score info sheet


DerekCaelin:
Update the terminology on the soil ID PropertiesScoreInfoSheet to reflect our use of "observations" rather than "data" for user captured information.

## Design 
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10731-13946&t=zcoy9j4qWEdeRXCp-0)

<img src="https://github.com/user-attachments/assets/dff3332e-541f-491f-8372-6bc8144a4249" width="300px">

## Acceptance Criteria

- [ ] 1st paragraph: the data entered -> the observations entered
- [ ] 2nd paragraph: user-entered data -> user-entered observations
- [ ] 3rd paragraph: the more data -> the more observations
- [ ] 4th paragraph: the following properties in -> the following in

Nongoal: "Cracks" reference already added

knipec:
@CourtneyLee333 Tangential question: I notice that in Spanish, "score" is translated as "calificación" for location score and as "puntuación" for soil properties score. 
Just wanted to mention it in case you think we should bother standardizing on one? I don't know enough Spanish to have an opinion though :)

CourtneyLee333:
Yeah, I see that on the soil info sheet as well. Let's leave it for now, and if we hear confusion about it we can look into standardizing.

Issue #2796, opened 2025-01-09
Terminology Updates - Location Score Info Sheet


DerekCaelin:
Add a paragraph to the Location Score Info Sheet, which clarifies that the location score is not dependent on user observations.

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10731-13946&t=zcoy9j4qWEdeRXCp-0)
<img src="https://github.com/user-attachments/assets/83d47037-64a2-42e5-85ff-90536087d6c2" width="300px">

## Acceptance Criteria
- [ ] Add paragraph: 
`Location score is based solely on soil maps and the site’s location. This does not take into account observations that have been entered by users at the site.`

knipec:
@DerekCaelin @CourtneyLee333 Quick question -- in the 2nd paragraph, the Figma says "LandPKS bases soil matches..." but the app currently says "LandPKS Soil ID bases soil matches..." 
Would you like me to change it to "LandPKS bases soil matches" per the Figma?

CourtneyLee333:
For both, let's use what's in the Figma design. So "LandPKS bases soil matches" in the second paragraph. And unbold "location score". Thanks!

Issue #2795, opened 2025-01-09
Terminology Updates - Soil ID Top Soil Matches sheet


DerekCaelin:
Update the terminology on the soil id Top Soil Matches info overlay sheet to reflect our use of "observations" rather than "data" for user captured information.

## Design 
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10731-13946&t=zcoy9j4qWEdeRXCp-0)

## Acceptance Criteria
<img src="https://github.com/user-attachments/assets/8b35f55c-8fef-4aa1-b162-31305e8324c9" width="300px">

- [ ] recommended data -> recommended observations

Issue #2793, opened 2025-01-09
Terminology Updates - Depths sheet


DerekCaelin:
Update the terminology on the soil depths sheet to reflect our use of "observations" rather than "data" for user captured information.

## Design 
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10731-13538&t=zcoy9j4qWEdeRXCp-0)

## Acceptance Criteria
<img src="https://github.com/user-attachments/assets/c906d17c-c9ab-49ee-888b-aa8cd8308b30" width="300px">

- [ ] Soil Data Inputs -> Show/hide soil observations

PR #2961, opened 2025-05-08
feat: Update terminology from 'data' to 'observations'

knipec:
## Description
- Make updates to terminology to refer to info collected about the soil as "observations" instead of "data".
- Also a couple "LandPKS Soil ID"s become "LandPKS" 
Should match ["Data terminology updates" in Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10731-12543&p=f&t=rru6wVtHf5EnWHQY-0)

### Related Issues
Addresses #2790, #2791, #2792, #2793, #2794, #2795, #2796, #2797

### DIFF
--- a/dev-client/src/translations/en.json
+++ b/dev-client/src/translations/en.json
@@ -63,7 +63,7 @@
         },
         "create": {
             "title": "Create Site",
-            "description": "Create a site here to save this location, collect data, and see more detailed reports.",
+            "description": "Create a site here to save this location, enter observations, and see more detailed reports.",
             "button_label": "Create Site Here",
             "gps": "Use my current location (GPS)",
             "pin": "Use map pin",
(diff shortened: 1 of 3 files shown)


dev-client/src/translations/en.json
```diff
@@ -327,7 +327,7 @@
             "delete_site": "Delete Site",
             "delete_site_modal": {
                 "title": "Delete site?",
-                "body": "This will permanently delete the site {{siteName}} and all of its observations. This can’t be undone.",
+                "body": "This will permanently delete the site's data and observations. This can’t be undone.",
```
paulschreiber, line 330: issue: we should show the site name in a destructive operation modal.

paulschreiber (changes requested):

CourtneyLee333:
@paulschreiber I agree with your comment, and it appears the changes have already been implemented. The same should be done for the delete project dialog.

dev-client/src/translations/en.json
knipec: Done, and updated the project string to include projectName as Courtney recommended!

paulschreiber (approved):

PR #2963, opened 2025-05-13
fix: lint-staged command works in subdirectory

paulschreiber:
## Description
lint-staged command works in subdirectory

Based on https://github.com/lint-staged/lint-staged/issues/621#issuecomment-2683043450

### DIFF
--- a/dev-client/package.json
+++ b/dev-client/package.json
@@ -147,11 +147,11 @@
     "jest"
   ],
   "lint-staged": {
-    "*.{ts,tsx}": [
-      "eslint --max-warnings 0 --ignore-path .gitignore --fix"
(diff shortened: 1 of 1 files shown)


dev-client/package.json
knipec, line 1: Issue, maybe?: Since earlier today I now have 5 entries that were added to my `git stash` that say "lint-staged automatic backup (<some commit hash>)". I haven't figured out exactly what is/was causing this, so maybe it was with some previous iteration of what we were trying. Mentioning in case you have any thoughts on it.

knipec: Ok nvm, I think this was at an intermediate stage, I haven't been able to repro with the new `lint-staged`

PR #2965, opened 2025-05-14
Feat/uuid11 types not removed

knipec:
## Description
If the test run fails that would make sense to me.
If not, I have no idea what was going on in the other test run when I tried to bump client-shared

## Related to
https://github.com/techmatters/terraso-mobile-client/pull/2939
https://github.com/techmatters/terraso-mobile-client/pull/2964 (other test)


knipec:
Note that the "test" workflow succeeded; it only appears to be failed for Conventional Commits reasons

PR #2967, opened 2025-05-16
fix: bump terraso-client-shared commit hash without bumping uuid version

knipec:
## Description
Alas! I thought it was ok to bump the uuid package version from 10 to 11, but it seems it causes the app to crash. So, sadly, taking client-shared back to a commit hash that uses uuid 10.0.0, and use uuid 10.0.0 in mobile-client as well.

Opening https://github.com/techmatters/terraso-mobile-client/issues/2968 to follow up on this.

### Related Issues
A tangential follow-up to a change for https://github.com/techmatters/terraso-backend/issues/1618 

### Verification steps
- App does not crash after login
- Tests pass

### DIFF
--- a/dev-client/package.json
+++ b/dev-client/package.json
@@ -81,7 +81,7 @@
     "react-native-svg": "^15.11.2",
     "react-native-tab-view": "^4.0.12",
     "reduce-reducers": "^1.0.4",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#a4df832",
+    "terraso-client-shared": "github:techmatters/terraso-client-shared#75bef62",
     "use-debounce": "^10.0.4",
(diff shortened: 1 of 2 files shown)


dev-client/package.json
```diff
@@ -81,9 +81,9 @@
     "react-native-svg": "^15.11.2",
     "react-native-tab-view": "^4.0.12",
     "reduce-reducers": "^1.0.4",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#acd2844",
+    "terraso-client-shared": "github:techmatters/terraso-client-shared#75bef62",
```
knipec, line 84: This is the commit where client-shared uses uuid v10 but the updated backend

paulschreiber (approved):

Issue #2978, opened 2025-06-05
Global Soil ID: Client supplies description & management for global soils


knipec:
on't show HTML in soil description and management to user
- `<br>` in English
- accent marks like `\u00e1` in Spanish
- maybe other stuff??

We should transform the strings we receive from the backend so we don't show the text to users with the HTML still in it.

-----
From [garo's slack post](https://tech-matters.slack.com/archives/C0376RCCXD2/p1749061990743019?thread_ts=1749056317.575569&cid=C0376RCCXD2), raw output for a random point:
```
"siteDescription": {
                        "Description_en": "Calcaric Cambisols are widely occurring soils with limited soil development. <br> These soils are calcareous in a soil layer below the surface.",
                        "Management_en": "In temperate regions the soils will have a high content of basic cations (potassium, calcium and magnesium), while in more humid regions they may have lower levels of soil nutrients. <br> These soils are well suited for crop production, and can be intensively used. <br> Fertilization should be used to improve productivity, and irrigation can be included to further boost productivity. ",
                        "Description_es": "Los Cambisoles Calc\u00e1reos son suelos muy extendidos con un desarrollo limitado del suelo. <br>Estos suelos son calc\u00e1reos en una capa del suelo por debajo de la superficie.",
                        "Management_es": "En las regiones templadas los suelos tendr\u00e1n un alto contenido de cationes b\u00e1sicos (potasio, calcio y magnesio), mientras que en las regiones m\u00e1s h\u00famedas pueden tener niveles m\u00e1s bajos de nutrientes del suelo. <br>Estos suelos son muy adecuados para la producci\u00f3n de cultivos, y pueden utilizarse de forma intensiva. <br>La fertilizaci\u00f3n debe utilizarse para mejorar la productividad, y puede incluirse el riego para aumentar a\u00fan m\u00e1s la productividad.",
                        "Description_ks": "Calcaric Cambisols are widely occurring soils with limited soil development. <br> These soils are calcareous in a soil layer below the surface.",
                        "Management_ks": "In temperate regions the soils will have a high content of basic cations (potassium, calcium and magnesium), while in more humid regions they may have lower levels of soil nutrients. <br> These soils are well suited for crop production, and can be intensively used. <br> Fertilization should be used to improve productivity, and irrigation can be included to further boost productivity. ",
                        "Description_fr": "Calcaric Cambisols are widely occurring soils with limited soil development. <br> These soils are calcareous in a soil layer below the surface.",
[...]

PR #2984, opened 2025-06-24
feat: move global soil match 'description' and 'management' to client i18n strings

knipec:
## Description
Since the English and Spanish strings for global soils' `description` and `management` were in a database that we'd be hosting anyway, we figured we could put them in the client internationalization files and use those strings instead.

From [discussion on slack](https://tech-matters.slack.com/archives/C0376RCCXD2/p1749076822486809): 
- This allows us to add and edit the translation strings more easily, without going into the different database deployments to update what's in there
- This means a slightly larger app bundle, but less sent over the network each time they run soil ID
- Note this is only for global soils, and only for these two pieces of soil info. The US soils will still retrieve all their info from the backend. Global soils will still retrieve other pieces of soil info from the backend.

Also, add a test for what is displayed for global vs US soil match info content.

### Related Issues
Addresses #2978 

### Verification steps
- Look at a soil match for a global location, confirm they conform to the design
- Look at a soil match for a US location, confirm they conform to the design


dev-client/src/screens/LocationScreens/components/soilInfo/SoilInfoDisplay.tsx
```diff
@@ -73,22 +74,69 @@ type SoilInfoDisplayUSProps = {
...
+  const displaySoilSeries = getSoilSeriesDisplayText(
+    soilInfo.soilSeries,
+    t,
+    i18n,
+  );
 
   return (
     <Column space="0px">
-      {soilInfo.soilSeries.description && (
+      {displaySoilSeries?.description && (
```
garobrik, line 129: nit: following the code, it seems like `displaySoilSeries` should never need a `.?`?

dev-client/src/translations/en.json
```diff
@@ -863,6 +863,888 @@
...
+                "management": "These soils have poor productivity, and require regular irrigation. Without irrigation these soils should not be cultivated, and would best be managed with native vegetation, or some type of arid-adapted vegetation. The higher clay content in the subsoil may aid in water retention. However, these soils should not be farmed without irrigation. "
+            },
+            "petric_gypsisols": {
+                "name": "Petric Gypsisols",
+                "description": "Petric Gypsisols occur in arid climates with subsoil accumulations of gypsum. Although the dry climate restricts cultivation of these soils, the soils can be productive when gypsum content is relatively low in surface and subsurface horizons and irrigation is possible. Petric Gypsisols have a root-limiting, cemented, petrocalcic horizon within 1 m of the soil surface. ",
+                "management": "These soils have poor productivity, and require regular irrigation. Without irrigation these soils should not be cultivated, and would best be managed with native vegetation, or some type of arid-adapted vegetation. "
```
garobrik, line 1276: question: based on these descriptions, it seems like it would be straightforward to create an entry for `Gypsisols` by only using the text that is present for every subtype of `Gypsisols`, do you wanna do that here in this PR?

garobrik (approved):
LGTM!

dev-client/src/translations/en.json
knipec: Ah yes, thanks! Will add to english and spanish

PR #2987, opened 2025-06-27
feat: Show version number on login screen

knipec:
## Description
Version number will appear on login screen in the same way it does in the User Settings screen, except white

On development environment:
Screenshot 2025-06-26 at 4 45 00 PM

### Related Issues
Implements https://github.com/techmatters/terraso-product/issues/950


knipec:
Ugh the sentry bot is showing errors based on an intermediate state on my local branch -- you can tell from the video replay that it's displaying strings that aren't part of the VersionIndicatorComponent text

PR #2989, opened 2025-07-03
fix: correctly translate global soil names

knipec:
## Description
Sooo I had failed to actually translate the global soil names in most places. This change does that. It's a bit awkward as we only want to translate when the dataRegion is global. 
The main functions are in `globalSoilI18nFunctions.ts`, and are used in several components.

Also add a test with somewhat sketchy mocking of i18n objects

### Related Issues
Addresses https://github.com/techmatters/terraso-product/issues/1311


knipec:
> question: i thought we wanted to translate the soil series name? like Luvic chernozems would be displayed as Chernozems Lúvicos for spanish users?

Ah yes true, thank you 🤦  I will look into why this isn't working then 🤔

garobrik (approved):
LGTM!

PR #2996, opened 2025-07-10
fix: Update client code to use new soilMatches endpoint for 1.3

knipec:
## Description
Previously it was fetchDataBasedSoilMatches, now that is being used to support 1.2. And for 1.3 it's just fetchSoilMatches. 

To fix the issue where the new backend is incompatible with both 1.2 and 1.3 clients simultaneously (details [in slack here](https://tech-matters.slack.com/archives/C0376RCCXD2/p1752170651976809))

### DIFF
--- a/dev-client/package.json
+++ b/dev-client/package.json
@@ -81,7 +81,7 @@
     "react-native-svg": "^15.11.2",
     "react-native-tab-view": "^4.0.12",
     "reduce-reducers": "^1.0.4",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#93f44f4bde8ba666f24994e33a3301b64b24250e",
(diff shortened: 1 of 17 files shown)


dev-client/package.json
```diff
@@ -81,7 +81,7 @@
     "react-native-svg": "^15.11.2",
     "react-native-tab-view": "^4.0.12",
     "reduce-reducers": "^1.0.4",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#93f44f4bde8ba666f24994e33a3301b64b24250e",
+    "terraso-client-shared": "github:techmatters/terraso-client-shared#ac601e6cf937c5c55fba36e7e0bd308054037d46",
```
garobrik, line 84: reminder: update commit when `client-shared` branch is merged

garobrik (approved):

PR #2998, opened 2025-07-15
fix: get elevation for all locations, not just US locations

### LINKED ISSUE #2664: Obtain valid elevation in non-us  locations sus
For locations outside the U.S., obtain elevation

paulschreiber:
## Description
Use open-meteo elevation API instead of the nationalmap.gov API. This allows for global elevation data.

### Related Issues
Fixes #2664.
Removes workaround from https://github.com/techmatters/terraso-mobile-client/pull/2991

### DIFF
--- a/dev-client/src/model/elevation/elevationService.ts
+++ b/dev-client/src/model/elevation/elevationService.ts
@@ -21,8 +21,8 @@ export const getElevation = async (
   lat: number,
   lng: number,
 ): Promise<number | undefined> => {
-  // 1. TypeScript requires values passed to URLSearchParams strings,
-  //    which is why we convert the floats to strings.
+  // TypeScript requires values passed to URLSearchParams to be strings,
+  // which is why we convert the floats to strings.
   const queryString = new URLSearchParams({
     longitude: formatCoordinate(lng),
     latitude: formatCoordinate(lat),
--- a/dev-client/src/model/elevation/elevationService.ts
+++ b/dev-client/src/model/elevation/elevationService.ts
@@ -17,47 +17,27 @@
 
 import {formatCoordinate} from 'terraso-mobile-client/util';
 
-const isErrorResult = (result: string) => {
-  const OUT_OF_RANGE_MESSAGE = 'Invalid or missing input parameters.';
-  const UNABLE_TO_PROCESS_MESSAGE = 'Unable to complete operation.';
-
-  // Call failed.  [Failed cloud operation: Open, Path: /vsimem/_0000096E.aux.xml]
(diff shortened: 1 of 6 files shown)


knipec (commented):
I haven't looked into it, but I see that with these changes, when I view a temporary location popover by clicking in the middle of the ocean, I get a crash: 
Screenshot 2025-07-16 at 1 03 15 PM

dev-client/src/model/elevation/elevationService.ts
```diff
@@ -17,47 +17,27 @@
...
-      `https://epqs.nationalmap.gov/v1/json/?${queryString}`,
+      `https://api.open-meteo.com/v1/elevation/?${queryString}`,
     );
-    const result = await response.text();
-    if (isErrorResult(result)) {
+    if (response.status !== 200) {
       elevation = undefined;
     } else {
-      elevation = parseFloat(JSON.parse(result).value);
+      const result = await response.text();
```
knipec, line 39: Suggestion: Could use `response.json()` here since as far as I can tell, we're sure the response will be json on success

dev-client/src/model/elevation/elevationService.ts
```diff
@@ -17,47 +17,27 @@
...
   //    which is why we convert the floats to strings.
-  // 2. This API uses X for longitude and Y for latitude. That's not a typo.
   const queryString = new URLSearchParams({
-    x: formatCoordinate(lng),
-    y: formatCoordinate(lat),
-    units: 'Meters', // TODO: switch based on user preference
+    longitude: formatCoordinate(lng),
+    latitude: formatCoordinate(lat),
   });
   let elevation;
```
knipec, line 30: Suggestion: Might help with clarity to explicitly set `elevation = undefined` here, perhaps instead of on line 37 -- that way it is clear in the catch block as well

knipec (approved):

dev-client/src/model/elevation/elevationService.ts
```diff
@@ -17,47 +17,27 @@
...
-  }
-
-  return false;
-};
-
 export const getElevation = async (
   lat: number,
   lng: number,
 ): Promise<number | undefined> => {
   // 1. TypeScript requires values passed to URLSearchParams strings,
```
knipec, line 24: Nitpick: Can remove the `1.` from the comment now that there's not a `2.` :)

dev-client/src/model/elevation/elevationService.ts
paulschreiber: If you write `let elevation = undefined`, the linter changes it back to `let elevation`. I did simplify the if/else logic though.

Issue #2775, opened 2025-01-07
Soil ID Screen: "No soil map data" Alert text doesn't wrap


DerekCaelin:
### App Version

224

### Platforms

- [ ] iOS
- [X] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

Text runs off the screen with a soil ID alert box when no soil map is available

### Steps to Reproduce

1. Create a temporary pin in a place without soil maps (eg the ocean)
2. Navigate to the Soil ID screen
3. See text

### Expected behavior

Text wraps normally

### Actual behavior

Text doesn't wrap: 
<img src="https://github.com/user-attachments/assets/80132042-972a-46ea-a187-0a6fb3ee00c1" width="200px">

### Additional context

I think that this is medium pri because the error message is unreadable. I assume this is low effort to fix, or may be resolved with alert box component work?
https://github.com/techmatters/terraso-mobile-client/issues/2366

DerekCaelin:
I think that this is medium pri because the error message is unreadable. I assume this is low effort to fix, or may be resolved with alert box component work?
https://github.com/techmatters/terraso-mobile-client/issues/2366

knipec:
Triage notes:
- Try the "flex: 1" fix that we've seen before?

knipec:
I think this is because if you have a component hierarchy like <Text><SomethingElse><Text>, something weird happens with React Native and it prevents the text from wrapping? I recall we had a similar issue with bullet points failing to wrap. 

When we fix this in AlertMessageBox, it should also be fixed in ErrorMessageBox. 

Note to self: `fix/messagebox-text-wrap` branch :)

knipec:
@CourtneyLee333 This isn't in for QA yet, but FYI I got a little carried away and implemented the `Alert` from the component library (though I named it `MessageBox` -- if you don't like that name, I can rename it to Alert; that just means a different thing in Javascript). 

Here are the 'warning' and 'error' variants of the MessageBoxes, with generic text for the `title` and `bodyText` props:
<img width="408" height="237" alt="Image" src="https://github.com/user-attachments/assets/21daf611-487d-4e3d-bbcb-6ed1f0ad1867" />

and here are some of the more specific message boxes we have:
<img width="407" height="556" alt="Image" src="https://github.com/user-attachments/assets/126418c1-e88e-43da-ad05-92ce8847d0e6" />

CourtneyLee333:
[...]

PR #3002, opened 2025-07-17
Fix: wrapping in AlertMessageBox & implement component library MessageBox

knipec:
## Description
There is an issue with component hierarchies like <Text><SomethingElse><Text> where something weird happens with React Native and it prevents the text from wrapping? I recall we had a similar issue with bullet points failing to wrap. And it should be fine to limit the MessageBox to expect only strings for its body text, not arbitrary React children.

To confirm with Courtney: 
- Ok to rename "Alert" in Figma to "MessageBox"?
- Colors

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2775
Also implements https://github.com/techmatters/terraso-mobile-client/issues/2366 per component library spec


dev-client/src/components/messages/MessageBox.tsx
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright © 2024 Technology Matters
```
paulschreiber, line 2:
```suggestion
 * Copyright © 2025 Technology Matters
```

paulschreiber (approved):

PR #3003, opened 2025-07-17
fix: Add watchman shutdown-server to clean-watchman

knipec:
## Description
I found that running 
```
watchman watch-del-all
watchman shutdown-server
```
fixed a watchman error in my local environment, that `make clean-watchman` alone did not fix.
Because `npm run clean-watchman` already does the watch-del-all part, I only added the shutdown-server part.
Source: https://stackoverflow.com/questions/49443341/watchman-crawl-failed-retrying-once-with-node-crawler 

However, I'm not sure if/how this will affect watchman in other repos -- like if you have a backend, mobile-client, and web-client running at the same time and run this in mobile-client, I'm not sure if you'd need to do something extra to make backend and web-client continue to hot reload?
[[TODO: test that if reviewers aren't sure]]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ setup-git-hooks:
 	@echo "git hooks installed"
 
 clean-watchman:
-	cd dev-client && npm run clean-watchman && watchman shutdown-server
+	cd dev-client && npm run clean-watchman
 
 clean-simulators:
 	{ type -p xcrun && xcrun simctl shutdown all && xcrun simctl erase all; }
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ setup-git-hooks:
 	@echo "git hooks installed"
 
 clean-watchman:
-	cd dev-client && npm run clean-watchman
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ setup-git-hooks:
 	@echo "git hooks installed"
 
 clean-watchman:
-	cd dev-client && npm run clean-watchman
+	cd dev-client && npm run clean-watchman && watchman shutdown-server
```
knipec, line 9: TODO: Confirm Iif/how this will affect watchman in other repos -- like if you have a backend, mobile-client, and web-client running at the same time and run this in mobile-client, I'm not sure if you'd need to do something extra to make backend and web-client continue to hot reload?

paulschreiber: As far as I know, watchman is only used by the React Native bits / the mobile client.

paulschreiber: issue: modify package.json so `clean-watchman` calls both commands. that way you get the same output from `npm run clean-watchman` and `make clean-watchman`.

garobrik (approved):

Issue #2994, opened 2025-07-09
App must target Android 15 (API level 35) or higher (by Aug 30)


knipec:
### App Version
1.3

### Description
From the [Google Play Console](https://play.google.com/console/u/0/developers/7394238132422662749/app/4976046501622296386/policy-center/issues/4987973894392924235/details):
<img width="339" height="250" alt="Image" src="https://github.com/user-attachments/assets/06fb1293-53c5-40cb-9509-e5fa18c33edd" />

- We suspect that if we update `expo` it may accomplish this. 
- We also have been holding off on an expo update because of some issue with updating `rnmapbox`? So we should check if things work with version 10.1.39 of rnmapbox. From Paul: we can make a branch with new RN + new Expo + new Mapbox if you think that makes more sense

https://github.com/techmatters/terraso-mobile-client/issues/2974 likely could be closed out if that mapbox update is required to enable this

### Additional context
[Slack conversation](https://tech-matters.slack.com/archives/C0376RCCXD2/p1751549069533089?thread_ts=1751400339.015759&cid=C0376RCCXD2)

knipec:
Guidance: 
- https://stackoverflow.com/questions/79686551/app-must-target-android-15-api-level-35-or-higher

knipec:
This should be updated in v252beta once that's out

PR #3006, opened 2025-07-22
fix: use expo-build-properties to set targetSdkVersion to 35 for android

knipec:
## Description
This is the simple solution to make the app target Android 15 (API level 35). 
I'd tried to update Expo from 52 to 53, but had a lot of issues resolving peer dependency versions between packages. 
So instead I just set it via expo-build-properties, as described: https://docs.expo.dev/versions/latest/sdk/build-properties/

I did specify the `compileSdkVersion` and `buildToolsVersion` using the values expo is already using by default, as it seemed like other folks online kept them in. But I imagine it doesn't much matter if those are included or not.

Also add `expo-build-properties` to the depcheck ignore list ([as noted here](https://github.com/upleveled/preflight/issues/628))

### Related Issues
Addresses #2994 

### Verification steps
When I delete the `android` directory and regenerate it with `npm run android`, gradle.properties now includes `android.targetSdkVersion=35`.

### DIFF
--- a/dev-client/package.json
+++ b/dev-client/package.json
@@ -45,6 +45,7 @@
     "expo-apple-authentication": "~7.1.3",
     "expo-auth-session": "~6.0.3",
     "expo-av": "~15.0.2",
+    "expo-build-properties": "~0.13.3",
     "expo-camera": "^16.0.18",
     "expo-constants": "^17.0.8",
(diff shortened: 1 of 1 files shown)


garobrik (approved):
am i understanding right that the purpose of specifying these values manually is to enable the expo 53 upgrade?

Issue #3004, opened 2025-07-18
Bump version


knipec:
Similar to https://github.com/techmatters/terraso-mobile-client/pull/2882/files

DerekCaelin:
This says 1.3.1 in the settings screen

PR #3007, opened 2025-07-22
chore: bump version to 1.4

knipec:
## Description
Bumps the mobile client app version number to 1.4
Same edits as the bump to 1.3 in https://github.com/techmatters/terraso-mobile-client/pull/2882 

### Related Issues
Addresses #3004


garobrik (commented):
question: why do we immediately bump to `1.4` instead of waiting to see whether we want to do a `1.3.1` or a `1.4`?

paulschreiber (approved):

Issue #2751, opened 2025-01-02
Display "Error" when SoilID returns "ALGORITHM_FAILURE"


DerekCaelin:

DerekCaelin:
@shrouxm paul suggested you'd be a good fit, would you please review and size this issue?

knipec:
Should be available when v252beta happens

PR #3008, opened 2025-07-24
fix: Show 'Error' instead of 'No matches' on algorithm failure

knipec:
### Related Issues
Addresses #2751 

### Verification steps
Look at the temporary location callout at a location that has an algorithm error (currently 43.04364, -92.30141 does). Make sure the callout says "ERROR" and there is a red alert on the soil id matches screen that says "Can't fetch soil map"


garobrik (approved):
I approve this error message.

Issue #2752, opened 2025-01-02
Soil ID Screen Alert: Display "Error" message in various scenarios


DerekCaelin:
On the Soil ID screen, provide the user with an alert message in key scenarios where an "Error" has occured. 

## References
- https://github.com/techmatters/terraso-product/issues/1180
- https://github.com/techmatters/soil-id-algorithm/issues/215

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=11352-32803&t=164Npv6xM28b2mMS-0)
![Image](https://github.com/user-attachments/assets/db62ba6c-a90e-4381-a92c-1457ed65aa4a)

## Acceptance Criteria
- Display Red alert: `Can’t fetch soil map` `LandPKS was unable to fetch the soil map. Try stopping and restarting the app. If the issue continues, try again later or submit a support request through the website landpks.terraso.org. `
   - [ ] Network error, can’t talk to backend
   - [ ] Soil Id algorithm crashes, backend returns “ALGORITHM_FAILURE”
   - [ ] Frontend times out waiting for backend: (20s)

DerekCaelin:
I'm OK with spending the two additional points to "do it right".

1. I _think_ that is ok for the error message to persist until the user closes and opens the app (or, I assume, enters observations). We could improve this later if we find it's a problem. 
2. Please do investigate what effort it would take to have the notification update if it suddenly hears back from the server after 20s.

PR #3011, opened 2025-08-01
feat: Soil ID should timeout after 20 seconds

knipec:
## Description
- Add a `ClientSoilIdStatus` that includes `'TIMEOUT'` as a possible status, to distinguish from `'ALGORITHM_FAILURE'`
- Use `'TIMEOUT'` if soil id doesn't respond in 20 seconds, and show "ERROR" in that case
- If soil id responds after the timeout (since the Promise will still be active), dispatch an action to update with the actual response info. There are two separate codepaths for tempLocation and site, which I feel like could maybe be refactored to use inheritance, but I didn't feel familiar enough with Typescript invest the time in doing it.

### Checklist
- [x] Still gotta write tests

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2752

### Verification steps
Test if "ERROR" or match(es) are shown for: 
- Temp loc + site
- Global + US
- Backend returns faster than timeout + slower 
- Selected soil + none
- Popover + Soil ID box on temp location/site dashboard + list of matches on soil id screen


dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
```diff
@@ -22,34 +22,57 @@ import {ThunkAPI} from 'terraso-client-shared/store/utils';
...
 } from 'terraso-mobile-client/model/soilIdMatch/soilIdMatches';
-import {AppState} from 'terraso-mobile-client/store';
+import {
+  updateSiteMatchesAfterTimeout,
+  updateTempMatchesAfterTimeout,
+} from 'terraso-mobile-client/model/soilIdMatch/soilIdMatchSlice';
+import {AppDispatch, AppState} from 'terraso-mobile-client/store';
+
+// TODO-cknipe: Should this live here or elsewhere? Also, change back to 20000
+const TIMEOUT_MS = 20000;
```
garobrik, line 38: nit:
```suggestion
const TIMEOUT_MS = 20_000;
```

dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
```diff
@@ -22,34 +22,57 @@ import {ThunkAPI} from 'terraso-client-shared/store/utils';
...
+  tempLocationEntry,
 } from 'terraso-mobile-client/model/soilIdMatch/soilIdMatches';
-import {AppState} from 'terraso-mobile-client/store';
+import {
+  updateSiteMatchesAfterTimeout,
+  updateTempMatchesAfterTimeout,
+} from 'terraso-mobile-client/model/soilIdMatch/soilIdMatchSlice';
+import {AppDispatch, AppState} from 'terraso-mobile-client/store';
+
+// TODO-cknipe: Should this live here or elsewhere? Also, change back to 20000
```
knipec, line 37: Remove this comment 😅

dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
```diff
@@ -22,34 +22,57 @@ import {ThunkAPI} from 'terraso-client-shared/store/utils';
...
-  __: ThunkAPI,
-) => fetchTempLocationBasedSoilMatches(coords);
+  {dispatch}: ThunkAPI,
+) => fetchTempLocationBasedSoilMatches(coords, dispatch);
+
+const timeoutError = {
+  __typename: 'SoilIdFailure' as const,
+  reason: 'TIMEOUT' as ClientSoilIdFailureReason,
+};
+type TimeoutError = typeof timeoutError;
```
garobrik, line 50: commnet: it feels a bit funny to be simulating GraphQL output here, instead of directly giving the error in the format we eventually convert the GraphQL response to

dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
```diff
@@ -22,34 +22,57 @@ import {ThunkAPI} from 'terraso-client-shared/store/utils';
...
-    return tempLocationEntryForStatus(coords, result.reason);
-  } else {
-    return tempLocationEntryForMatches(
-      coords,
-      result.matches,
-      result.dataRegion,
-    );
-  }
+  const result = await Promise.race([apiPromise, timeoutPromise]);
+  return tempLocationEntry(coords, result);
```
garobrik, line 75: question: can we encapsulate the promise race/timeout logic independently of the soil/API logic?

dev-client/src/model/soilIdMatch/soilIdMatchSlice.ts
```diff
@@ -60,6 +62,20 @@ const soilIdMatchSlice = createSlice({
...
       flushErrorEntries(state.siteDataBasedMatches);
     },
+    updateTempMatchesAfterTimeout: (state, action) => {
+      const coords = action.payload.coords;
+      const key = coordsKey(coords);
+      const updatedResponse = action.payload.response;
+      const soilIdEntry = tempLocationEntry(coords, updatedResponse);
+      state.locationBasedMatches[key] = soilIdEntry;
+    },
+    updateSiteMatchesAfterTimeout: (state, action) => {
```
garobrik, line 72: suggestion: remove `AfterTimeout` from action names

garobrik (commented):
looking good so far! will re-review after tests

dev-client/src/model/soilIdMatch/actions/soilIdMatchActions.ts
```diff
@@ -22,34 +22,57 @@ import {ThunkAPI} from 'terraso-client-shared/store/utils';
...
+      resolve(timeoutError);
+    }, TIMEOUT_MS);
   });
+  const apiPromise = soilIdService
+    .fetchSoilMatches(coords, {
+      depthDependentData: [],
+    })
+    .then(response => {
+      dispatch(updateTempMatchesAfterTimeout({coords, response}));
+      return response;
```
knipec, line 71: This is only needed if the timeout promise has already returned, because nobody will look at the response.

dev-client/src/model/utils/promiseTimeoutUtil.ts
```diff
@@ -0,0 +1,53 @@
+/*
```
garobrik, line 1: this is very nitpick, but this code isn't really related to the app's models, so it might make more sense outside of that folder

garobrik (approved):
LGTM!

PR #3017, opened 2025-08-06
fix: Change soil id warning alert text to not reference 'site'

knipec:
## Description
Minor string update to make the text work for both temporary locations and sites

### Related Issues
Based on comment https://github.com/techmatters/terraso-product/issues/1251#issuecomment-3145050099


garobrik (approved):
mebbe spanish?

PR #3024, opened 2025-08-14
fix: repaired missing privacy_policy_label

johannesparty:
also added command "npm run check-i18n" which will find most missing translation strings. this should probably run automatically at some point.

also moved a few strings into general section to avoid duplication (privacy_policy_link..., and terms_of_service_link...).

conformed the check script found the missing value and now all the values are good.
(it falsely reports projects.sites.sort., but not an actual problem).

### DIFF
--- a/dev-client/find-i18n-keys.mjs
+++ b/dev-client/find-i18n-keys.mjs
@@ -1,0 +1,234 @@
+#!/usr/bin/env node
+// find-i18n-keys.mjs
+// Usage: node find-i18n-keys.mjs [path1] [path2] ... [--catalog path/to/catalog.json] [--json]
+
+// easiest way to run: 'npm run check-i18n'
+// todo: run with other linters automatically; maybe add github action too.
+import fs from 'node:fs';
+import path from 'node:path';
+
+const exts = new Set(['.js', '.jsx', '.ts', '.tsx']);
+const IGNORE_DIRS = new Set([
+  'node_modules',
+  '.git',
+  'dist',
+  'build',
+  '.next',
+  'out',
+  'coverage',
+]);
+
+const args = process.argv.slice(2);
+const jsonOutput = args.includes('--json');
+
+function popFlagWithValue(flag, arr) {
+  const i = arr.indexOf(flag);
+  if (i === -1) return null;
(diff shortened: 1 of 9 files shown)


dev-client/find-i18n-keys.mjs
knipec, line 1: Could this go in the `dev-client/scripts` directory? (+ update path in package.json if so)

dev-client/find-i18n-keys.mjs
```diff
@@ -0,0 +1,234 @@
+#!/usr/bin/env node
```
knipec, line 1: We typically copy-paste the Copyright comment at the top of our files (with the updated year)

dev-client/find-i18n-keys.mjs
knipec, line 1: Could be helpful to add a disclaimer that if components are created that take string keys outside of the format of the existing regexes, missing strings may not be caught.

knipec, line 1: This is awesome, thank you!

I didn't do a line-by-line review of this file, so let me know if you want a more thorough review of it; could help to go over it on a call.

knipec (approved):

dev-client/find-i18n-keys.mjs
johannesparty: done

dev-client/scripts/localization/find-i18n-keys.mjs
```diff
@@ -204,27 +228,32 @@ if (jsonOutput) {
...
     );
-    if (inScanNotCatalog.length === 0) console.log('(none)');
-    else inScanNotCatalog.forEach(k => console.log(k));
+    console.log('  to see the list: npm run check-i18n -- --extras ');
+    if (showKeysInFileNotSource) {
+      inCatalogNotScan.forEach(k => console.log(`  '${k}'`));
+    }
+
+    console.log(
+      `Found ${inScanNotCatalog.length} keys in source not in ${catalogPath}`,
```
knipec, line 244: Nitpick: I'd put this at the top since it's the most critical info IMO

PR #3026, opened 2025-08-19
fix: Update terminology for soil id error alert message

knipec:
## Description
Update the error alert message string on the soil ID screen 

FYI: You can get an error via timeout of 20 seconds (as of [PR 3011](https://github.com/techmatters/terraso-mobile-client/pull/3011)), or if the soil ID algorithm crashes, or if there's a network error.

### Related Issues
One of the changes to address https://github.com/techmatters/terraso-mobile-client/issues/2752


johannesparty (approved):
Looks fine; could include link for the website instead of just the string, but probably not important.  I assume the actual text was approved by others.

PR #3027, opened 2025-08-20
fix: added basic posthog support

johannesparty:
## Description
Add posthog support (issue 1313)

### Verification Steps
Confirmed it is uploading events to newly-created posthog project (doesn't affect legacy landpks)


dev-client/src/App.tsx
```diff
@@ -61,7 +79,15 @@ const store = createStore(persistedReduxState);
 const App = () => {
   return (
     <AppWrappers store={store}>
-      <AppContent />
+      <PostHogProvider
```
paulschreiber, line 82: issue: this belongs in AppWrappers, not here.

dev-client/src/config/index.ts
```diff
@@ -81,4 +81,6 @@ export const APP_CONFIG = {
   googleClientId,
   googleRedirectURI,
   microsoftRedirectURI,
+  posthogApiKey: 'phc_JItUNq4tPzj1S0MNSFIvkqXeQVSKXu0eigpgKGJ9V48',
```
knipec, line 84: This file is public, so I imagine we'd want to put this in .env instead (and update the .env.sample to note its existence. I expect we'd treat it similarly to PUBLIC_MAPBOX_TOKEN or Sentry things. 

Also if we're real sticklers about security we probably don't want this in a public GitHub PR and we may want to get a new API key if possible? Though lmk if this is a special case for some reason.

knipec (commented):
I agree with Paul's comments about file organization here.

johannesparty:
Closing because new PR added with these issues resolved (hopefully).

PR #3033, opened 2025-08-22
build: update node to 22.18.0

paulschreiber:
## Description
update node to 22.18.0


johannesparty:
22.12 to 22.18 seems unlikely to cause problems.  I'm not listed as a reviewer, but I would be happy to approve.

Issue #3020, opened 2025-08-12
1.3.1 localization


knipec:
- Generate files from POEditor
- But wait! There are new strings now that haven't been imported into POEditor, make sure we do something with them! Also, are there any we do NOT want?
- Also do this: https://github.com/techmatters/terraso-product/issues/1305

PR #3044, opened 2025-08-28
feat: Update strings for Spanish and Ukrainian

knipec:
## Description
Follow [the localization steps](https://docs.google.com/document/d/1IEy2LX1_ZSVYTZkAaL9P0IePm2huslViU7BC89KY_Lw/edit?tab=t.0#heading=h.f7emuqbk7q2x) in the Terraso development guide to generate updated `.po` files and `.json` files for the languages we support.

### Related Issues
Addresses #3020


johannesparty (approved):
Looks good.  The comment in WelcomeScreen.tsx confused the check-i18n script, but that is a limitation with the script.  Next month I want to make the script automatic and then we should decide to either make the script better to at least handle a convention where you can still put in a comment.

Of course, I didn't actually check the text, especially UK, but I'm assuming that is good.

Issue #3025, opened 2025-08-19
Crash on "learn more" from pin in the ocean


knipec:
### Description

New elevation API is causing crash in 1.3

### Steps to Reproduce

1. Drop a pin in the ocean
2. Tap "learn more" button on the popover

**Behavior:** Crash with "Error: Text strings must be rendered within a <Text> component."

### Additional context

New elevation API returns `0` in the ocean. This causes 
```
        {elevation && (
          <LocationDetail
            label={t('geo.elevation.title')}
            value={renderElevation(t, elevation)}
          />
        )}
```
to evaluate to `0` because javascript, and the JSX interprets that as text that's not in a <Text> component.

So in this case where elevation is 0, we could:
A) Always show elevation as 'unknown' (on temp location popover & location dashboard)
B) Always show elevation as '0' (on temp location popover & location dashboard)
C) Don't show elevation on the location dashboard, but show it as unknown on the temp location popover (as the code currently makes it seem is desired, but I find a little weird)

knipec:
Triage notes: 
- Option B, and also make sure land and death valley work fine
- Does the API ever return unknown?

knipec:
I expect this to be available to test once v257beta is out

DerekCaelin:
This is addressed in v258.

PR #3051, opened 2025-09-02
fix: elevation 0 shows on temp location callout and on location dashboard

knipec:
## Description
No longer crashes when viewing location dashboard for elevation 0 (for example, in the ocean). 
Negative elevations continue to work fine.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/3025


johannesparty (approved):
Looks good; actually much cleaner now, mainly checking in renderElevation.

Issue #3061, opened 2025-09-16
Ukrainian: Temporary Popover & Dashboard - User reports elevation not displaying


DerekCaelin:
### App Version

1.3.1 (Build 256)

### Platforms

- [ ] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

A user reports that sites in Ukraine do not display elevation.

### Steps to Reproduce

User reports dropping a pin anywhere gives them sites with unknown elevation. I was initally able to reproduce but now can't. 

### Expected behavior

Elevation is displayed on temporary location popover and side dashboard.

**English**
<img width="200" src="https://github.com/user-attachments/assets/328f3380-5655-4fb0-a808-4604695f3bac" />

### Actual behavior

Elevation is "невідомий" - unknown. 

**Ukrainian**
<img width="200"  src="https://github.com/user-attachments/assets/cdd91d1d-fa37-4bb5-af53-0d3f160412a2" />  <img width="200" src="https://github.com/user-attachments/assets/af42596f-a645-46c6-bc56-657ec02abbc2" />

### Additional context

Maybe something about the API?

knipec:
It was quick, so I have a PR out for this (brief description of the issue there): https://github.com/techmatters/terraso-mobile-client/pull/3062 

Though, tangential note: this does make me wonder if there is any number entry that we need to rethink now that we are aiming to support a language that uses commas as decimal separators. How do folks expect to enter coordinates into the map search? 
In English, we'd enter lat/long like: 
<img width="398" height="335" alt="Image" src="https://github.com/user-attachments/assets/0ce71853-1166-4761-9b7e-4f11d9e8967f" />
...in Ukrainian would it be like this? If so, we have somewhat unintuitive behavior. 
<img width="397" height="333" alt="Image" src="https://github.com/user-attachments/assets/99fd0886-9e24-455d-bcf9-72f54daa0452" />

PR #3062, opened 2025-09-16
fix: send English coodinates to elevation API

knipec:
Using Ukrainian localization, decimal separators are commas. So when our language was set to Ukrainian, instead of 
`https://api.open-meteo.com/v1/elevation/?longitude=103.9&latitude=43`
we'd send
`https://api.open-meteo.com/v1/elevation/?longitude=103%2C9&latitude=43`
to the elevation API, and it would return `undefined` instead of the actual elevation. 

Fix this by always sending the coordinates to the API with English localization (but still display everything to the user with proper localization). 

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/3061

### DIFF
--- a/dev-client/src/model/elevation/elevationService.ts
+++ b/dev-client/src/model/elevation/elevationService.ts
@@ -15,7 +15,7 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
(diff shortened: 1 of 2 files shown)


garobrik (approved):
i enjoy this bug 🐞 LGTM!

PR #3079, opened 2025-09-25
feat: posthog events with email

### LINKED ISSUE #3046: Posthog: filter users of the beta track
We have a beta program actively using the app, and don't want to add noise to our data.

Options:
- filter beta users in posthog [using these methods](https://posthog.com/tutorials/filter-internal-users)
- don't include posthog code in beta releases
   - We already do this for some features, where we _include_ feature flags etc

johannesparty:
## Description
- Patch expo-localization so compatible with Xcode 26 which is stricter on switch statements without all cases
- Add identity to PostHog including email, name, and email_domain
- Add platform to PostHog for filtering out non 'production' platform

### Checklist
- [ Y  ] Corresponding issue has been opened
- [ N ] New tests added

- [ N/A ] English strings reviewed and copyedited
- [ N/A ] Strings are localized
- [ N/A ] Images are compressed (TinyPNG or svgo)
- [ N/A ] Sample environment file updated (when environment variables changed)
- [ Y ] Verified on mobile
- [ N/A ] Verified on desktop
- [ N/A ] Verified accessibility ([Accessibility - Web applications](https://docs.google.com/document/d/1VwPDyLqw7r_iISMgDZr1FdAmAqqOIIVIxQqsYA-HaHE/edit?usp=sharing))
### Related Issues
This is the solution to issue 3046 (posthog cohort has been set up for the filtering)
https://github.com/techmatters/terraso-mobile-client/issues/3046

### Verification steps
- Toggle filter button on and off and observe list of events
- The setting looks like this:
image

- Go to posthog properties to update the internal users cohort

### DIFF
--- a/dev-client/patches/expo-localization+16.0.1.patch
+++ b/dev-client/patches/expo-localization+16.0.1.patch
@@ -1,0 +1,14 @@
+diff --git a/node_modules/expo-localization/ios/LocalizationModule.swift b/node_modules/expo-localization/ios/LocalizationModule.swift
+index 73b2b9d..575ff8d 100644
+--- a/node_modules/expo-localization/ios/LocalizationModule.swift
++++ b/node_modules/expo-localization/ios/LocalizationModule.swift
+@@ -123,6 +123,9 @@ public class LocalizationModule: Module {
+       return "roc"
+     case .iso8601:
+       return "iso8601"
++    @unknown default:
++      log.error("Unhandled `Calendar.Identifier` value: \(calendar.identifier), returning `iso8601` as fallback. Add the missing case as soon as possible.")
++      return "iso8601"
+     }
+   }
+ 
--- a/dev-client/src/app/PostHog.tsx
+++ b/dev-client/src/app/PostHog.tsx
@@ -16,7 +16,9 @@
  */
 
 import {ReactNode, useEffect, useRef} from 'react';
-import {AppState} from 'react-native';
+import {AppState, Platform} from 'react-native';
+
+import Constants from 'expo-constants';
 
 import {NavigationContainerRef} from '@react-navigation/native';
 import {PostHogProvider, usePostHog} from 'posthog-react-native';
@@ -112,6 +114,17 @@ export function PostHog({children, navRef}: Props) {
         // batching knobs (tune as you like)
         flushAt: 10,
         flushInterval: 5000,
+        // Provide app version, build, and platform for all events
+        customAppProperties: properties => ({
+          ...properties,
+          $app_version: Constants.expoConfig?.version || 'unknown',
+          $app_build:
+            Platform.OS === 'ios'
+              ? Constants.expoConfig?.ios?.buildNumber
+              : Constants.expoConfig?.android?.versionCode?.toString() ||
+                'unknown',
+          platform: APP_CONFIG.environment || 'development',
+        }),
       }}
       // IMPORTANT: turn off SDK's nav autocapture to avoid nav hook errors
       autocapture={{captureScreens: false}}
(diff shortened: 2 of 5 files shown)


dev-client/src/app/PostHog.tsx
```diff
@@ -15,17 +15,95 @@
...
+
+      // Extract email domain
+      const emailDomain = currentUser.email.split('@')[1];
+      if (emailDomain) {
+        userProperties.email_domain = emailDomain;
+      }
+
+      // Only add name if it exists
+      if (fullName) {
+        userProperties.name = fullName;
```
knipec, line 70: Do we care about the user's name if we already have their email?

dev-client/patches/expo-localization+16.0.1.patch
knipec, line 1: Ooh, good to know about patch-package!

dev-client/src/app/PostHog.tsx
knipec, line 1: General comment: FYI in the PR description, "Fixes #3046" is the thing to avoid saying if you don't want the associated task to be automatically closed on merge. (IMO we should change our PR template to not use "Fixes")

dev-client/src/app/PostHog.tsx
```diff
@@ -36,26 +114,85 @@ export function PosthogLifecycle() {
...
+          $app_version: Constants.expoConfig?.version || 'unknown',
+          $app_build:
+            Platform.OS === 'ios'
+              ? Constants.expoConfig?.ios?.buildNumber
+              : Constants.expoConfig?.android?.versionCode?.toString() ||
+                'unknown',
+          platform: APP_CONFIG.environment || 'development',
+        }),
+      }}
+      // IMPORTANT: turn off SDK's nav autocapture to avoid nav hook errors
```
knipec, line 189: Curious what this is about?

dev-client/src/app/PostHog.tsx
knipec, line 1: Here's my general understanding, if it seems like I'm missing anything, let me know!
- So far we will send posthog events only when the screen navigation changes, plus the one test event on app open. (Though, https://us.posthog.com/project/14604/activity/explore seems to be recording what people click on, is that just what PostHog does by default?)
-  Each event will note the user's email, id, first/last name, plus the environment.
- You can filter out defined cohorts of test users, plus add a filter for the environment [in Posthog Activity](https://us.posthog.com/project/14604/activity/explore)

dev-client/src/app/PostHog.tsx
```diff
@@ -36,26 +114,85 @@ export function PosthogLifecycle() {
...
+      autocapture={{captureScreens: false}}
+      debug={true}>
```
knipec, line 191: Out of curiosity, if this is false, do you know what would be missed?

knipec (approved):
Seems good to me, just have a few small comments and some things I'm curious about

dev-client/patches/expo-localization+16.0.1.patch
johannesparty: yep, it is cool.  suggested by chatgpt to resolve this problem.  it advised against pulling the later version of the expo-localization component because they said you frequently run into problems not using the versions expo selects automatically.  the patch, however, makes the exact same change in a later version of expo-localization.

dev-client/src/app/PostHog.tsx
johannesparty: changed description.

dev-client/src/app/PostHog.tsx
johannesparty: removed redundant setting, but as you probably guessed, this is so we can filter out if platform != production in posthog queries.

Issue #3045, opened 2025-08-29
Posthog: Capture Account, Site, and Project Creation


DerekCaelin:
Add basic data capture around use of the app for key events.

We want to know:
- when where and how are sites getting created
- when are projects created
- when, whether, and how people are adding sites to projects
- what type of accounts are people using for sign in
- are people selecting a soil

## References
* [Posthog resource](https://posthog.com/docs/getting-started/send-events) 

## Acceptance Criteria
**Events:**
- [ ] select soil (note: won't do, rate soil only)
- [x] "site created" - Properties:
       - [ ] latitude
       - [ ] longitude
       - [x] method (tap on map, enter coordinates, etc).
       - [x] Add to project? (UUID)
- [x] "project created" - properties
   - [x] domain name of creator's email (goal: understand how many institution collaborations are using landpks to collaborate)
- [x] add person to project 
- [x] transfer site
- [x] account created -  properties: 
     - [x] method (Google, Apple, Microsoft)

DerekCaelin:
@johannesparty I'm looking at site_created event and I see multiple values in latitude and longitude, respectively. There appears to be a larger than expected unique count of sites as well. [Link to insight](https://us.posthog.com/insights/new#insight=TRENDS&q=%7B%22kind%22%3A%22InsightVizNode%22%2C%22source%22%3A%7B%22kind%22%3A%22TrendsQuery%22%2C%22series%22%3A%5B%7B%22kind%22%3A%22EventsNode%22%2C%22math%22%3A%22total%22%2C%22name%22%3A%22site_created%22%2C%22event%22%3A%22site_created%22%7D%5D%2C%22version%22%3A2%2C%22interval%22%3A%22day%22%2C%22dateRange%22%3A%7B%22date_to%22%3Anull%2C%22date_from%22%3A%22-14d%22%2C%22explicitDate%22%3Afalse%7D%2C%22trendsFilter%22%3A%7B%22display%22%3A%22ActionsTable%22%2C%22formulaNodes%22%3A%5B%5D%7D%2C%22breakdownFilter%22%3A%7B%22breakdowns%22%3A%5B%7B%22type%22%3A%22event%22%2C%22property%22%3A%22latitude%22%2C%22histogram_bin_count%22%3A10%7D%2C%7B%22type%22%3A%22event%22%2C%22property%22%3A%22longitude%22%2C%22histogram_bin_count%22%3A10%7D%5D%7D%7D%7D%20) - thoughts?

[...]

PR #3080, opened 2025-09-27
feat: posthog specific events

johannesparty:
## Description
- Implement additional specific posthog events 

### Checklist
- [X] Corresponding issue has been opened
- [ ] New tests added

### Related Issues
For issue #3045

### Verification steps
- The new events are soil_match_rating, site_transfer, login, site_created, project_created and team_member_created
- Perform each of these and check posthog for the resulting event and confirm all the desired info is present and correct

Note: the "new_account" field in "login" event will always be false until new backend is deployed.  there is a separate PR for the backend changes.

### DIFF
--- a/dev-client/src/screens/AddUserToProjectScreen/AddUserToProjectRoleScreen.tsx
+++ b/dev-client/src/screens/AddUserToProjectScreen/AddUserToProjectRoleScreen.tsx
@@ -18,7 +18,7 @@
 import {useCallback, useEffect, useState} from 'react';
 import {useTranslation} from 'react-i18next';
 
-import {TabActions} from '@react-navigation/native';
+import {usePostHog} from 'posthog-react-native';
 
 import {addMessage} from 'terraso-client-shared/notifications/notificationsSlice';
 import {ProjectRole} from 'terraso-client-shared/project/projectTypes';
@@ -43,7 +43,6 @@ import {useIsOffline} from 'terraso-mobile-client/hooks/connectivityHooks';
 import {useRoleCanEditProject} from 'terraso-mobile-client/hooks/permissionHooks';
 import {addUserToProject} from 'terraso-mobile-client/model/project/projectGlobalReducer';
 import {AppBar} from 'terraso-mobile-client/navigation/components/AppBar';
-import {TabRoutes} from 'terraso-mobile-client/navigation/constants';
 import {useNavigation} from 'terraso-mobile-client/navigation/hooks/useNavigation';
 import {MinimalUserDisplay} from 'terraso-mobile-client/screens/AddUserToProjectScreen/components/MinimalUserDisplay';
 import {ProjectRoleRadioBlock} from 'terraso-mobile-client/screens/AddUserToProjectScreen/components/ProjectRoleRadioBlock';
@@ -59,6 +58,7 @@ export const AddUserToProjectRoleScreen = ({projectId, userId}: Props) => {
   const {t} = useTranslation();
   const dispatch = useDispatch();
   const navigation = useNavigation();
+  const posthog = usePostHog();
 
   const project = useSelector(state => state.project.projects[projectId]);
   const newUser = useSelector(state => state.account.users[userId]);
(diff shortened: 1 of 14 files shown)


dev-client/src/screens/CreateSiteScreen/CreateSiteScreen.tsx
```diff
@@ -48,9 +52,39 @@ export const CreateSiteScreen = (props: Props = {}) => {
...
+      if (result.payload) {
+        const site = result.payload;
+
+        // Determine creation method based on props
+        let creationMethod = 'manual_entry'; // Default when no coords provided
+        if ('coords' in props && props.coords) {
+          creationMethod = 'tap_on_map';
+        } else if ('elevation' in props && props.elevation !== undefined) {
+          creationMethod = 'current_location';
+        }
```
knipec, line 66: Could you explain this reasoning? I think elevation and coords are both passed when we navigate to the CREATE_SITE screen... And actually I'm confused because that seems to not comply with the Props type definition 🤔 ...Am I missing something here?

dev-client/src/screens/CreateSiteScreen/CreateSiteScreen.tsx
```diff
@@ -48,9 +52,39 @@ export const CreateSiteScreen = (props: Props = {}) => {
...
+
+        // Get project details from the store if site has a project
+        const project = site.projectId ? projects[site.projectId] : null;
```
knipec, line 69: Side note FYI: Currently I expect `project` to always be null here. As far as I recall there is not currently a way to create a site affiliated with a project; you have to create the site, then transfer it to a project. But I see there's code that would allow for a projectId to be passed in, so this is good :)

dev-client/src/screens/SiteTransferProjectScreen/SiteTransferProjectScreen.tsx
```diff
@@ -204,9 +208,49 @@ export const SiteTransferProjectScreen = ({projectId}: Props) => {
...
+        site_name: site?.siteName || 'unknown',
+        from_project_id:
+          site?.projectId === UNAFFILIATED.projectId
+            ? 'unaffiliated'
+            : (site?.projectId as string) || 'unknown',
+        from_project_name: site?.projectName || 'unaffiliated',
+        from_project_privacy: fromProject?.privacy || 'none',
+        to_project_id: projectId,
+        to_project_name: project?.name || 'unknown',
+        to_project_privacy: project?.privacy || 'unknown',
```
knipec, line 234: Nitpick: I might prefer `??` instead of `||` for these. 

(I don't really expect a real projectId to be 0, but if it is javascript would treat that as falsy and return 'unknown'.)

dev-client/src/screens/AddUserToProjectScreen/AddUserToProjectRoleScreen.tsx
```diff
@@ -77,12 +77,29 @@ export const AddUserToProjectRoleScreen = ({projectId, userId}: Props) => {
       dispatch(
         addUserToProject({userId: newUser.id, role: selectedRole, projectId}),
       );
+
+      // Track team member addition in PostHog
+      posthog?.capture('team_member_added', {
+        project_id: projectId,
+        project_name: project?.name,
+        new_member_id: newUser.id,
+        new_member_email: newUser.email,
```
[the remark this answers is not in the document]
johannesparty: Yep, I discussed in some detail with Derek last week.  Ultimately he decided that we already have access to this information in a different way.  I don't really know if that is the right thing, but went with what Derek thought plus it is pretty convenient to be able to filter in PostHog web site itself.

dev-client/src/screens/LocationScreens/components/soilInfo/RateSoilMatchFormSheet.tsx
```diff
@@ -27,24 +32,60 @@ import {FormOverlaySheet} from 'terraso-mobile-client/components/sheets/FormOver
...
+      const locationScore = soilMatch?.locationMatch?.score ?? 0;
+      const propertiesScore = soilMatch?.dataMatch?.score ?? 0;
+      const combinedScore = soilMatch?.combinedMatch?.score ?? 0;
+
+      const soil_rank =
+        soilMatch?.combinedMatch?.rank ?? soilMatch?.locationMatch?.rank ?? 0;
+
+      posthog?.capture('soil_match_rating', {
+        rating: matchRating.toLowerCase(),
+        site_name: siteName || 'temporary_location',
```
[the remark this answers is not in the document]
johannesparty: ok so I could change  "|| 'temporary location' to 'unknown' (or leave it out altogether).  what do you think?

dev-client/src/screens/SiteTransferProjectScreen/SiteTransferProjectScreen.tsx
johannesparty: I decided to leave a ||, but hard to say.  if siteName exists but is, '', is it better to return '' or 'unknown'?
Probably doesn't matter.

knipec (approved):
Looks good to me!

PR #3096, opened 2025-10-21
feat: redux remembers user ratings offline

knipec:
## Description
- User should now be able to select "Yes", "No", or "Unsure" and the UI should reflect that in all places. 
- This should work offline, but will not update the backend yet. (It will update userRatings in local Redux state.)
- Closing and re-opening the app should remember user ratings. (Redux state is saved in app data, and loaded when the app is opened.)
- Claude helped write a bunch of unit tests

**Upcoming work:**
- These changes update the backend when online
- Handle the case where the selected soil disappears from the soil match list (this is the remaining `TODO-cknipe` comment)

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/3107

### DIFF
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -78,7 +78,7 @@
         "react-redux": "^9.2.0",
         "reduce-reducers": "^1.0.4",
         "setimmediate": "^1.0.5",
-        "terraso-client-shared": "github:techmatters/terraso-client-shared#8a36e04487c536a9fcba6b740d97132f021aaec7",
+        "terraso-client-shared": "github:techmatters/terraso-client-shared#34c6d00b115d9ee04b4a73332fe7b4c5fc57e444",
         "use-debounce": "^10.0.6",
         "uuid": "^11.1.0",
         "yup": "^1.7.1"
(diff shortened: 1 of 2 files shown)


dev-client/src/model/soilMetadata/localSoilMetadataActions.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const updateUserRatingsObject = (
+  inputRating: UserRatingEntry,
+  existingRatings: Array<Maybe<UserRatingEntry>>,
+) => {
+  // Remove prior SELECTED rating if new soil got selected
+  let updatedRatings =
+    inputRating.rating === 'SELECTED'
+      ? existingRatings.filter(entry => entry?.rating !== 'SELECTED')
+      : [...existingRatings];
```
johannesparty: Probably clearer this way, but... you could do : existingRatings instead of : [...existingRatings]  because you make a copy in the following filter anyhow.  Probably easier to maintain as-is though.

knipec: Yeah I was debating this and went with the copy since I don't expect it to be big, and I was like "ehh just in case someone otherwise wouldn't realize it should eventually become a copy 🤷"

PR #3099, opened 2025-10-22
fix: add pre-commit hook to prevent committing sensitive environment …

johannesparty:
Add validation to prevent accidental commits of .env and .env.save files which contain secrets and credentials.

## Description
Prevents checking in .env and .env.save

### Verification steps
Try to check in one of those files.  It should fail, but please only try to check in a file without any sensitive keys/tokens.

### DIFF
--- a/dev-client/.husky/pre-commit
+++ b/dev-client/.husky/pre-commit
@@ -1,13 +1,13 @@
 #!/usr/bin/env sh
 
 # Prevent committing sensitive environment files
-if git diff --cached --name-only | grep -qE "(^|/)\.env$|(^|/)\.env\.save$"; then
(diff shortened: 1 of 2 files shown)


dev-client/.husky/pre-commit
knipec: Just making sure I understand the regex -- does this only prevent `.env` and `.env.save` from being committed?

johannesparty: yes, only */.env and */.env.save , and I asked chatgpt to analyze the regex and it explained and gave clear examples.  claude made the regex.

johannesparty: - also just .env and .env.save in base directory is prevented.
- I added .env.local to the list too.

PR #3106, opened 2025-10-27
fix: add safe area insets for Android soft navigation buttons

johannesparty:
This is to fix problem Steve reported 20251027 as seen below:
image

Also fixes this problem:
image

to test, make sure to also test big screen and iOS.  I tested on "Medium Phone API 36" Android 16, iPhone 17 Pro Max Simulator with iOS 26, and "Pixel 2" simulator running Android 11.

### DIFF
--- a/.github/workflows/build-shared-android.yml
+++ b/.github/workflows/build-shared-android.yml
@@ -90,14 +90,16 @@ jobs:
         run: ./gradlew bundleRelease
 
       - name: Sign Android app bundle
+        id: sign_app
         if: ${{ inputs.upload == 'true' }}
-        uses: r0adkll/sign-android-release@v1.0.4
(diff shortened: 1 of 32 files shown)


knipec (commented):
Oo nice!

johannesparty:
Improved code, getting rid of constants (much nicer).  Here are test results:
iOS:
image

Pixel2:
image


Pixel 6a:
image

knipec (approved):
Recommend trying to move the bottomPosition logic into somewhere like `AppContent` that is more general than just the Fab and the BottomNavigator. I'm thinking we may not ever want content (like text or other buttons, in addition to Fabs and BottomNavigators) to be displayed under the Android 3-button menu bar on phones that have it. But if that ends up being hard then I'll approve so you can fix this particular bug.

PR #3110, opened 2025-10-29
fix: upgrade React Native to 0.81.5 to fix Android 16 camera hang

johannesparty:
Upgrades from React Native 0.81.4 to 0.81.5 which includes a fix for the Android 16 bug where launchCameraAsync() hangs indefinitely when camera permissions are already granted.

This affects Pixel devices with September 2025 security update. The fix is at the framework level, so no code changes are needed.

Reference: https://github.com/expo/expo/issues/39480

How to test: use camera to do soil color.

### DIFF
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -61,7 +61,7 @@
         "react": "19.1.0",
         "react-dom": "19.1.0",
         "react-i18next": "^15.7.4",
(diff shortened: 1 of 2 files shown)


knipec (approved):

johannesparty:
This was confirmed to work!  Yahoo!

PR #3114, opened 2025-10-30
fix: centralize safe area bottom and expand its coverage

johannesparty:
## Description
- Added SafeBottomPaddingView.tsx and SafeScrollView.tsx
- Use these components with other component hierarchies as appropriate.

This is definitely cleaner than previous attempts.

I tried adding globally but there were always exceptions and so this seemed safer -- we can add support where needed knowing we won't break anything else.

While this fixes lots of locations, I did not fix everything, but not too hard to add more as needed.

For now I checked in SAFE_AREA_TODO.md too.  I have mixed feelings about checking in stuff like this because it seems messy, but I think it is a good place to keep info like this for later reference, so I went ahead.  You may want to look at that file to get more details.


dev-client/src/components/SafeScrollView.tsx
knipec: Nitpick: Could put these `SafeXYZ` files in a directory

dev-client/src/components/SafeScrollView.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {useMemo} from 'react';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+
+import {IScrollViewProps, ScrollView} from 'native-base';
```
knipec: FYI: In general we intend to move away from native-base. Though it seems throughout the project we're chaotic about which ScrollView we use (`native-base` vs `react-native` vs `react-native-gesture-handler`) so I don't really think one more use is a big deal.

johannesparty:
new changes:
(1) moved the safe... files as recommended by reviewer.
(2) reworked the "notes" screens to account for popup keyboards and small screens with lots of text.

this is Claude's summary of the changes:

 1. SiteNoteForm.tsx ⭐ Core Fix

  Changes:
  - Added Platform import
  - Added scrollEnabled={Platform.OS === 'android' ? true : false}

  Review: ✅ GOOD
  - iOS: Disables TextInput scrolling so outer ScrollView handles all scrolling (fixes bounce-back)
  - Android: Enables TextInput scrolling (necessary for Android's keyboard handling)
  - Clean, minimal change with clear platform-specific logic

  ---
  2. ScreenFormWrapper.tsx 🔧 Major Refactor

  Changes:
  - Added keyboard tracking with useState and useEffect
  - Added totalButtonAreaHeight calculation
  - Replaced SafeScrollViewWithButtons with SafeScrollView + custom padding
  - Added contentContainerStyle with dynamic paddingBottom
  - Added platform-specific buttonBoxStyle for Android keyboard transform
  - Changed from SafeScrollViewWithButtons to bare SafeScrollView with disableAutoPadding={true}
  - Removed all flex={1} and flexGrow={1} from children

  Review: ✅ GOOD
  - Properly tracks keyboard height on both platforms
  - Calculates appropriate padding (totalButtonAreaHeight = buttonRowHeight + max(10, insets.bottom) + 20)
  - Uses paddingBottom in contentContainerStyle (standard React Native approach)
  - Android-specific transform moves buttons up when keyboard appears
  - iOS uses KeyboardAvoidingView with proper offset
  - Clean separation of concerns

  Potential Concerns: ✅ ADDRESSED
  - Inline styles warning - Acceptable for dynamic styles
  - Dependency array for keyboard useEffect - Correctly uses empty array since callbacks don't depend on changing values

  ---
  3. SafeScrollView.tsx 🎯 Enhanced Flexibility

  Changes:
  - Added disableAutoPadding prop
  - Changed from array-based style merging to conditional logic
  - When disableAutoPadding={true}, passes through contentContainerStyle unchanged
  - When disableAutoPadding={false}, applies smart padding logic

  Review: ✅ EXCELLENT
  - Backwards compatible (default disableAutoPadding=false)
  - Fixes the array-style-merging bug (later styles would override earlier ones)
  - Now properly respects explicit paddingBottom in contentContainerStyle
  - Clean, well-documented logic

  ---
  4. EditSiteNoteContent.tsx, AddSiteNoteScreen.tsx, EditPinnedNoteScreen.tsx 🧹 Cleanup

  Changes (identical across all 3):
  - Removed flex={1} from Column
  - Removed flexGrow={1} from Box wrapper

  Review: ✅ CORRECT
  - These flex properties were causing layout constraints inside ScrollView
  - Removing them allows natural content height calculation
  - Consistent changes across all three screens

dev-client/src/components/ScreenFormWrapper.tsx
```diff
@@ -83,46 +108,127 @@ export const ScreenFormWrapper = forwardRef(
...
 
+    // Calculate total button area height including Row's internal padding plus extra clearance
+    // The Box's onLayout measures the container but we need extra space for comfortable scrolling
+    const totalButtonAreaHeight =
+      buttonRowHeight + Math.max(10, insets.bottom) + 20;
+
+    // Track keyboard height to adjust ScrollView padding
+    useEffect(() => {
+      const keyboardWillShow = Keyboard.addListener(
+        Platform.OS === 'ios' ? 'keyboardWillShow' : 'keyboardDidShow',
```
knipec: Why do we listen to different events on ios vs android?

dev-client/src/components/ScreenFormWrapper.tsx
```diff
@@ -83,46 +108,127 @@ export const ScreenFormWrapper = forwardRef(
...
+      );
+
+      return () => {
+        keyboardWillShow.remove();
+        keyboardWillHide.remove();
+      };
+    }, []);
+
+    // Shared button row component
+    const buttonRow = (
```
knipec: Style: Typically things like this would be components instead of named variables. Same with renderScrollContent.

dev-client/src/components/ScreenFormWrapper.tsx
```diff
@@ -83,46 +108,127 @@ export const ScreenFormWrapper = forwardRef(
...
+          size="lg"
+          label={t('general.done')}
+        />
+      </Row>
+    );
+
+    // Shared scroll content - render function to ensure fresh evaluation
+    const renderScrollContent = () => {
+      return (
+        <React.Fragment>
```
knipec: Nit: Shouldn't need the Fragment if there's just one component in it

dev-client/src/components/ScreenFormWrapper.tsx
knipec: Is my understanding here correct? --> 
1. On Android you have to move the buttons up above the keyboard, but on iOS you don't have to -- is that because KeyboardAvoidingView handles making the content (including button box) avoid the keyboard on iOS, but not on android?

2. Buttons are out of the scroll view component, so that they're visible no matter where you are in the scroll? (I didn't get the branch to check so far)

3. We no longer use `SafeScrollViewWithButtons` so that file can be removed, along with references to it in comments

dev-client/src/components/ScreenFormWrapper.tsx
johannesparty: Keyboard avoiding handling is quite different between Android and iOS; I'll provide an overview separately.

dev-client/src/components/ScreenFormWrapper.tsx
johannesparty: Consider it gone.

dev-client/src/components/ScreenFormWrapper.tsx
johannesparty: I tried changing it by defining the buttonRow component here, but then react thought it was new on every render which caused a loop of re-rendering over and over.  this way the component stays the same.  supposedly I could use useMemo() to fix it with components, but then it really would be more complicated than necessary (though perhaps more idiomatic), so for now I'm leaving it if you don't object.

dev-client/src/components/StaticMapView.tsx
knipec: What's the pointerEvents stuff about?

dev-client/package.json
```diff
@@ -160,7 +160,7 @@
     "lodash": "^4.17.21",
     "patch-package": "^8.0.1",
     "prettier": "^3.6.2",
-    "react-devtools": "^7.0.0",
+    "react-devtools": "^4.28.5",
```
knipec: Why did this change happen?

documentation/KeyboardHiding.md
```diff
@@ -0,0 +1,190 @@
...
+
+**Debug Tools**
+
+I added some stuff here to make diagnosing and testing easier. I prefer leaving this stuff as part of the project, but we should discuss.
+
+(1) If you add
+
+    DEBUG_ENABLED="true"
+
+to your .env file, then you will get extra debug on-screen and in the console.log. If you don't want to rebuild, simply edit src/config/index.ts and change the last false to true in the debugEnabled expression near the bottom.
```
knipec: Maybe add it as false to the .env.sample?

dev-client/src/components/ScreenFormWrapper.tsx
```diff
@@ -82,49 +140,248 @@ export const ScreenFormWrapper = forwardRef(
...
-        <KeyboardAvoidingView
-          behavior={Platform.OS === 'ios' ? 'padding' : 'height'} // Handle platform-specific keyboard avoidance
-          style={styles.view}>
+    // Calculate total button area height for comfortable scrolling clearance
+    // buttonRowHeight already includes the Row's internal padding
+    const totalButtonAreaHeight = buttonRowHeight + 20;
+
+    // Track keyboard height to adjust ScrollView padding
+    useEffect(() => {
+      const keyboardWillShow = Keyboard.addListener('keyboardDidShow', e => {
```
knipec: Yay this part makes much more sense to me than it used to when the listeners were platform-specific!

knipec (approved):
I didn't comb through that finely this time, but thanks for looking into this!

dev-client/src/components/safeview/SafeBottomPaddingView.tsx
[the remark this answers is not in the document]
johannesparty: yep, I just removed it.

dev-client/src/components/StaticMapView.tsx
johannesparty: On iOS if you click on the icon in the site list, nothing would happen.  I think the map view was eating the events, so have to pass them through somehow.

documentation/KeyboardHiding.md
johannesparty: Done

docs/iPhone Keyboard.png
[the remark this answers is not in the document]
knipec: Yeah, I agree! I found having an internal wiki helpful at my previous job.

PR #3121, opened 2025-11-04
fix: change done to save on note editing screens

johannesparty:
## Description
Changed "Done" to "Save" as requested.
Will affect all three note editing screens.

image


knipec (approved):

knipec:
The review has a bunch of commits, but I approve the last one that's the string change :)

PR #3130, opened 2025-11-11
Fix: claude.md describes import path patterns (minor) 

knipec:
## Description
Very minor change because claude was using relative paths


johannesparty (approved):
Good

PR #3142, opened 2025-11-24
fix: make site and project top not sticky

lesliehfrancis:
Move sticky header elements (title, filter, offline alert, create button) into scrollable list headers so they scroll with content instead of staying fixed at the top. This increases visible screen real estate for viewing sites and projects.

Changes:
- SiteListBottomSheet: Move title, filter, and offline alert to ListHeaderComponent so they scroll with the site list
- ProjectListScreen: Always render ProjectList with scrollable header containing create button and empty state
- ProjectList: Add header prop for filter modal and offline alert
- ProjectSitesScreen: Move filter modal to scrollable list header

The create button now:
- Scrolls with the project list (not sticky)
- Remains visible when there are no projects
- Is hidden only when offline

🤖 Generated with Claude Code

### Checklist
- [ x] Corresponding issue has been opened
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/dev-client/src/screens/ProjectSitesScreen.tsx
+++ b/dev-client/src/screens/ProjectSitesScreen.tsx
@@ -262,20 +262,30 @@ export function ProjectSitesScreen({
       <SiteCardList
         showButtons={showButtons}
         header={
-          <ListFilterModal
-            searchInput={
-              <TextInputFilter
-                name="search"
-                placeholder={t('site.search.placeholder')}
-                label={t('site.search.accessibility_label')}
+          <>
+            <RestrictByProjectRole role={PROJECT_EDITOR_ROLES}>
+              <Box mb="12px">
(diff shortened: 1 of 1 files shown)


johannesparty (commented):
I think it would make sense to have Carissa review this because they are more knowledgeable with react native.  Please add her to the reviewers list.

dev-client/src/screens/ProjectSitesScreen.tsx
knipec: Is it intentional that the "Transfer sites" button is sticky? Or should that be moved into the SiteCardList header?

knipec (approved):
Nice!

Also, if there's a corresponding github issue for this, it's helpful to note it in the PR description. (That way it will also create a note in the issue to this PR.)

PR #3143, opened 2025-11-24
fix: defer filter and sort application until APPLY button is clicked

lesliehfrancis:
Change the filtering logic to use appliedValues instead of values so that sort and filter choices are only applied to the list when the user clicks the APPLY button, not in real-time as they change the options.

This affects:
- Sites List (SiteFilterModal)
- Projects List (ListFilterModal)
- Transfer Sites screen (ListFilterModal)

Search still applies immediately as expected.

🤖 Generated with Claude Code

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


johannesparty (approved):
Looks good.

PR #3155, opened 2025-12-02
fix: bump to version 1.3.4 so ios builds work

johannesparty:
## Description
need to update to 1.3.4 to do iOS beta build.


knipec (changes requested):
You'll want to run `npm install` or something to update the version in package-lock.json. (Though I'm not sure what the consequences are for not doing it.)

Issue #2397, opened 2024-10-31
Site bottom sheet fills 100% of map when it should have a max of 75% of screen. 


DerekCaelin:
### App Version

199

### Platforms

- [x] iOS
- [X] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

We implemented a max bottom sheet height with #1483, but now the bottom sheet can fill the entire screen. 

### Steps to Reproduce

1. go to sites screen
2. pull sheet up as far as you can
3. see issue

### Expected behavior

max height is 75% of the screen.

### Actual behavior

Max height is 100% of screen. 

### Additional context
<img width="200px" src="https://github.com/user-attachments/assets/fc725380-4eaf-4242-a033-34d767306f24">

CourtneyLee333:
This issue is present in v201 in both Android and iOS.

knipec:
Triage notes: 
- 50 and 100 seem to be the snap points -- search the code for "snap points"

DerekCaelin:
See work done to accomplish https://github.com/techmatters/terraso-mobile-client/issues/1483

PR #3157, opened 2025-12-04
fix: disallow site bottom sheet from filling 100% of map

knipec:
## Description
Turn off dynamic sizing for the SiteListBottomSheet.

[Dynamic sizing](https://gorhom.dev/react-native-bottom-sheet/dynamic-sizing) is used when you want the sheet to automatically size to fit its content, but I don't think there's any need for that here.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/2397

### Verification steps
Try to drag the Sites List up to fully cover the map. When you release, the highest it should snap to is 75% of the available height.


johannesparty (approved):
A nice simple one!

PR #3166, opened 2025-12-09
fix: logout clears the state

johannesparty:
Logout will clear the local state, after checking to make sure no outstanding sync records.
It clears all the state, just like if you uninstalled and reinstalled the app.

There is one setting that is lost this way -- one preference related to color picking.  I don't think this is important or necessarily should be device-specific rather than project or site or user specific.  And doing things this way simplified things a lot!


dev-client/src/store/logoutReducer.ts
knipec: Maybe put a comment about logout clearing the color picker data you mentioned in whichever redux slice file has it?

Issue #3048, opened 2025-09-02
Component: Soil Match Tile


DerekCaelin:
Create Component "Soil Match" according to Figma Spec

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7327-9259&t=gl54f0UOKyCBeDFV-0)
<img width="445" height="404" alt="Image" src="https://github.com/user-attachments/assets/0f17926e-c9a8-41e5-9134-2eb4cdca4f30" />

DerekCaelin:
@CourtneyLee333 and I will throw component-specific comments here

CourtneyLee333:
@knipec I'm adding feedback here to keep it separated from the main ticket which will focus on functionality. This all looks great to me, except one piece. When I select a soil, the Selected variant of the SoilMatchTile is showing in the list as well as at the top of the Soil ID screen. I expected the SoilMatchTile to only show the Selected variant at the top of the screen, while the one in the list shows the default style. (I think I have a vague memory of us talking about this, so forgive me if we made a different decision here.)

<img width="273" alt="Image" src="https://github.com/user-attachments/assets/fc286d9b-5574-4cd7-9b84-617e296a76e6" /> 

<img width="273" height="542" alt="Image" src="https://github.com/user-attachments/assets/3ce551c9-0653-470b-b3eb-0e33618cedb1" />

knipec:
@CourtneyLee333 Johannes pointed out the "Last calculated 20 minutes ago" on the screenshot on the right in your comment above. Just curious if there are plans for that upcoming or that's one of the "someday we'd like" features? :)

DerekCaelin:
Great. Let's not include "last updated" as a requirement now. I think we can close this issue once Carissa implements the "percent match instead of Selected Soil" change. For efficiency's sake, Carissa, if you share a screenshot with that experience and make sure it's in our release candidate I think we can call it complete without another round of QA.

PR #3169, opened 2025-12-10
fix: selected soil tile in match list should be Default variant, not Selected

knipec:
## Description
When a soil has been selected, it should still show its match percentage in the list of soil matches (which is the `Default` match tile variant), not show "Selected Soil" (which is the `Selected` match tile variant)

Fixes the issue in #3048 noted in [this comment](https://github.com/techmatters/terraso-mobile-client/issues/3048#issuecomment-3618423576)


knipec:
> btw, the screenshot in the problem referenced shows "last calculated" date for soil. Is that coming soon? I kind of like it but not sure how it would work.

Oh I hadn't even noticed that! I don't think there are plans for that work yet, but I'll check with Courtney

Issue #3090, opened 2024-09-27
Implement Language Selector


DerekCaelin:
Implement language selector in the settings menu. 

## Design
[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=10203-26039&t=xIob47dxnX9bGyFW-1)

<img width="590" height="506" alt="Image" src="https://github.com/user-attachments/assets/afa8ec55-94af-4eee-a205-1710caf45fd4" />

## Acceptance Criteria
- [ ] As a user, I want to be able to select what language I want the UI to be from the languages available to the user.

DerekCaelin:
@CourtneyLee333 and I spoke.

> 1. The selected language would be saved in local app storage, not as part of the user's account info, right?

local storage is fine.

> 2. We'd continue to default to the device's language, right? 

Yep!

> 3. Should there be an in-app way to clear the setting and just go back to using the device's language?

Preference would be to just have a picker with language options, not to display a "default".

> 4. The users are only choosing their language with the selector, not their locale, right?

Would like to discuss this at standup Tuesday!

PR #3170, opened 2025-12-11
feat: user can select language

knipec:
## Description
On UserSettingsScreen, user sees an item like "Language: English" that will open an affordance to allow the user to choose a language. Selecting a language will save the selection to KV storage and immediately update the app's language. 

https://github.com/user-attachments/assets/27b611ed-99a7-444f-afb8-bb3333059b91

### Related Issues
Implements https://github.com/techmatters/terraso-mobile-client/issues/3090


johannesparty (approved):
I don't know Spanish or Ukrainian, but I can confirm it is 2025 :).

PR #3175, opened 2025-12-15
feat: add hint to improve soil matches by entering site data

lesliehfrancis:
Shows an informational message on the soil matches section when no soil data has been entered, encouraging users to add observations to improve match accuracy. The message only appears for sites (not temporary locations) when there are no surface cracks and no depth-dependent soil properties recorded.

Fixes the empty soil data check to properly examine depthDependentData instead of depthIntervals, ensuring soil properties (texture, color, rock fragments) are correctly detected.

### Checklist
- [ x] Corresponding issue has been opened
- [ ] New tests added

### Related Issues
Fixes Bug 1126, New strings need to be localized

### DIFF
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "LandPKS Soil ID",
-  "version": "1.3.3",
+  "version": "1.4.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "LandPKS Soil ID",
-      "version": "1.3.3",
+      "version": "1.4.0",
       "hasInstallScript": true,
       "dependencies": {
         "@craftzdog/react-native-buffer": "^6.1.1",
@@ -78,7 +78,7 @@
         "react-redux": "^9.2.0",
         "reduce-reducers": "^1.0.4",
         "setimmediate": "^1.0.5",
-        "terraso-client-shared": "github:techmatters/terraso-client-shared#34c6d00b115d9ee04b4a73332fe7b4c5fc57e444",
+        "terraso-client-shared": "github:techmatters/terraso-client-shared#f05988f40dadebe1cb7ff622e301059301e0ad09",
         "use-debounce": "^10.0.6",
         "uuid": "^11.1.0",
         "yup": "^1.7.1"
@@ -118,7 +118,7 @@
         "lodash": "^4.17.21",
         "patch-package": "^8.0.1",
         "prettier": "^3.6.2",
-        "react-devtools": "^4.28.5",
+        "react-devtools": "^7.0.1",
         "react-native-svg-transformer": "^1.5.1",
         "react-test-renderer": "19.1.0",
         "ts-node": "^10.9.2",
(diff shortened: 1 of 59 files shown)


dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -83,13 +91,42 @@ export const SoilIdMatchesSection = ({
...
+    Array.isArray(soilData.depthDependentData) &&
+    soilData.depthDependentData.length > 0 &&
+    soilData.depthDependentData.some(
+      depthData =>
+        depthData.texture ||
+        depthData.colorHue !== undefined ||
+        depthData.colorValue !== undefined ||
+        depthData.colorChroma !== undefined ||
+        (depthData.rockFragmentVolume !== undefined &&
+          depthData.rockFragmentVolume !== null),
```
paulschreiber: question: would it make sense to use `.any?` or a similar function here?

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -83,13 +91,42 @@ export const SoilIdMatchesSection = ({
 
 type MatchTilesProps = SoilIdMatchesSectionProps & {soilIdOutput: SoilIdOutput};
 
+const isEmptySoilData = (soilData: SoilData): boolean => {
```
knipec: Might make a comment specifying why we care only about the existence of certain types of data (since if only slope data exists we'll still return true)

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -118,18 +155,47 @@ const MatchTiles = ({siteId, coords, soilIdOutput}: MatchTilesProps) => {
...
-          }
-          onPress={() => onMatchTilePress(soilMatch)}
-          variant={getTileVariant(soilMatch, userRatings, selectedSoilId)}
-        />
-      ));
+      const sortedMatches = getSortedMatches(soilIdOutput.matches);
+      return (
+        <>
+          {showImproveMessage && (
+            <Box
```
knipec: Optional suggestion: extract this Box and its children into its own component like `ImproveMatchesInfoBox` or something

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
```diff
@@ -118,18 +155,47 @@ const MatchTiles = ({siteId, coords, soilIdOutput}: MatchTilesProps) => {
...
+              backgroundColor="primary.50"
+              px="12px"
+              py="12px"
+              borderRadius="8px"
```
knipec: I don't see a border on the Figma design for this -- is this needed?

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
lesliehfrancis: Changed to check all depthDependant data.  Added translations to translation .json's.  Had to do merge with code from main

knipec (commented):
Feel free to ping me if you want to talk through anything!

dev-client/package-lock.json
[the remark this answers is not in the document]
lesliehfrancis: Should I remove package-lock.json?  Claude checked it in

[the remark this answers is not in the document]
lesliehfrancis: To make these changes, I need a lot more information on how to get all of the colors, etc

dev-client/src/screens/LocationScreens/components/soilId/SoilIdMatchesSection.tsx
knipec: This would be good to ask Derek and Courtney since we're not sure! 
I know some data that you can enter doesn't actually affect soil matches, but I don't confidently remember which data. And even if I did, I'm not sure if we want to reflect that via when we show the message.
I personally would probably lean towards checking all data; just keeps it less complicated imo. But still probably worth a checkin with the product folks!

PR #3183, opened 2025-12-23
feat: posthog screen recording v2

johannesparty:
- add screen recording  through posthog
- can enable/disable with posthog feature flag

### DIFF
--- a/dev-client/app.config.ts
+++ b/dev-client/app.config.ts
@@ -175,6 +175,7 @@ const defaultConfig: ExpoConfig = {
         ios: {},
         android: {
           gradleVersion: '8.14.3',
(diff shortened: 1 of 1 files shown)


knipec (commented):
Are there any automated tests that could be added here? That could be helpful, especially since it's kind of complicated code. Dunno how easy it is to mock what posthog gives, but might be nice to have automated tests for if polling happens when we do/don't expect

knipec (approved):
Re: the latest commit: I was a bit concerned in case one of the old Android users was someone we want to make sure to communicate with -- I don't know how to check that personally but Derek said he didn't see any recent PostHog events from devices running android 6, 7, or 8, so seems good to me!

PR #3184, opened 2025-12-23
fix: android map icons layer

johannesparty:
android map thumbnails would "bleed through" top when you push all your sites up in sites screen.
why?  probably because when mapbox draws the bitmaps, it does so outside react so some of the react clipping doesn't work properly. 
this hack gets temporary thumbnails and renders those in react so the layering works.
the thumbnails are just in memory (base64), so no cleanup needed.

note: this PR is a bit messy and mostly Claude-generated, just to see if it will work.  I might spend a little more time reviewing it myself.

### DIFF
--- a/dev-client/__tests__/snapshot/__snapshots__/SiteTabsScreen-test.tsx.snap
+++ b/dev-client/__tests__/snapshot/__snapshots__/SiteTabsScreen-test.tsx.snap
@@ -1282,19 +1282,31 @@ exports[`renders correctly 1`] = `
                                             "width": "100%",
                                           },
                                           {
+                                            "backgroundColor": "#e0e0e0",
                                             "overflow": "hidden",
                                           },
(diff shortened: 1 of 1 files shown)


dev-client/src/components/StaticMapView.tsx
```diff
@@ -116,27 +115,64 @@ export const StaticMapView = ({
...
+          style={styles.fill}
+          styleURL={Mapbox.StyleURL.Satellite}
+          scaleBarEnabled={false}
+          logoEnabled={false}
+          zoomEnabled={false}
+          scrollEnabled={false}
+          pitchEnabled={false}
+          rotateEnabled={false}
+          attributionEnabled={false}
+          pointerEvents="none"
```
knipec: Is this the same as before when pointerEvents wasn't passed in as a prop?

PR #3185, opened 2025-12-24
fix: android map thumbnails bleeding through scroll containers

johannesparty:
On Android, MapView components don't respect clipping in ScrollViews, causing map thumbnails to "bleed through" and appear on top of other content when scrolling.

Fix by capturing each MapView as a static image using takeSnap() and replacing it with an Image component. The MapView is rendered off-screen (left: -9999) while capturing to avoid visual artifacts.

(new 12/24): For Android specifically, implements a queue system to load MapViews sequentially since multiple simultaneous off-screen MapViews don't reliably fire their callbacks.

### DIFF
--- a/dev-client/src/components/StaticMapView.tsx
+++ b/dev-client/src/components/StaticMapView.tsx
@@ -15,8 +15,15 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import {useCallback, useMemo, useRef, useState} from 'react';
-import {Image, StyleProp, StyleSheet, View, ViewStyle} from 'react-native';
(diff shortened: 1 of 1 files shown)


dev-client/src/components/StaticMapView.tsx
```diff
@@ -30,6 +37,44 @@ import {
...
+      continue;
+    }
+
+    isProcessingQueue = true;
+    next.callback();
+    return;
+  }
+};
+
+const markMapLoadComplete = () => {
```
knipec: What does this do? If a StaticMapView was unmounted but it was its turn, it calls this. Does that mean it is assumed to be the last one in the queue, or just that ??

PR #3189, opened 2025-12-30
fix: sanitize filenames and URLs in data export

johannesparty:
- Add sanitizeFilename for local file saves (names with / were failing amount others).
- Unify sanitization logic in utils/sanitize.ts, so mostly use the same sanitization for export and local file save.

### DIFF
--- a/dev-client/src/utils/sanitize.ts
+++ b/dev-client/src/utils/sanitize.ts
@@ -1,0 +1,72 @@
+/*
+ * Copyright © 2025 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+/**
+ * Core sanitization for names used in filenames or URL paths.
+ *
+ * Handles:
+ * - Unicode normalization (NFC)
+ * - Control characters removed
+ * - Filesystem-invalid chars (/\:*?"<>|) → hyphens
+ * - URL-problematic chars (#&=%()') removed
+ * - Multiple hyphens collapsed
+ * - Leading/trailing dots and hyphens trimmed
+ */
+const sanitizeNameCore = (name: string): string => {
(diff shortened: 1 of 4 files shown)


dev-client/src/utils/sanitize.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Sanitizes a string for use in a URL path segment.
+ * Converts spaces to hyphens for cleaner URLs.
+ *
+ * @example
+ * sanitizeNameForUrl("Project: Phase 1")   // "Project-Phase-1"
+ * sanitizeNameForUrl("Tested 12/29/2025")  // "Tested-12-29-2025"
+ * sanitizeNameForUrl("Київ Site")          // "Київ-Site"
+ * sanitizeNameForUrl("What's Next?")       // "Whats-Next"
+ */
+export const sanitizeNameForUrl = (name: string): string => {
```
knipec: Sorry if I've asked this before, but is it an issue that there may be duplicate names (even for the same user)? If you export a site with a duplicate name as another site, will it just pick one? If so, probably fine for now, but an edge case we probably want to consider?

dev-client/src/utils/sanitize.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Preserves spaces for readability.
+ *
+ * @example
+ * sanitizeFilename("Project: Phase 1")     // "Project- Phase 1"
+ * sanitizeFilename("Tested 12/29/2025")    // "Tested 12-29-2025"
+ * sanitizeFilename("What's Next?")         // "Whats Next"
+ * sanitizeFilename("Файл (test)")          // "Файл test"
+ */
+export const sanitizeFilename = (name: string): string => {
+  const MAX_LENGTH = 100; // Conservative limit; reserves space for extension
```
knipec: Just curious how this number was chosen and what the comment means?

dev-client/src/utils/sanitize.ts
johannesparty: fine, there is a unique token in the url too.

dev-client/src/utils/sanitize.ts
knipec: Answer: Turns out file names are 255 bytes max for platforms. If unicode, could be fewer characters. 

But mainly, looking at constants.ts, we currently have `SITE_NAME_MAX_LENGTH = 50` and `PROJECT_NAME_MAX_LENGTH = 50`, so anything greater than 50 should be entirely safe unless we change one of those. 

So I suppose we could use like `min(max(SITE_NAME_MAX_LENGTH, PROJECT_NAME_MAX_LENGTH),100)` but I'm not pressed, I doubt those constants will change much 🤷

dev-client/src/utils/sanitize.ts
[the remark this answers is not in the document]
knipec: Answer: Pretty confident. Johannes looked up the actual rules, and this is doing more sanitization than necessary, which should be fine.

PR #3190, opened 2025-12-30
fix: improved export error dialog

johannesparty:
- Use the existing ErrorDialog for export error messages (confirms to established look and feel)
- Some wording changes for better consistency
-


dev-client/src/screens/DataExportScreen/DataExportScreen.tsx
```diff
@@ -77,8 +79,20 @@ export function DataExportScreen({
...
   const infoSheetRef = useRef<ModalHandle>(null);
+  const errorDialogRef = useRef<ModalHandle>(null);
 
   const [isDownloading, setIsDownloading] = useState(false);
+  const [errorType, setErrorType] = useState<
+    'save' | 'create_link' | 'reset_link'
+  >('save');
+
+  const showError = useCallback(
+    (type: 'save' | 'create_link' | 'reset_link') => {
```
knipec: Nit: could make a named type for this, like ExportErrorType or something, since we use it in more than one place

dev-client/src/screens/DataExportScreen/DataExportScreen.tsx
```diff
@@ -215,33 +230,23 @@ export function DataExportScreen({
         }
 
         // Handle failure cases - destructure to help TypeScript
-        const {canceled, error} = result as {
+        const {canceled} = result as {
           success: false;
           error: string;
```
knipec: Would it be useful to log the specific error? Or will the backend log it?

knipec (approved):

dev-client/src/screens/DataExportScreen/DataExportScreen.tsx
johannesparty: good idea, but not changing now.

PR #3191, opened 2025-12-30
fix(android): block READ_MEDIA_* permissions for Google Play compliance

johannesparty:
- Google Play rejected the app for declaring READ_MEDIA_IMAGES and READ_MEDIA_VIDEO . 
- The app uses Android Photo Picker instead, which doesn't require these permissions on Android 13+.


johannesparty:
> I'm curious: Were those permissions always being requested, and it’s just now that the play store stopped us from requesting them? Or were they newly requested with some change?

yes they were "always" required (well, at least not new requirements).

PR #3192, opened 2025-12-30
Fix: go to version 1.4.1 to satisfy stores

johannesparty:
- changed version to 1.4.1 because google and apple reject 1.4.0 now.
- google and apple both same or lower version after submitting to App Stores, even if rejected (android) or never released (iOS).

### DIFF
--- a/.github/workflows/build-shared-android.yml
+++ b/.github/workflows/build-shared-android.yml
@@ -119,5 +119,5 @@ jobs:
           releaseFiles: ${{ steps.sign_app.outputs.signedFile }}
           track: internal
           status: draft
-          # changesNotSentForReview: true
+          changesNotSentForReview: true
           inAppUpdatePriority: 2


.github/workflows/build-shared-android.yml
```diff
@@ -119,5 +119,5 @@ jobs:
           releaseFiles: ${{ steps.sign_app.outputs.signedFile }}
           track: internal
           status: draft
-          # changesNotSentForReview: true
+          changesNotSentForReview: true
```
knipec: Did this being commented out mean that previously we were automatically sending all changes for google play store review? (I see that change [here](https://github.com/techmatters/terraso-mobile-client/commit/976d7591f2257e397f70fc07a1b0441afe6e3b94) but can't tell why it was done.)

knipec (commented):
Curious about the changesNotSentForReview part, but everything else looks good!

knipec (approved):

PR #3193, opened 2025-12-30
fix: revert mapbox code and try new approach

johannesparty:
- undid all the recent mapbox stuff including
   - d89dddb1 fix: android map thumbnails bleeding through scroll containers
   - 671edec2 fix: android map icons layer
- finally found a fix and it is really simple, essentially one line
   - added  surfaceView=false for android

the first commit is the "undo", the second commit is adding the new setting for Android.
(it probably wouldn't hurt for iOS either, but wasn't necessary and didn't want to make unnecessary changes to iOS).

### DIFF
--- a/dev-client/src/components/StaticMapView.tsx
+++ b/dev-client/src/components/StaticMapView.tsx
@@ -16,7 +16,7 @@
  */
 
 import {useMemo} from 'react';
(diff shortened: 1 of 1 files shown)


knipec (approved):
Ooo I'm glad you figured this out! Much much cleaner 😄

PR #3197, opened 2025-12-31
fix: android permissions retry

johannesparty:
- changed android permissions so play store will accept
- updated version to 1.4.2 since 1.4.1 is now taken.


knipec (commented):
I'm confused because as far as I can tell, nothing changed about the READ_MEDIA_IMAGES or READ_MEDIA_VIDEO permissions, which are the ones that the error mentions.

PR #3205, opened 2026-01-07
feat: add Copy JWT Access Token option in settings (dev only)

### LINKED ISSUE #3204: add ability to load JWT token for debugging graphql
Add a button in user settings copy the token to the clipboard, like this:

Image

This should only appear in non-production builds.

Once you have the button, you can use with utilities such as Postman.

See example below: put in the URL for your server, then on query you can click to build a query in the "Query" tab (I did allExportTokens here).  

Image

Finally, click on authorization and paste in the copied JWT access token:

Image

Now when you hit the "Query" button, you'll get results for the user associated with the JWT.

You can also paste the JWT into "jwt.io" website.  Easiest to understand as "CLAIMS TABLE" on the right.  

Finally, to validate the signing, paste in the "SECRET" from the .env file's JWT_SECRET value (which could be improved).  If testing against staging/production, you can find the JWT secret from render.

Image

Yes I've exposed my JWT here, but it is a local one for my local server and it expires, so I'm not worried.

johannesparty:
Add a menu item in settings that copies the current user's JWT access token to the clipboard. Only visible in non-production environments to help with debugging and API testing.

### Related Issues
Fixes #3204

### Verification steps
Look at issue 3204 and you can follow along.
Also, should confirm doesn't show up in production.

### DIFF
--- a/dev-client/src/screens/UserSettingsScreen/components/menu/CopyAccessTokenItem.tsx
+++ b/dev-client/src/screens/UserSettingsScreen/components/menu/CopyAccessTokenItem.tsx
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2026 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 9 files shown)


dev-client/src/screens/UserSettingsScreen/components/menu/CopyAccessTokenItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const {t} = useTranslation();
+
+  const handleCopyToken = useCallback(async () => {
+    const token = await getAPIConfig().tokenStorage.getToken('atoken');
+    if (token) {
+      await Clipboard.setStringAsync(token);
+    }
+  }, []);
+
+  if (APP_CONFIG.environment === 'production') {
```
knipec: Just confirming that we want this available in staging, not just local dev builds?

Issue #3200, opened 2026-01-02
Sites screen: search bar: user can't complete typing coordinates after first valid lat/long entered.


DerekCaelin:
### App Version

322

### Platforms

- [ ] iOS
- [x] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

When typing coordinates in search bar, at first valid lat/long:
- search bar disabled

### Steps to Reproduce

On sites tab:
1. tap on search bar, type in lat long, e.g. `20.4, 30.5235234`
2. See issue

### Expected behavior

user can continue to type lat long into search bar until coordinates are complete

### Actual behavior

At the first valid lat long `20.4, 3` search bar disappears and you can't finish typing

Issue #2017, opened 2024-08-21
Map search with coordinates should use search result functionality


CourtneyLee333:

knipec:
Triage notes: 
- Does mapbox search have this logic, or do we need to implement it ourselves?
- Seems like a medium priority feature request -- @derek can you attach this to the ["polishing up the app"](https://github.com/techmatters/terraso-product/issues/346) work?
- Ruxandra was in this code recently

PR #3216, opened 2026-01-22
fix: map search shows suggestion for coordinates

knipec:
## Description
The commits should be pretty reasonable to look through if you want to review that way. 

### Changes
- Typing valid lat, long into map search should now show a suggestion for those coordinates instead of zooming the map directly there and opening a callout. This makes the debounce unnecessary.
- Should be able to zoom to a lat, long when offline
- Aside: latitude +/- 90 should be valid for coordinates
- Restructure to reduce complexity
- Add some tests

### Related Issues
Fixes https://github.com/techmatters/terraso-mobile-client/issues/3200 
By implementing https://github.com/techmatters/terraso-mobile-client/issues/2017 

### Verification steps
- Type a valid lat,long in map search
- Confirm clicking the suggestion moves you to that lat,long
- Confirm that regular mapbox suggestions also continue to work
- Confirm search still disappears when the callout is open

### DIFF
--- a/dev-client/src/util.ts
+++ b/dev-client/src/util.ts
@@ -128,6 +128,10 @@ export const sortCompare = (
 };
 
 export const isValidCoordinates = (input: string) => {
+  // Valid strings are of the form <sign><latitude>,<sign><longitude>, where:
+  //  <sign> is optional, can be + or -
+  //  <latitude> can be an integer or decimal from -90 to 90 (inclusive)
+  //  <longitude> can be an integer or decimal from -180 to 180 (inclusive)
   const coordRegex =
     /^([-+]?90(?:\.0+)?|[-+]?[1-8]?\d(?:\.\d+)?),\s*([-+]?180(?:\.0+)?|[-+]?((1[0-7]\d)|([1-9]?\d))(?:\.\d+)?)$/;
 


dev-client/src/util.ts
```diff
@@ -129,7 +129,7 @@ export const sortCompare = (
 
 export const isValidCoordinates = (input: string) => {
   const coordRegex =
-    /^([-+]?[1-8]?\d(?:\.\d+)?),\s*([-+]?180(?:\.0+)?|[-+]?((1[0-7]\d)|([1-9]?\d))(?:\.\d+)?)$/;
+    /^([-+]?90(?:\.0+)?|[-+]?[1-8]?\d(?:\.\d+)?),\s*([-+]?180(?:\.0+)?|[-+]?((1[0-7]\d)|([1-9]?\d))(?:\.\d+)?)$/;
```
johannesparty: I like using regex for this, but hard to read in this form.  might be worth adding a comment explaining what this captures.  Also just curious what you used to write the regex.  I used to use online regex tools and hand-edit and check with examples you could put into the tool.  But now ... claude?

dev-client/src/util.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      expect(isValidCoordinates('12.345,67.890')).toBe(true);
+    });
+
+    it('accepts negative coordinates', () => {
+      expect(isValidCoordinates('-12.345, -67.890')).toBe(true);
+      expect(isValidCoordinates('-12.345, 67.890')).toBe(true);
+      expect(isValidCoordinates('12.345, -67.890')).toBe(true);
+    });
+
+    it('accepts boundary values', () => {
```
johannesparty: Looks like these are the cases where you fixed the regex :)
This is almost documentation.  But could add the summary of what should/shouldn't work here too I guess.

johannesparty (approved):
Looks good, lots of tests too :).
A couple of behaviors:
- in simulator, if you type coordinate and hit enter, nothing happens, you have to click on the lat,long suggestion.  not important because doesn't really matter for real phones unless some is using an external keyboard (rare?).
- I think the algorithm adds lat,long suggestion if the input has the syntax for it (regex), and then adds the mapbox suggestions.  when given coordinates, do we ever get mapbox suggestions?  if not, we could (a) skip 2nd part, and (b) make it so any num,num is handled specially, but shows an error saying must be within xyz range.  again, not a problem, just a thought.
- I'm not sure it is obvious you can type in coordinates, seems like a help text below the input box would be nice, but I guess that's more a question for Courtney.

Issue #3202, opened 2026-01-06
Why can't I simulate offline mode with Little Snitch anymore?


knipec:
See Terraso development guide steps for "Offline debugging"

PR #3220, opened 2026-01-29
docs: add info about simulating offline mode with Little Snitch tool

knipec:
## Description
This allows developers to use the "Little Snitch" tool to simulate offline mode in the app, but not lose connection to Metro (so the app still logs to the console, and you can use debug tools).

I added info under ["Offline debugging" in the Terraso Development Guide](https://docs.google.com/document/d/1IEy2LX1_ZSVYTZkAaL9P0IePm2huslViU7BC89KY_Lw/edit?tab=t.0#heading=h.takbf32jwxwi)

Honestly I don’t understand why this makes a difference, because allegedly the url I have in the code is the same as how Android OS checks connectivity natively?? But based on my manual testing it does seem to make a difference! Otherwise changing the profile won't cause the app to recognize it should be in offline mode, it just fails to connect to a bunch of stuff.

@johannesparty There's a lot of context here, so let me know when you get around to this PR and I can talk thru some of the context!

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/3202 as best as I can right now

### DIFF
--- a/dev-client/src/App.tsx
+++ b/dev-client/src/App.tsx
@@ -82,6 +82,18 @@ if (persistedReduxState) {
 }
 const store = createStore(persistedReduxState);
 
+/* Developer FYI: To enable the Little Snitch tool to simulate offline mode without losing connection to Metro:
+ * - Uncomment the code below
+ * - import NetInfo from '@react-native-community/netinfo'
+ * - See "Offline Debugging" section in Terraso Development Guide for more info
+ */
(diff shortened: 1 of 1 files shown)


johannesparty (approved):
It would be nice to simplify the explain the individual rules a little, but not sure it is worth your time to do so now; current rules work, so maybe good enough!

PR #3226, opened 2026-02-03
fix: the FF_testing flag will always be off in prod

knipec:
## Description
Even if it was somehow on in kvstorage, `FF_testing` feature flag will always be off in prod. 

### Related Issues
Not sure how it happened, but this is because derek [reported](https://tech-matters.slack.com/archives/C06BQTCAUJF/p1770065160193659) seeing UI behind that flag in a prod build

### DIFF
--- a/dev-client/src/config/featureFlags.test.ts
+++ b/dev-client/src/config/featureFlags.test.ts
@@ -73,7 +73,7 @@ test('feature flags can be disabled', () => {
   });
 });
 
-test('testing feature flag is disabled in production', () => {
+test('testing feature flag is default disabled in production', () => {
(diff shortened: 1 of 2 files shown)


knipec:
> Was it getting turned on in prod by accident sometimes?
> 
> Are you sure we don't want a secret way to turn it on in prod? Seems like that might be nice. (could selectively enable with posthog FF or with cloudflare FF). But that's a discussion for another day.

Oo the posthog FF is a good idea, though agree it's for another day! I'm honestly not sure how Derek ended up seeing it in prod (the feature flag being weirdly "on" in app storage is just the only real hypothesis I have).

Issue #3211, opened 2026-01-15
Sites screen: Missing header for empty site list bottom sheet


DerekCaelin:
### App Version

322

### Platforms

- [x] iOS
- [x] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

There should be a header for the bottom sheet when the user has no sites, to provide them with guidance on how to get started with LandPKS. 

### Steps to Reproduce

1. Have no sites
2. Go to "Sites screen"
3. Look at bottom sheet
4. See issue

### Expected behavior

Bottom sheet header says `+ Create Site`

[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=11024-4381&t=EhIFEUrCOhUrvhCG-0)
<img width="300" src="https://github.com/user-attachments/assets/e5c64806-a99f-44c8-becb-c036aa9e9567" />

### Actual behavior

No header is visible

<img width="300"  src="https://github.com/user-attachments/assets/1d5dd59d-efd2-4a66-8926-cad6cce54d41" />

### Additional context

This is a regression. Important because usability tests demonstrated that users needed help getting started.

DerekCaelin:
Medium: we found in usability testing that people needed help figuring how to create a site.

knipec:
Should be available for QA in the next staging build (likely v338beta)

PR #3233, opened 2026-02-11
fix: show Create Sites header when site list is empty

knipec:
## Description
Looks like a recent-ish change made some code conditional on `isEmpty`, but only showed the `ListHeaderComponent` when the site list was not empty. We should still show the header when the site list is empty.

Aside: The code style of having `ListHeaderComponent` as a react element variable instead of a component with JSX feels kind of weird, but should be fine. I suppose it makes it cleaner in the places it's used, since you don't have to pass props.

### Related issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/3211


johannesparty (approved):
A bit strange to have ListHeaderComponent in both places, but makes sense, I don't see a problem, easy to understand.

PR #3234, opened 2026-02-13
fix: delete obsolete work-tracking md files

johannesparty:
Delete a few obsolete files.


knipec (approved):
Out of curiosity: Were these all completed, or just not something we intend to follow up on?

Issue #2198, opened 2024-09-25
Offline: Create/Edit Notes


DerekCaelin:

DerekCaelin:
Issues addressed or in backlog. closing task

Issue #3221, opened 2026-01-30
Project conflict dialog shown when deleting a site


knipec:
## Repro steps:
1. Go to a site's settings screen
2. Delete it
Expected: App navigates to the sites home page, site is deleted, shows no error
Actual: App navigates to the sites home page (correct), site is deleted (correct), shows "Project conflict" error dialog: 
<img width="361" height="311" alt="Image" src="https://github.com/user-attachments/assets/86cab828-12ac-4880-875a-f462fdc48a42" />
 

## Technical notes: 
- useNavToBottomTabsAndShowSyncError is being called due to missing site, from `SiteTabsScreen`, a parent of SiteSettingsScreen.
- Order of operations today:
 LOG  SiteSettingsScreen onDelete
 LOG  render SiteSettingsScreen
 LOG  deleteSite.fulfilled
 LOG  render SiteTabsScreen
 LOG  render SiteSettingsScreen
 LOG  SiteTabsScreen handleMissingSite
 LOG  SiteSettingsScreen handleMissingSite
 LOG  SiteSettingsScreen deleteSite dispatch done awaiting in onDelete
- If we just navigate before deleting, the new screen may display the deleted site before it gets deleted (because site deletion currently requires a round trip to the server). See notes from https://github.com/techmatters/terraso-mobile-client/pull/2839 

### Possible solutions: 
- Instead of sitePurposelyDeleted as a state variable within SiteSettingsScreen, make the site slice know which sites are being deleted. 
  - Add a deletingSiteId field to the site slice. The flow would be:
    - SiteSettingsScreen.onDelete dispatches an action to set deletingSiteId = siteId
    - Dispatch deleteSite (which clears deletingSiteId when done)
    - SiteTabsScreen checks: if siteId === deletingSiteId, call navToBottomTabs() without error instead of navToBottomTabsAndShowSyncError()
- Delete site from local redux and navigate after that but before server deletion completes
- Implement offline site deletion. Then we will necessarily be deleting the site locally.

DerekCaelin:
I can reproduce this.

knipec:
Once the offline sites branch gets merged to main, this issue will likely be resolved. The implementation for the fix feels a bit weird, but I haven't come up with any issues it causes in user-facing behavior.

DerekCaelin:
Derek to test solution in new build.

DerekCaelin:
Just tested on build 350, this issue is resolved!

PR #3265, opened 2026-04-08
feat: create sites offline (mobile-client)

knipec:
## Description
Enable:
- Creating and updating sites offline
- Creating, updating, and deleting site notes offline

Also: 
- Alter some project-related UI when offline
- Handle fetching elevation [as detailed here](https://app.mural.co/t/techmatters9109/m/techmatters9109/1772136926429/9b7af7501a416e158fb3047cfbcf2da3e1b2f539)
- If user edits a deleted site note, we silently just discard the edits and the note remains deleted.
- We decided not to do anything special with map tiles, so they will only load when viewed online.
- Add ENV variable `SYNC_DEBUG_ENABLED` for sync-specific logging
- Change PostHog events: instead of `sync_conflict`, there is now `missing_data_sync_error`, `site_push_failure`, `soil_data_push_failure`, and `soil_metadata_push_failure` (the push_failure ones now contain site name and failure reason).

### Related Issues
Implements https://github.com/techmatters/terraso-product/issues/1144 and https://github.com/techmatters/terraso-mobile-client/issues/2198


dev-client/package.json
knipec, line 1: Reminder to update this and package-lock once the client-shared PR is merged

dev-client/src/model/sync/actions/syncActions.test.ts
```diff
@@ -161,7 +161,7 @@ describe('pushUserData', () => {
         },
       ],
       soilMetadataResults: null,
-      clientMutationId: null,
+      siteResults: null,
```
knipec, line 162: Was it ok to remove the clientMutationId? I'm not clear on what that was used for --> Supposedly this is fine. 

Claude: 
> clientMutationId is a Relay GraphQL convention. Here's how it works:

> The mechanism: The client sends a clientMutationId in the mutation input, and the server echoes the exact same value back in the response. That's all it does — pure echo, no server-side logic.

> What it's for: It was designed to solve a problem with mutation request/response correlation in environments where multiple mutations might be in flight simultaneously. If you fire off 3 mutations at once, the clientMutationId lets you match each response to the request that triggered it. Think of it like a request tracking ID chosen by the caller.

> Why it's mostly obsolete: Modern GraphQL clients (Apollo, urql, etc.) handle request correlation at the transport layer — each HTTP request/response or WebSocket message is already paired. You don't need an application-level correlation ID. The Relay spec originally required it, but even Relay made it optional years ago.

> Why it's still in the other types: It's likely there because the backend uses a framework (like Graphene-Django with Relay) that automatically adds it to every mutation payload type. 

> For your branch specifically: The pushUserData mutation response type was presumably defined by hand (or refactored) as part of the sync work, and whoever did it simply didn't include the vestigial clientMutationId field. That's the right call — there's no reason to add it to new types, and the mobile client was already setting it to null in every mock, confirming it was never used.

dev-client/src/components/dataRequirements/handleMissingData.ts
```diff
@@ -20,23 +20,29 @@ import {useCallback} from 'react';
...
 export const useNavToBottomTabsAndShowSyncError = (
   missingEntityType?: string,
 ) => {
   const navigation = useNavigation();
   const syncNotifications = useSyncNotificationContext();
+  const siteDeletedByUser = useSelector(selectSiteDeletedByUser);
 
   return useCallback(() => {
     navigation.popTo('BOTTOM_TABS');
+    if (missingEntityType === 'site' && siteDeletedByUser) {
```
knipec, line 31: I think we'd want this in the other hooks in this file as well (and if there are any other places we show the sync error based on missing data). Or we'd want better code design so we don't need to add it to multiple places?

Though, note: It was only called in the SiteSettingsScreen before, so maybe this is fine? Maybe just put a comment or something like "this is only in the missing data handler that can be called from the screen where the user can delete a site"

dev-client/src/model/soilData/soilDataSlice.ts
```diff
@@ -107,6 +107,7 @@ export const setSoilData = (
 export const deleteSoilData = (state: Draft<SoilState>, siteIds: string[]) => {
   for (const siteId of siteIds) {
     delete state.soilData[siteId];
+    delete state.soilSync[siteId];
```
knipec, line 110: Good catch!

dev-client/src/model/site/siteSlice.ts
```diff
@@ -37,6 +37,17 @@ import {
...
+  // pull. This suppresses the "sync conflict" error dialog that would otherwise
+  // fire when ScreenDataRequirements detects the site is missing — on a
+  // user-initiated delete, silently navigating back is the right behavior.
+  //
+  // We don't track which site was deleted, so in theory: if the user deletes
+  // site A, then navigates to site B, and a pull removes site B before the
+  // flag is cleared, the error dialog for site B would be incorrectly
+  // suppressed. This is near-impossible in practice because the pull itself
+  // clears the flag.
+  siteDeletedByUser: false,
```
knipec, line 53: I think this is an improvement and it'll probably work most of the time, but the design here feels weird. 

Let's say the user is offline, deletes site A, navigates to site B, makes some changes, and returns online while still viewing site B. Someone has since deleted site B. When the user returns online, the push for site B should result in the dialog, but it doesn't. 

Not sure if it's worth fixing right now, but I'll update https://github.com/techmatters/terraso-mobile-client/issues/3221 to capture it anyway.

knipec: Oh right, you can't delete a site offline. While I do still feel the code design is unintuitive, I can't currently think of a user-facing issue this would cause.

dev-client/__tests__/integration/store/sync/PushDispatcher.test.tsx
knipec: Anything to add here?

dev-client/src/model/elevation/elevationService.ts
```diff
@@ -42,3 +42,32 @@ export const getElevation = async (
...
   return elevation;
 };
+
+const ELEVATION_FETCH_TIMEOUT_MS = 10000; // 10 seconds
+
+/**
+ * Fetch elevation for coordinates. For use outside React components.
+ * Returns undefined if fetch fails or times out (non-blocking).
+ */
+export const fetchElevationForCoords = async (
```
knipec: What's the difference between this and getElevation? Should they be merged or named better?

dev-client/src/model/site/actions/localSiteActions.ts
knipec, line 1: Can you delete a site note offline? Why wouldn't that be here? --> The siteSlice doesn't need any logic to happen in the action; it's all just in the reducer.

dev-client/src/model/site/actions/localSiteActions.ts
```diff
@@ -0,0 +1,94 @@
...
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+import {v4 as uuidv4} from 'uuid';
+
+import {
+  SiteAddMutationInput,
+  SiteNoteAddMutationInput,
+  SiteNoteUpdateMutationInput,
+  SiteUpdateMutationInput,
```
knipec: Seems kinda not preferred to use the types from now-deprecated mutations. But we could push that off

dev-client/src/model/site/actions/siteActions.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (isFlagEnabled('FF_offline')) {
+    if (syncDebugEnabled) {
+      console.log(
+        '🗑️ deleteSiteNoteAction (local)',
+        input.id,
+        'for site',
+        input.siteId,
+      );
+    }
+    return Promise.resolve(input);
```
knipec, line 133: Huh? How does this actually delete the site note? --> Oh, because it's all done in the reducer, no logic is needed before then

dev-client/src/model/site/siteGlobalReducer.ts
```diff
@@ -31,12 +32,23 @@ import {
...
+  'site/addSite',
+  siteActions.addSiteAction,
+);
 
 export const updateSite = createAsyncThunk(
   'site/updateSite',
+  siteActions.updateSiteAction,
+);
+
+// Used for update actions that are not yet supported offline
```
knipec, line 49: FYI: Currently I believe this only includes removing a site from a project

dev-client/src/model/site/siteSlice.ts
```diff
@@ -44,24 +69,52 @@ export const fetchSitesForUser = createAsyncThunk(
...
 );
 
 export const setSites = (
   state: Draft<SiteState>,
   sites: Record<string, Site>,
 ) => {
-  state.sites = sites;
+  const badIncoming = Object.values(sites).filter(
+    s => typeof s.latitude !== 'number' || typeof s.longitude !== 'number',
+  );
```
knipec: Curious if this can actually happen? Or did Claude just add it as a defensive safeguard? --> Claude added as defensive safeguard; let's delete

dev-client/src/model/site/siteSlice.ts
```diff
@@ -81,13 +134,32 @@ export const updateProjectOfSite = (
 export const deleteSites = (state: Draft<SiteState>, siteIds: string[]) => {
   for (const siteId of siteIds) {
     delete state.sites[siteId];
+    delete state.siteSync[siteId];
```
knipec, line 119: Is there a logic to which actions set the siteSync in the action vs. in the reducer? --> Oh, this action is only used in the siteGlobalReducer. Generally I think actions shouldn't modify state.

dev-client/src/model/soilData/actions/remoteSoilDataActions.ts
```diff
@@ -98,7 +95,7 @@ export const unsyncedDataToMutationInputEntry = (
 
 export const soilDataMutationResponseToResults = (
   unsyncedChanges: SyncRecords<SoilData, unknown>,
-  response: SoilDataPushEntry[],
+  response: Array<{siteId: string; result: SoilDataPushEntryResultFragment}>,
```
knipec, line 98: Can't really remember why this changed --> Ah, to use the actual shape from the GraphQL fragment and avoid the typecast in syncActions.ts: `response.soilDataResults as SoilDataPushEntry[],` because we had been kind of lying about the type previously

dev-client/src/model/sync/results.ts
```diff
@@ -45,6 +46,40 @@ export const applySyncResults = <D, E>(
   const upToDateData = getValuesForCurrentRevisions(records, results.data);
   const upToDateErrors = getValuesForCurrentRevisions(records, results.errors);
 
+  const staleDataIds = Object.keys(results.data).filter(
+    id => !(id in upToDateData),
+  );
+  const staleErrorIds = Object.keys(results.errors).filter(
+    id => !(id in upToDateErrors),
+  );
+  if (staleDataIds.length > 0 && syncDebugEnabled) {
```
knipec: Code cleanliness: Should these log sections be extracted to a different function?

dev-client/src/model/sync/syncGlobalReducer.ts
```diff
@@ -58,18 +78,65 @@ export const syncGlobalReducer = createGlobalReducer(builder => {
...
+    });
+    if (syncDebugEnabled) {
+      console.log('━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━');
+      console.log(
+        `- ${sitesNeedingElevation.length} sites still need elevation after pull:`,
+      );
+      sitesNeedingElevation.forEach(siteToPrint => {
+        console.log(payload.sites[siteToPrint].name);
+      });
+    }
```
knipec: The log statements really clutter things up. Should we keep them, remove them, or extract them into functions?

dev-client/src/model/sync/syncGlobalReducer.ts
```diff
@@ -58,18 +78,65 @@ export const syncGlobalReducer = createGlobalReducer(builder => {
     setExportTokens(state.export, payload.exportTokens);
     updateSoilIdStatus(state.soilData, 'ready');
     setLastPullTimestamp(state.devOnly, Date.now());
+
+    // Mark sites missing elevation as needing sync so their elevation
+    // is fetched and pushed on the next push.
```
knipec, line 87: This feels a bit roundabout. We could instead fetch elevation immediately, and then mark things as synced, but I don't think it really makes much of a difference 🤷. Opinions welcome!

dev-client/src/screens/SitesScreen/components/TemporaryLocationCallout.tsx
```diff
@@ -152,8 +154,7 @@ type ElevationDisplayProps = {
 
 const ElevationDisplay = ({elevation, t}: ElevationDisplayProps) => {
   const isOffline = useIsOffline();
-
-  if (isOffline) {
+  if (isOffline && elevation.value === undefined) {
```
knipec: Can we be consistent about elevation being `null` vs `undefined`? I think we only need one

dev-client/src/screens/SitesScreen/utils/siteFeatureCollection.ts
```diff
@@ -19,16 +19,39 @@ import {Site} from 'terraso-client-shared/site/siteTypes';
 
 export const siteFeatureCollection = (
   sites: Pick<Site, 'id' | 'latitude' | 'longitude'>[],
-): GeoJSON.FeatureCollection<GeoJSON.Point> => ({
-  type: 'FeatureCollection',
-  // TODO: Filter sometimes returns undefined; figure out why and fix
```
knipec, line 24: Is this TODO comment why we have the badSites code?

dev-client/src/model/site/actions/localSiteActions.ts
johannesparty: I vote for fixing now.

claude analysis:


all four of those imported types correspond to deprecated mutations:                                                                                                                                                                  
                                                                                                                                                                                                                                              
  ┌─────────────────────────────┬────────────────────────┬──────────────────┐                                                                                                                                                                 
  │            Type             │ Corresponding mutation │   Deprecated?    │                                                                                                                                                                 
  ├─────────────────────────────┼────────────────────────┼──────────────────┤                                                                                                                                                                 
  │ SiteAddMutationInput        │ addSite                │ Yes ✓            │                                                                                                                                                                 
  ├─────────────────────────────┼────────────────────────┼──────────────────┤                                                                                                                                                                 
  │ SiteUpdateMutationInput     │ updateSite             │ Yes ✓ (just now) │
  ├─────────────────────────────┼────────────────────────┼──────────────────┤                                                                                                                                                                 
  │ SiteNoteAddMutationInput    │ addSiteNote            │ Yes ✓            │
  ├─────────────────────────────┼────────────────────────┼──────────────────┤                                                                                                                                                                 
  │ SiteNoteUpdateMutationInput │ updateSiteNote         │ Yes ✓            │
  └─────────────────────────────┴────────────────────────┴──────────────────┘                                                                                                                                                                 
   
  However, this file isn't actually calling the deprecated mutations — it's reusing their input types as the shape for local Redux state operations (creating/updating sites and notes locally before they get pushed via pushUserData). So   
  the code works fine, it's just depending on types whose only schema-level consumer is deprecated.
                                                                                                                                                                                                                                              
  Two options:                                                         

  1. Leave as-is — the types still exist in the schema (deprecation doesn't remove them), and they happen to be the right shape for what this file needs. Pragmatic.                                                                          
  2. Migrate to push input types — use SitePushInputEntry and SitePushNoteInput from the generated GraphQL types instead. Cleaner dependency chain, but the shapes are slightly different (push types use siteId instead of id,
  newNotes/updatedNotes/deletedNoteIds instead of separate mutation inputs, etc.), so it's not a drop-in replacement.                                                                                                                         
                                                                       
  If the plan is to eventually remove the deprecated mutations entirely (not just mark them deprecated), then migrating off these types now would avoid a future breakage. But if they'll stay around for a while, option 1 is fine.          
                                                                       
  Want me to look at whether these types are used elsewhere in the mobile client, to gauge the scope of a migration?                                                                                                                          
                                                                       
❯ yes                                                                                                                                                                                                                                         
                                                                       
  Searched for 1 pattern (ctrl+o to expand)                                                                                                                                                                                                   
                                                                                                        
⏺ They're used in 6 files across the mobile client:                                                                                                                                                                                           
                                                                                                                                                                                                                                              
  ┌────────────────────────────────────────────────────────────┬─────────────────────────────┬─────────────────────────────────────────────────┐                                                                                              
  │                            File                            │         Types used          │                     Purpose                     │                                                                                              
  ├────────────────────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────────────────────────────┤                                                                                              
  │ model/site/actions/localSiteActions.ts                     │ All 4                       │ Local Redux state operations (offline)          │                                                                                              
  ├────────────────────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────────────────────────────┤                                                                                              
  │ model/site/actions/siteActions.ts                          │ All 4                       │ Actual GraphQL mutation calls (the online path) │
  ├────────────────────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────────────────────────────┤                                                                                              
  │ screens/CreateSiteScreen/CreateSiteScreen.tsx              │ SiteAddMutationInput        │ Create site form handler                        │
  ├────────────────────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────────────────────────────┤                                                                                              
  │ screens/CreateSiteScreen/components/CreateSiteView.tsx     │ SiteAddMutationInput        │ Create site view callback type                  │
  ├────────────────────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────────────────────────────┤                                                                                              
  │ screens/SiteNotesScreen/AddSiteNoteScreen.tsx              │ SiteNoteAddMutationInput    │ Add note form handler                           │
  ├────────────────────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────────────────────────────┤                                                                                              
  │ screens/SiteNotesScreen/components/EditSiteNoteContent.tsx │ SiteNoteUpdateMutationInput │ Edit note form handler                          │
  └────────────────────────────────────────────────────────────┴─────────────────────────────┴─────────────────────────────────────────────────┘                                                                                              
                                                                       
  The split is roughly:                                                                                                                                                                                                                       
  - siteActions.ts — this is the file that actually calls the deprecated mutations (the online/legacy path). Once FF_offline is always on, this file's mutation calls become dead code.
  - localSiteActions.ts — reuses the same types for local state, but never calls the mutations.                                                                                                                                               
  - Screen components — use the types as the shape for form data passed to either the online or offline action creators.
                                                                                                                                                                                                                                              
  So the migration scope is moderate — 6 files, ~20 references. Not urgent since the types will continue to exist in the schema as long as the deprecated mutations do, but worth noting that siteActions.ts is the file that's actually      
  calling the deprecated mutations and would be the first to clean up when FF_offline is permanently on.

dev-client/src/components/dataRequirements/handleMissingData.ts
johannesparty: not sure, sounds like we should talk about it.

dev-client/src/model/site/siteSlice.ts
johannesparty: yep, but I do wonder if we should implement site delete offline.  I feel like yes, but also feel like we could add it later.

dev-client/src/model/soilData/actions/remoteSoilDataActions.ts
johannesparty: I can't help you

dev-client/src/model/sync/results.ts
johannesparty: perhaps, but let's just leave it.

dev-client/src/model/sync/syncGlobalReducer.ts
johannesparty: let's discuss

dev-client/src/model/sync/results.ts
knipec: Oh, I already extracted a few 🙃

dev-client/src/screens/SitesScreen/components/TemporaryLocationCallout.tsx
johannesparty: sounds like using null most everywhere is good (works better than undefined because db layer returns it plus json has null value, but undefined is more implicit).

lots of one-line file changes in mobile-client, commit 8305198f.
two files changed in terraso-client-shared, commit 439657f.

i listed the commits so we can decide to defer this if we want.
but... I say go for it!

dev-client/src/screens/SitesScreen/utils/siteFeatureCollection.ts
johannesparty: claude thinks the only way this could happen is if the redux state is corrupted on disk, so claude recommends keeping it to be defensive.  

but I think if redux state is corrupted, probably a bigger problem, so this probably isn't the way to fix it.  probably the way to fix it would be to validate the state when reading it in and either delete it (hopefully pull will work), or try to at least save the valid parts (could be tricky).

I also wonder if the thing above had to do with the bug when entering coordinates where it would start to parse  before you entered both lat and long 0r something (you fixed this bug).

let's discuss!

knipec: - We think we can remove this code. Carissa will do this.
- Perhaps a separate tech debt task to do validation on the redux state on app open, but we don't know that's actually an issue / worth doing, so probably not going to bother

dev-client/src/model/sync/syncGlobalReducer.ts
knipec: DONE: Add an example to the comments

DONE: What if this happens and you're only a viewer? Does it change? Should it? (Make sure it doesn't retry if it fails) --> It does allow elevation to update via viewers. No error happens.

dev-client/src/model/site/siteSlice.ts
knipec: I'd strongly vote for let's add "delete sites offline" later :) 
I'll add a comment about it though

dev-client/src/model/site/actions/remoteSiteActions.ts
```diff
@@ -0,0 +1,123 @@
...
+    ).map(n => n.id);
+
+    if (isNew) {
+      return {
+        siteId,
+        isNew: true,
+        name: site.name,
+        latitude: site.latitude,
+        longitude: site.longitude,
+        elevation: site.elevation ?? undefined,
```
knipec: Should be `null`, not `undefined` now. 
Also, if we're going with `null`, fix up the parts of the code that use `?.elevation` because that will continue resolve to undefined

johannesparty (approved):

PR #3266, opened 2026-04-08
fix: fetch name from apple signon

johannesparty:
1- Grab name from sign-on-with-apple (never worked before) and pass to backend for saving.  Note this requires backend code to be updated to accept the name this way.

2- Display <first>...<last>@<domain> when name isn't available.  for example "iamcool@domain.com" would show as "i...l@domain.com".  Note this only applies if actual name not available.  show up in notes and in team members in a project.

### DIFF
--- a/dev-client/src/auth/index.ts
+++ b/dev-client/src/auth/index.ts
@@ -79,9 +79,15 @@ interface AuthTokens {
   is_new_account?: boolean;
 }
 
+type AppleNameInfo = {
+  firstName?: string;
+  lastName?: string;
+};
+
 async function exchangeToken(
   identityJwt: string,
   provider: Omit<AuthProvider, 'google'> | `google-${'ios' | 'android'}`,
+  appleNames?: AppleNameInfo,
 ) {
   let body: any = {
     provider,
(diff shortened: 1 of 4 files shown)


dev-client/src/auth/index.ts
```diff
@@ -105,23 +121,32 @@ async function exchangeToken(
 
 const apiConfig = getAPIConfig();
 
-async function getAppleIdToken() {
+type AppleSignInResult = {
+  idToken: string;
+  firstName?: string;
+  lastName?: string;
+};
```
knipec, line 128: Kinda pedantic, but I think the type should be 
``` 
type AppleSignInResult = {
  idToken: string;
  firstName: string | undefined;
  lastName: string | undefined;
};
```
There aren't that many scenarios in which it matters. But as written this type definition is saying "the object might not have this attribute at all", and its instantiation is more like "the object always has the attribute, it just might be undefined"

(I think same thing with the `AppleNameInfo` type definition)

knipec (approved):

dev-client/src/auth/index.ts
johannesparty: done

PR #3273, opened 2026-04-22
chore: clean up build-time warnings 

johannesparty:
Suppresses most build-time noise that was making real warnings hard to spot, and adds elapsed-time reporting to rebuild_all.

two sets of changes: 
1) filter/script change, which you can just ignore I think.
2) package.json/package-lock.json, only did some minor updates, but cleaned up some warnings.

if you want to test, just do "npm run" to see the rebuild-all-... options.

Filter changes (scripts/filter-ios-logs.js):
- Added NOISE_PATTERNS for transitive npm deprecations that aren't fixable from this repo (inflight, lodash.*, rimraf, glob, abab, domexception, boolean, integrity-check warnings), xcbeautify "will be run during every build" warnings, iOS simulator "Failed to send CA Event" chatter, and the known PostHog/libwebp VP8LPredictor duplicate-symbol collision.
- Added preamble buffering for multi-line ⚠️ xcodeproj warnings whose identifying detail only arrives a few lines later (e.g. VP8L).
- Added "iOS Bundled" / xcbeautify "›" / npm prefix to NEW_ENTRY_PATTERN so noise blocks terminate cleanly against Metro progress output.
- Timing: on "iOS Bundled" line, emit "App launched in Xm SSs"; on __REBUILD_EXIT__ sentinel from rebuild_all, emit "Build failed after Xm SSs" if the app never launched. Ctrl+C path stays silent.

rebuild_all changes:
- Use `npm ci` instead of `npm install` for reproducible clean installs; fails loudly on package.json/lock drift rather than auto-reconciling.
- Pipe entire script output through the log filter via `exec`, so npm ci deprecations, watchman chatter, and xcbeautify all get filtered — not just `npm run ios`.
- Ctrl+C cleanup trap that kills Metro (:8081) and xcodebuild, which the RN CLI spawns as detached process groups.
- Clear Metro + haste + watchman caches after a fresh install, since stale caches cause "Unable to resolve module" failures when the node_modules tree changes.
- EXIT trap emits __REBUILD_EXIT__:$? sentinel for the filter's timing logic; cleanup trap disables it so Ctrl+C stays silent.
- Added print_usage() showing npm-script and bash invocations.

package.json changes:
- Bumped @react-native/babel-preset 0.76.6 → 0.81.6 to match RN 0.81.5. Removes the jscodeshift transitive tree (51 packages) and clears three @babel/plugin-proposal-* deprecation warnings.
- Added rebuild-all-ios and rebuild-all-android shortcuts alongside the generic rebuild-all.

package-lock.json changes:
- npm audit fix: 37 → 24 vulnerabilities (critical eliminated, including fast-xml-parser RCE). Remaining 24 are all in dev/test-only chains (react-devtools, native-base, jest-expo) requiring major-version downgrades to fix; deferred.
- npm update baseline-browser-mapping 2.8.17 → 2.10.21 to silence the "data is over two months old" nag.

### DIFF
--- a/dev-client/scripts/filter-build-logs.js
+++ b/dev-client/scripts/filter-build-logs.js
@@ -1,7 +1,9 @@
 #!/usr/bin/env node
 
 /**
- * Filters noisy iOS simulator log messages from stdin.
+ * Filters noisy lines from iOS / Android build + runtime output on stdin.
+ * Covers xcodebuild + Metro + iOS simulator logs on the iOS side, and Gradle
+ * + Kotlin/Java compiler + Metro + logcat-like output on the Android side.
  *
  * All patterns use block suppression: when a line matches, it and all
  * subsequent continuation lines are suppressed until '}}' (end of an NSError
@@ -12,7 +14,7 @@
  * To add a new pattern, just add a regex to NOISE_PATTERNS. No need to worry
  * about whether the message might be split — block suppression handles it.
  *
- * Usage: some_command 2>&1 | node scripts/filter-ios-logs.js [--debug]
+ * Usage: some_command 2>&1 | node scripts/filter-build-logs.js [--debug]
  *   --debug: show all lines, prefixing filtered ones with     [filtered]
  */
 
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -123,6 +123,7 @@
         "jest-expo": "~54.0.12",
         "lint-staged": "^15.5.2",
         "lodash": "^4.17.21",
+        "node-pty": "^1.1.0",
         "patch-package": "^8.0.1",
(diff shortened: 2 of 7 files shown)


dev-client/scripts/filter-build-logs.js
knipec, line 1: If we encounter logs in the future for ios (or android) that we want to filter out, would we just add the full text of them here wrapped in `/` and `/`? 

Maybe a comment somewhere letting other folks reading the code know how to use this?

knipec (approved):

dev-client/scripts/filter-build-logs.js
johannesparty: looking now.

knipec (approved):

PR #3274, opened 2026-04-23
fix: remove FF_offline and FF_select_soil flags

### LINKED ISSUE #2160: Remove FF_offline feature flag
Once the FF_offline feature flag has been turned on in production for a while, and we feel we will never want to turn it off:
- Remove the flag from the codebase
- Remove codepaths that will no longer be hit (that were conditional on the flag being false)

knipec:
## Description
Removes feature flags that have been on for a while now in prod: 
- `FF_offline` (for changing soilData, soilMetadata, and site data offline)
- `FF_select_soil` (for rating a soil match)

### Related Issues
Fixes https://github.com/techmatters/terraso-mobile-client/issues/2160

### Claude prompt
The `FF_offline` feature flag is permanently enabled and we're removing it entirely; no trace of it should remain in the code. Please do this. Start by grepping for all references to FF_offline across the codebase (including string literals, test files, and config). Then work through them: remove the flag check and keep only the flag-on codepath, clean up any code that exists solely to support this flag (e.g., mocks of the featureFlags module that only mock FF_offline), and delete tests that only cover the flag-off path. For tests covering both paths, drop the flag-off cases and keep flag-on as the new baseline.

If something isn't straightforward — e.g., a shared helper that's only reachable via the flag-off path, newly-dead code with unclear ownership, or related config/env vars — skip it and collect the question. Handle all the straightforward cases first.
Don't commit. At the end, give me: (1) files changed with a one-line summary each, (2) the questions and edge cases you collected, and (3) anything deleted or remaining that's worth a second look.

... 

Now do the same thing with the `FF_select_soil` flag.


johannesparty (approved):
briefly reviewed, plus asked claude to review.

one thing we found:
site.notes.offline, site.soil_id.matches.selector, and site.offline.alert_body exist for es, fr, ka, sw, uk even though removed from es.json.

ok handling this separately when we do the localization.

knipec:
> briefly reviewed, plus asked claude to review.
> 
> one thing we found: site.notes.offline, site.soil_id.matches.selector, and site.offline.alert_body exist for es, fr, ka, sw, uk even though removed from es.json.
> 
> ok handling this separately when we do the localization.

Ah good catch -- turns out `npm run localization-check-missing` only checks if non-english strings are missing based on what's in the en.json, not the other way as well, so that wouldn't have caught it. Deleted.

PR #3276, opened 2026-04-24
fix: ios launch banner

johannesparty:
fix the rebuild-all script.
feel free to make fun of me for wasting time on this.
at least I think it works well now.

### DIFF
--- a/dev-client/scripts/filter-build-logs.js
+++ b/dev-client/scripts/filter-build-logs.js
@@ -20,6 +20,13 @@
 
 const readline = require('readline');
 
+// Defensive: force synchronous stdout in case this filter is ever invoked
+// as a non-terminal pipeline stage. When stdout is a TTY (the common case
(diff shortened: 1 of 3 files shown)


knipec (approved):
Summary from conversation: rebuild_all script 
- does a clean build (for if something in the build isn't working)
- filter out console warnings that we know are harmless
- does `npm run {ios, android}`
- computes how long it took to run. And when it launches the app, some extra characters in the script didn't detect when it was launching. So now it will correctly show the time it took to launch the app (in the console).

dev-client/scripts/filter-build-logs.js
knipec, line 1: Could change the overall name of the PR to be more accurate though

johannesparty: this was done! (used to say iOS)

PR #3281, opened 2026-04-28
fix: i18n checking enhancements

johannesparty:
Overhauls the i18n tooling on top of the smarter scanner work.
                                                                                                                                                                                  
  scripts/localization/ becomes scripts/i18n/, and the three prefixes (localization-*, check-i18n, find-i18n-*) collapse into two: i18n-* for in-house tools and poeditor-* for
  external-service ops. New npm run i18n-check is the bundled gate — it runs four checks (source ↔ catalog, locale variable consistency, locale completeness, lingering TK        
  placeholders) and aggregates findings into errors that block and warnings that inform, each with copy-pasteable fix commands.
                                                                                                                                                                                  
  The same gate now runs in three places from one source of truth: pre-commit hook (~0.8s), CI on every push/PR, and as a poeditor-merge pre-flight. Two new helper commands      
  accompany it: npm run i18n-fill-missing (creates TK <english> placeholders) and npm run i18n-delete-unused (dry-run by default; -- --yes to delete). Plus: poeditor-merge
  refuses to run on main outside --dry-run, the broken find-variables.test.mjs is repaired, and localization-merge.md moves next to its script as poeditor-merge.md.              
                                                            
  Wiki updated separately.


knipec (approved):
Summary: When you do a git commit or ci build (or poeditor merge), it now runs the check for these 4 things. It'll stop if the check fails.

PR #3285, opened 2026-04-29
feat: i18n-check warns on empty sections; delete-unused prunes them

johannesparty:
Empty sections — `"speed_dial": {}` left behind after every child key was deleted — were invisible to both the check and the cleanup script. The catalog walker only collects leaf keys, and deleteAtPath in delete-unused only removed the leaf, never the now-empty parent.

Two changes:

  - check.mjs: walk every locale JSON for empty plain-object sections, surface them as a warning (same severity as "unused keys"). Reports per-locale counts and dotted paths. Fix hint points at i18n-delete- unused, which now handles the cleanup.

  - delete-unused.mjs: after the unused-key delete pass, run pruneEmpty over every locale to remove `"foo": {}` blocks. The prune pass runs even when there are zero unused keys, so the same command tidies up pre-existing empty sections too. Dry-run shows everything that would change; --yes applies.

Both pre-existing empty sections (e.g. from a manual edit) and ones that would result from the unused-key deletion are caught.

### DIFF
--- a/dev-client/scripts/i18n/check.mjs
+++ b/dev-client/scripts/i18n/check.mjs
@@ -18,12 +18,13 @@
  */
 
 /*
- * The bundled i18n gate. Runs four checks in turn and aggregates results
+ * The bundled i18n gate. Runs the checks in turn and aggregates results
  * into ERRORS (block CI / pre-commit / poeditor-merge) and WARNINGS (printed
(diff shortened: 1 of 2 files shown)


knipec (approved):
From the description: What are "hint points"?

Issue #3247, opened 2026-02-25
Implement Text Input Component: ProjectDashboard: Settings


DerekCaelin:
Implement Text Input Component on the Project Dashboard Settings tab. Replace existing text input fields with new component. View parent issue for context.

## References
- [Text Input Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7992-21173&t=76lsnMhgCwbf1byD-4)
- [Buttons Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7327-9580&t=76lsnMhgCwbf1byD-4)

Related issues:
- [ ] https://github.com/techmatters/terraso-mobile-client/issues/3246

## Design
**[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12920-17196&t=76lsnMhgCwbf1byD-4) as of 2/25/26: review design in Figma for up to date details.**

<img width="392" height="636" alt="Image" src="https://github.com/user-attachments/assets/97c9ea38-21d1-4bbb-89b8-76ac648d4cf0" />

bjorwol:
Confirmed resolved. Tested on 1.4.7 staging, behaving correctly.

CourtneyLee333:
Again, this isn't working so great if I edit my Project Name to less than 3 char - there is nowhere to tap out of the TextInput without tapping into the Project Description field, but the Save button is disabled with no clues as to why. (This happens on Android, not on iOS)

CourtneyLee333:
looks good now!

Issue #3246, opened 2026-02-25
Implement Text Input Component: Create Project


DerekCaelin:
Implement Text Input Component on the Create Project screen. Replace existing text input fields with new component. View parent issue for context.

## References
- [Text Input Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12914-20807&t=76lsnMhgCwbf1byD-4)
- [Buttons Component](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7327-9580&t=76lsnMhgCwbf1byD-4)

Related issues:

## Design
**[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12920-17098&t=76lsnMhgCwbf1byD-4) as of 2/25/26: review design in Figma for up to date details.**

<img width="437" height="442" alt="Image" src="https://github.com/user-attachments/assets/b9034b01-b21f-4212-a64d-143b4cd5f56d" />

knipec:
Perhaps do https://github.com/techmatters/terraso-mobile-client/issues/3289 as part of this

DerekCaelin:
@knipec, @CourtneyLee333 confirmed that this design is still valid, barring the TextField/TextInput verbiage

bjorwol:
Confirmed resolved. Tested on 1.4.7 staging, behaving correctly.

CourtneyLee333:
Same issue with not knowing why the Create button is disabled.

CourtneyLee333:
looks good now!

Issue #3245, opened 2026-02-25
Implement Text Input Component: Add Depth to Site


DerekCaelin:
Implement Text Input Component on the Add Depth to Site sheet. Replace existing text input fields with new component. View parent issue for context.

## References
- [Text Input Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7992-21173&t=76lsnMhgCwbf1byD-4)
- [Buttons Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7327-9580&t=76lsnMhgCwbf1byD-4)

Related issues:
- #3243 
- #3244 

## Design
**[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12914-20807&t=xTGW1GHkM4fJEXHl-4) as of 2/25/26: review design in Figma for up to date details.**

<img width="302" height="532" alt="Image" src="https://github.com/user-attachments/assets/97fdda2f-c8fe-4055-a479-1c1ef3bcd4db" />

bjorwol:
Confirmed resolved. Tested on 1.4.7 staging, behaving correctly.

CourtneyLee333:
Looks good!

Issue #3244, opened 2026-02-25
Implement Text Input Component: Depth Options in Site


DerekCaelin:
Implement Text Input Component on the Add Depth to Project sheet. Replace existing text input fields with new component. View parent issue for context.

## References
- [Text Input Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7992-21173&t=76lsnMhgCwbf1byD-4)
- [Buttons Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7327-9580&t=76lsnMhgCwbf1byD-4)

## Design
**[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12914-20809&t=76lsnMhgCwbf1byD-4) as of 2/25/26: review design in Figma for up to date details.**

<img width="393" height="722" alt="Image" src="https://github.com/user-attachments/assets/e6651e39-5832-47ac-8b27-3775fd38d54c" />

bjorwol:
Confirmed resolved. Tested on 1.4.7 staging, behaving correctly.

CourtneyLee333:
I was able to create this situation. I think the 0000001 is fine behavior, but the 100000 in the To field should trigger an error. This is also reproduceable in Add Depth to site and Add Depth to project screens. 

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/0cccc4e9-1cba-4463-abfe-ca240e9791dc" />

CourtneyLee333:
Looks good!

PR #3288, opened 2026-04-30
feat: [component library] create TextField and FormTextField + replace some usages

knipec:
### TextField 
Standalone field to enter text. Based on [Figma design](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7992-21395&m=dev)

```ts
type TextFieldType = "text" | "email" | "numeric";

// Props for both TextField and FormTextField
type SharedTextFieldProps = {
  // ---- Identification / accessibility ----
  label?: string; // floating label; shows in `placeholder` location if input value and placeholder are both unspecified
  placeholder?: string; // text shown when input value is unspecified
  testID?: string; // forwarded to underlying RN TextInput
  accessibilityLabel?: string; // defaults to `label`, adds `required` for screen readers because asterisk doesn't translate

  // ---- Behavior / Display ----
  type?: TextFieldType; // default 'text'
  multiline?: boolean; // input box is taller, allows newlines, expands vertically with text
  readOnly?: boolean; // displays-only as the helperText
  required?: boolean; // renders asterisk after label
  helperText?: string;
  style?: RNPTextInputProps["style"];

  // ---- Max character count ----
  showCounter?: boolean; // requires maxLength (TS-enforced)
  maxLength?: number; // disallows typing more characters
};
```

```ts
// Props for `TextField` only (controlled)
type TextFieldProps = SharedTextFieldProps & {
  value: string; // often want this to be a state variable
  onChangeText?: (value: string) => void; // often want this to set a state variable
  onBlur?: () => void; // uncommon
  error?: string; // caller-controlled; renders red text below
};
```

### FormTextField
Like TextField, but uses a Formik context to drive some behavior. (Formik is often also used with Yup)

```ts
// Props for `FormTextField` only
  name: StringFieldKeys<TValues>; // must be a formik field name
  onChangeText?: (value: string) => void; // often not needed; only for additional non-formik side-effects
  onBlur?: () => void; // often not needed; only for additional non-formik side-effects
};

// Note: Formik supplies `value`, `error`, plus behavior for `onChangeText`, `onBlur` 
// (but caller can specify additional behavior). 
```

## Related issues
**Create components:**
- https://github.com/techmatters/terraso-mobile-client/issues/1718

**new TextField callers:**
- DeleteAccountConfirmForm
- SiteSettingsScreen

**new FormTextField callers:** 
- DepthTextInputs
  - https://github.com/techmatters/terraso-mobile-client/issues/3244
  - https://github.com/techmatters/terraso-mobile-client/issues/3245
- CreateProjectForm: https://github.com/techmatters/terraso-mobile-client/issues/3246
- EditProjectForm: https://github.com/techmatters/terraso-mobile-client/issues/3247

**Do later due to complexity:**
- AddTeamMemberForm
- CreateSiteForm
- SiteNoteForm (uses FormInput, but I think it should be a different component because the visual treatment is distinct)


johannesparty (approved):
The documentation in TextField.md is nice (although maybe needs updating; see below).  But I wish the "how to use" parts were inn TextField.tsx and FormTextField.tsx because it seems like that's where developer's will go.  It looks like vscode will format "jsdoc" comments which can have .md formatting, but only on hover, so not sure it is really better.  Maybe the best option is to add a comment with a pointer to the .md file in both TextField and FormTextField.  What do you think?

So far I've read the .md file (seems good), and I've read the TextField and FormTextField, seems good to me but I'm hardly an expert.

I did ask claude to review, and mostly it thinks the docs need updating (attached below).  It duplicates a lot of what is in the md file.

Probably going to eat lunch soon, but maybe we can go through it briefly afterwards.  I'm send you a message on slack.  I did do "approve", but we should talk first ideally.


[pr-3288-review.md](https://github.com/user-attachments/files/27773769/pr-3288-review.md)

dev-client/src/components/inputs/TextField.tsx
```diff
@@ -0,0 +1,211 @@
...
+  TextFieldType,
+  TYPE_PRESETS,
+} from 'terraso-mobile-client/components/inputs/TextFieldHelpers';
+import {theme} from 'terraso-mobile-client/theme';
+
+/* TextField — pure, controlled text input.
+ *
+ * Knows nothing about Formik. Caller passes value/onChangeText/error directly.
+ * For Formik-driven forms, use FormTextField (which wraps TextField). */
+
```
knipec: Notes to self: 
- Can "how to use" comments be as useful as the .md file was? Look into jsdoc formatting?
- Can also link to Figma design and remove the .md file to streamline sources of truth.

PR #3294, opened 2026-05-11
feat: move env out of project to make it harder for claude to read by accident

johannesparty:
see https://sites.gitbook.com/preview/site_FIWnZ/technical-notes/privacy-and-security for details.

note that poeditor script change is unrelated, but I ran into it when testing this.
basically need to explicitly pull the tags because git fetch doesn't fetch tags on branches which no longer exist.
(but the tag and data it points to still does exist, so all good).id

ideally that change could have been a separate PR, but doesn't seem worth the time to split it off.


CLAUDE.md
```diff
@@ -89,4 +89,17 @@ cd dev-client
...
+The .env file is not in the repo. Real credentials live at ~/secrets/<project>/.env
+and are loaded at runtime by the npm scripts (via dotenv-cli) — they are not
+present in the interactive shell environment.
+
+When writing scripts that consume environment variables:
+- Never print, log, or echo values from process.env or os.environ
+- Never include env values in error messages or stack traces
+- If verifying a variable is set, print only its presence ("API_KEY: set"),
+  never the value
+- Don't write env values into files inside the project directory
```
knipec, line 105: I haven’t looked into this a lot, but I've seen articles mentioning the “pink elephant problem”-- that explicitly telling AI not to do something may make it more likely to do it: https://eval.16x.engineer/blog/the-pink-elephant-negative-instructions-llms-effectiveness-analysis 

The suggestion is, when possible, to switch to affirmative commands, telling it what to do (like “only do X” instead of “never do Y”)... I don't have obvious ways to rephrase, maybe AI does 😅

CLAUDE.md
```diff
@@ -89,4 +89,17 @@ cd dev-client
 - The app uses a shared library: `terraso-client-shared`
 - Git hooks enforce code style and commit message structure
 - Supports offline functionality with custom persistence layer
-- always run "npm run format" before committing.
\ No newline at end of file
+- always run "npm run format" before committing.
+
+## Secrets handling
+
+The .env file is not in the repo. Real credentials live at ~/secrets/<project>/.env
```
knipec, line 96: Is there a benefit to even telling AI where the real credentials live, if we don't want them to read them?

dev-client/package.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "LandPKS Soil ID",
+  "name": "terraso-mobile-client",
```
knipec, line 2: Why this change?

.claude/settings.json
knipec, line 1: General comments: 
1. We may want to make a note on `.env.sample` and/or the readme to developers about where to put their .env?
2. Wiki: Could help for the wiki to be clear about which things the codebase does for you, and which things you have to do manually (put .env in the ~/secrets directory)
3. Wiki: Not sure how to do the steps here: https://sites.gitbook.com/preview/site_FIWnZ/technical-notes/privacy-and-security/audit-build-pipeline

CLAUDE.md
johannesparty: interesting, I'll look into that.
the existing text (mostly) came from claude itself.

CLAUDE.md
johannesparty: hm... probably not.  but it can figure it out so might not matter.
I'll review and tweak (here and in backend -- same text more or less)

dev-client/package.json
johannesparty: apparently that is one of the package.json rules for name although it isn't enforced.  lower case only, no spaces.  so some lint thingy was complaining, and changing doesn't actually affect anything.

https://docs.npmjs.com/cli/v11/configuring-npm/package-json
this only says "new" names must not have uppercase etc., but anyhow better to change it.
doesn't really matter because we never import it anywhere else, but trying to get rid of warnings :)

CLAUDE.md
johannesparty: ok I decided I like this here because i do want claude to write scripts that reads the file.  plus the path is explicit in the code.  I just don't want claude to read the file, which is solved other ways.

CLAUDE.md
johannesparty: I did some more searching and there are mixed feelings on this topic.
One problem with positive only is a lot of things can't be expressed that way.
And if you are very explicit (we are), most evidence is that the negative works just as well.

.claude/settings.json
johannesparty: 1, done; 2, let me think about it, but wiki separate anyhow; 3, yep, but the idea seems valid just not so easy to do potentially.  I think we are good for now though.

knipec (approved):
Weird that the last commit message mentions changing SYNC_DEBUG_ENABLED since it appears not to have, but seems reasonable

johannesparty:
> Weird that the last commit message mentions changing SYNC_DEBUG_ENABLED since it appears not to have, but seems reasonable

yep I think it was deleted and added back or something.

Issue #3248, opened 2026-02-25
Implement Text Input Component: Add Team Member


DerekCaelin:
Implement Text Input Component on Add Team Member screen. Replace existing text input fields with new component. View parent issue for context.

## References
- [Text Input Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12914-20807&t=76lsnMhgCwbf1byD-4)
- [Buttons Component](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7327-9580&t=76lsnMhgCwbf1byD-4)

## Design
**[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12920-17324&t=76lsnMhgCwbf1byD-4) as of 2/25/26: review design in Figma for up to date details.**

<img width="651" height="832" alt="Image" src="https://github.com/user-attachments/assets/f69b5123-d50a-4ed9-8e89-0be5293c1eb9" />

bjorwol:
Confirmed resolved. Tested on 1.4.7 staging, behaving correctly.

CourtneyLee333:
on both Android and iOS there is no ability to tap out of the TextInput to trigger the "Valid email address required" error

CourtneyLee333:
looks good now!

Issue #3240, opened 2026-02-25
Implement Text Input Component: Create Site Form


DerekCaelin:
Implement Text Input Component on the create site form. Replace existing text input fields with new component. View parent issue for context:

## References
- [Text Input Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7992-21173&t=76lsnMhgCwbf1byD-4)
- [Buttons Component Documentation](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=7327-9580&t=76lsnMhgCwbf1byD-4)

## Design
**[Figma](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12914-21103&t=B9PjrQI3kfkQZK1i-0) as of 2/25/26: review design in Figma for up to date details.**
<img width="410" height="468" alt="Image" src="https://github.com/user-attachments/assets/ab4b3355-6692-4a54-a8f2-bb3542f353d2" />

bjorwol:
Confirmed resolved. Tested on 1.4.7 staging, behaving correctly.

CourtneyLee333:
On Android, I don't see the helper text or error state for the Site Name unless I tap into a coord field. This means that the Create Site button is disabled and user has no clue as to why. Add to Project and Data Privacy control doesn't trigger the error state. I know we talked about the trigger happening when user taps out of the TextInput component, but that doesn't seem to be the behavior. And I'm now thinking that might not be the right call, because if I only want to enter the site name and then hit Create button, without editing any of the other settings, I don't see the error text and the button is disabled. I  also don't think that pattern is applied to the lat and long fields. The behavior for Site Name implemented in the Site Config screen seems fine, so perhaps we should use that. 

(On iOS, I do see the Site Name error upon tapping out of the field.)

<img width="300"  alt="Image" src="https://github.com/user-attachments/assets/bbdf5b38-6a24-4191-b42b-1d8ff1252574" />

CourtneyLee333:
looks good now!

PR #3297, opened 2026-05-19
feat: use FormTextField in a couple more cases + minor form UI updates

knipec:
## Description
Use TextField and FormTextField in AddTeamMemberForm and CreateSiteForm. All places that we've identified so far now have their text fields replaced.

Also, fix a few issues from the previous PR (#3288) and better match forms to [Figma design](https://www.figma.com/design/KMJbhjVyrrLFHj6jGcX9lN/LandPKS-Design-Prototype-working-file?node-id=12914-13954&m=dev).

### Related Issues
#3248 
#3240


CLAUDE.md
```diff
@@ -52,7 +52,7 @@ cd dev-client
 - `src/navigation/` - React Navigation setup
 - `src/model/` - Redux slices and business logic
 - `src/services/` - External API integration
-- `src/translations/` - i18n files (en.json, es.json, uk.json)
+- `src/translations/` - i18n files (en.json, es.json, etc.)
```
johannesparty, line 55: :)

dev-client/src/components/inputs/TextField.tsx
```diff
@@ -181,7 +181,9 @@ export const TextField = forwardRef<RNTextInput, TextFieldProps>(
                 {helperText}
               </HelperText>
             ) : null}
-            {counterText !== undefined && (
+            {/* Counter is suppressed while an error is showing so the error
+             * occupies its slot, rather than the two stacking. */}
+            {counterText !== undefined && !showError && (
```
johannesparty: I was thinking maybe this should be !counterText instead of counterText !== undefined, but it probably doesn't matter.  still as long as counterText is never just empty string.

dev-client/src/translations/en.json
```diff
@@ -312,7 +312,7 @@
             "help_text": "Enter the email address of the person’s Terraso account. A Terraso account can be created by installing the LandPKS Soil ID app and signing in.",
             "user_does_not_exist": "No Terraso account found",
             "user_in_project": "Already a team member",
-            "invalid_email": "Enter a valid email address"
+            "invalid_email": "Valid email address required"
```
johannesparty, line 316: I only see en.json; did you ask claude to change the wording for the other languages?

johannesparty (commented):
I made a few comments, but looks good.
I think the snapshot might still be wrong, but maybe just didn't rerun yet.

dev-client/src/components/inputs/TextField.tsx
knipec: Sure, I can use `counterText`

dev-client/src/translations/en.json
knipec: I hadn't because the new string was fairly similar to the old one, but I can change it

johannesparty (approved):
all good

PR #3300, opened 2026-06-10
feat: remind to re-sync the soil-id DB when soil descriptions change

johannesparty:
The WRB soil Description/Management narratives shown in the app live in `soil.match_info.*` in our i18n files, **but the backend serves the same narratives from the soil-id `wrb_fao90_desc` Postgres table**, not from these JSON files. So when a POEditor merge changes those keys, the backend copy drifts until it is separately re-synced.

This adds a reminder to `poeditor-merge.mjs`: when a merge touches any `soil.match_info.*` key, it prints a note to re-sync the soil-id DB via `soil-id-algorithm/scripts/wrb_descriptions_sync.py` (review the diff, then `--write`) and regenerate/redistribute the soil-id-db dump. The reminder prints on the commit and `--no-commit` paths (not on `--dry-run`, which changes nothing); behavior is otherwise unchanged.

### DIFF
--- a/dev-client/scripts/i18n/poeditor-merge.mjs
+++ b/dev-client/scripts/i18n/poeditor-merge.mjs
@@ -1088,6 +1088,31 @@ async function main() {
 
   const commitMessage = commitLines.join('\n');
 
+  // Heads-up if WRB soil description translations changed: the backend serves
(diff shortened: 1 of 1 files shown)


knipec (approved):
Seems helpful!

PR #3302, opened 2026-06-13
refactor: send raw Munsell to soil-ID instead of client-side LAB

johannesparty:
Stop converting Munsell → LAB on the client for the soil-ID match request. Send the stored hue/value/chroma directly via the backend's new `colorMunsellNumeric` input; the backend converts to LAB using its soil-id reference table, so the conversion happens in exactly one place.

This is the **mobile counterpart to backend techmatters/terraso-backend#2052** (the soil-ID API now accepts Munsell color), and it closes a latent inconsistency between mobile's `npm-munsell` LAB and the backend's table-derived LAB (slightly different L\*a\*b\* for the same Munsell triple).

### Changes
- `soilIdMatchInputs.ts` — drop the `mhvcToLab` import and the `soilDataLabColorInput` helper; build `colorMunsellNumeric` inline from `colorHue`/`colorValue`/`colorChroma`, keeping the all-three-numbers guard (partial Munsell is meaningless to send).
- Tests updated (raw-forwarding case + omit-when-a-coordinate-is-missing case).
- `npm munsell` stays — still used by `colorDetection.ts` (photo gamut check) and `colorConversions.ts` (swatches). Only the soil-ID request path drops the LAB conversion.
- Sync/push path untouched (already sends raw `colorHue`/`colorValue`/`colorChroma`).

### ⚠️ Behavior change worth a look in review
Out-of-gamut photo-derived colors (chroma ≳ 30) previously shipped an interpolated LAB to the backend. Now those Munsell triples won't match the backend's reference table, so the server silently drops the color for that depth (no error). Intentional — those were junk values.

### ⚠️ Merge prerequisite (cross-repo)
Depends on **terraso-client-shared** GraphQL types including `MunsellColorInput` / `colorMunsellNumeric` (generated from the #2052 backend schema). The `terraso-client-shared` pin in `dev-client/package.json` must bump to a tagged release that has them before this can merge — they were only generated locally during development so far.

### DIFF
--- a/dev-client/package.json
+++ b/dev-client/package.json
@@ -112,7 +112,7 @@
     "react-redux": "^9.2.0",
     "reduce-reducers": "^1.0.4",
     "setimmediate": "^1.0.5",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#2026-04-23.0",
+    "terraso-client-shared": "github:techmatters/terraso-client-shared#6cb06d2b40562f8a7f9f28a1e4f349cc005763cb",
     "use-debounce": "^10.0.6",
     "uuid": "^11.1.0",
     "yup": "^1.7.1"
--- a/dev-client/package.json
+++ b/dev-client/package.json
@@ -112,7 +112,7 @@
     "react-redux": "^9.2.0",
     "reduce-reducers": "^1.0.4",
     "setimmediate": "^1.0.5",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#6cb06d2b40562f8a7f9f28a1e4f349cc005763cb",
+    "terraso-client-shared": "github:techmatters/terraso-client-shared#2026-06-17.0",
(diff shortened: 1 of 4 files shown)


dev-client/package.json
```diff
@@ -112,7 +112,7 @@
     "react-redux": "^9.2.0",
     "reduce-reducers": "^1.0.4",
     "setimmediate": "^1.0.5",
-    "terraso-client-shared": "github:techmatters/terraso-client-shared#2026-04-23.0",
+    "terraso-client-shared": "github:techmatters/terraso-client-shared#6cb06d2b40562f8a7f9f28a1e4f349cc005763cb",
```
knipec: I assume once the client-shared PR gets merged, you tag it and update this?

knipec (approved):
I imagine this is already in the backend but could be helpful to quick checkin with @garobrik since they implemented the LandPKS color module stuff

PR #3305, opened 2026-06-25
fix: allow soil depths up to 300cm (and log real push failures)

johannesparty:
## Summary
- Raise `DEPTH_MAX` from 200 to 300, lifting the cap on all three site-level depth-interval paths (NRCS, BLM, CUSTOM). NRCS users now see an Add Depth button to extend past 200cm; BLM/CUSTOM users can add intervals up to 300 instead of 200.
- Mirror the recent pull-side logging fix (#3301) on the push side so a rejected `pushUserData` thunk surfaces the real error instead of the `{message: "Rejected"}` placeholder.

## Notes
- The NRCS preset table is unchanged (still ships 0..200). Users opt in to deeper intervals by adding them.
- terraso-backend still rejects intervals > 200 with `INVALID_DATA`. Until the backend cap is raised, the client-side change isn't useful end-to-end — but it's a no-op for users who don't try to add a > 200 interval.
- The push-side logging fix only affects thunk-level rejections (network errors, etc). Per-record server failures still come back via the existing `logPushResults` channel as enum reasons.

## Test plan
- [ ] Site with NRCS preset: Add Depth button appears; can add an interval ending at 300
- [ ] Site with BLM preset: Add Depth button works as before; can add an interval ending at 300
- [ ] Site with CUSTOM preset: existing add/edit flows still work; cap is 300
- [ ] Try to add an interval ending > 300: validation rejects with the standard message
- [ ] Trigger a push thunk rejection (e.g. airplane mode mid-push): console shows a real error object, not `{message: "Rejected"}`

### DIFF
--- a/dev-client/package-lock.json
+++ b/dev-client/package-lock.json
@@ -83,7 +83,7 @@
         "react-redux": "^9.2.0",
         "reduce-reducers": "^1.0.4",
         "setimmediate": "^1.0.5",
-        "terraso-client-shared": "github:techmatters/terraso-client-shared#2026-06-17.0",
+        "terraso-client-shared": "github:techmatters/terraso-client-shared#2026-06-26.1",
         "use-debounce": "^10.0.6",
         "uuid": "^11.1.0",
         "yup": "^1.7.1"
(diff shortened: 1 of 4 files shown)


knipec (approved):
Reminder to update package json once shared-client is in

PR #3311, opened 2026-07-01
fix: update version to 1.4.7

knipec:
## Description
iOS doesn't let us make a new staging build unless the version number is greater than the previously globally-released version. So, bump version to 1.4.7 using the update-version script


johannesparty (approved):
yes -- a simple one :)

Issue #3312, opened 2026-07-02
Crash "pointer being freed was not allocated"


johannesparty:
### App Version

1.4.6 (347)

### Platforms

- [x] iOS
- [ ] Android
- [ ] iOS simulator
- [ ] Android emulator

### Description

Sentry Report:

https://techmatters.sentry.io/issues/7587577261/?groupId=7587577261&pageEnd=2026-07-02T05%3A13%3A19.307&pageStart=2026-07-01T05%3A13%3A19.307&project=4506299761623040&referrer=alert_email&source=issue_details

Claude Analysis:
Root cause

Reading bottom-up through the RNMBX frame:

1. SiteMap.tsx:305 renders <Mapbox.Images images={mapImages} /> with three require()d PNGs (sitePin, selectedSitePin, temporarySitePin — SiteMap.tsx:359).
2. To load each, RNMBX enqueues an RNMBXImageQueueOperation. In RNMBXImageQueue.swift:68 its start() does DispatchQueue.global(qos: .default).async { ... } and at line 72 calls React's RCTImageLoader.loadImage.
3. That resolves the local asset via RCTImageFromLocalAssetURL → +[UIImage imageNamed:inBundle:withConfiguration:].
4. imageNamed: reaches into Apple's CUICatalog / -[CUICatalog negativeCache] → -[NSCache setName:]. CUICatalog's cache is not thread-safe. When two of these run concurrently on background threads, they corrupt the malloc heap → pointer being freed was not allocated → SIGABRT.

The queue at RNMBXImageQueue.swift:127 sets no maxConcurrentOperationCount, so all three pins can resolve at once on separate background threads — exactly the race window.

Why only one report

It needs several things to line up at once: two+ of the pin images resolving simultaneously, both landing off the main thread, and hitting CUICatalog's cache in the narrow unsynchronized window (usually first map load, cold catalog cache). Most of the time one image wins the cache before the others touch it, or they don't overlap. So it's a genuine rare heisen-crash, not something broadly broken.

Options

It's third-party + Apple, but you already ship native patches (patches/native-base+3.4.28.patch), so a patch-package fix is on-brand here:

- A — serialize the queue (smallest, safest): patch RNMBXImageQueue.swift:127 to add result.maxConcurrentOperationCount = 1. All observed crashing frames are inside this one queue, so serializing its own loads removes the race for these three pins. One line, low risk.
[...]

PR #3313, opened 2026-07-02
fix: serialize RNMBX image queue to avoid CUICatalog heap corruption

johannesparty:
## Official bug report: 

https://github.com/techmatters/terraso-mobile-client/issues/3312

Note: this is a simple patch in RN MapBox image loading, but it occurs infrequently so maybe not urgent.  On the other hand, we could be having invalid free without detecting, which is pretty bad.

## Problem

A Sentry crash report showed a native `SIGABRT` — `pointer being freed was not allocated` — from the iOS asset-catalog cache:

```
libsystem_malloc  ___BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED
CoreFoundation    -[NSCache setName:]
CoreUI            -[CUICatalog negativeCache]
...
React             RCTImageFromLocalAssetURL
LandPKSSoilID     closure in RNMBXImageQueueOperation.start (RNMBXImageQueue.swift:72)
libdispatch       _dispatch_worker_thread2
```

## Root cause

`SiteMap.tsx` renders `<Mapbox.Images images={mapImages} />` with three local pin PNGs (`sitePin`, `selectedSitePin`, `temporarySitePin`). To load each, `@rnmapbox/maps` enqueues an `RNMBXImageQueueOperation` whose `start()` dispatches to a background global queue and calls React's image loader, which resolves the local asset via `+[UIImage imageNamed:inBundle:]`.

That reaches into Apple's `CUICatalog` cache, which is **not thread-safe**. The RNMBX download `OperationQueue` sets no `maxConcurrentOperationCount`, so multiple pin images can resolve simultaneously on separate background threads and race on that cache — corrupting the heap and aborting.

This is rare (a single report) because it requires two+ images resolving at once, both off the main thread, hitting the narrow unsynchronized window on a cold catalog cache — usually one image wins the cache first.

## Fix

`patch-package` patch that sets `maxConcurrentOperationCount = 1` on the RNMBX image download `OperationQueue`, serializing these loads so they can't race on CUICatalog. Applied automatically via the existing `postinstall` → `patch-package` hook; takes effect on the next native iOS build.

## Notes

- Native-only change — needs a native rebuild (not just a Metro reload) to take effect.
- Patch is pinned to `@rnmapbox+maps+10.2.5`; a version bump will require regenerating it.
- Hard to unit-verify (native heisen-race); confidence rests on the patch matching the crash stack exactly — all crashing frames are inside this one queue.

### DIFF
--- a/dev-client/patches/@rnmapbox+maps+10.2.5.patch
+++ b/dev-client/patches/@rnmapbox+maps+10.2.5.patch
@@ -1,0 +1,17 @@
+diff --git a/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXImageQueue.swift b/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXImageQueue.swift
+index 0560de2..ca2d772 100644
+--- a/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXImageQueue.swift
++++ b/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXImageQueue.swift
+@@ -127,6 +127,12 @@ class RNMBXImageQueue {
+   var imageQueue : OperationQueue = {
+     let result = OperationQueue()
+     result.name = "comp.mapbox.RNMBX.DownloadImageQueue"
++    // Serialize image loads. Each operation resolves a local asset off the main
++    // thread via +[UIImage imageNamed:inBundle:], which reaches into Apple's
++    // non-thread-safe CUICatalog cache. Concurrent operations race on that cache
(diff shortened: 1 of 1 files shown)


knipec (approved):
Sounds like we're not sure how to have confidence this fixes the issue, but it seems unlikely to cause new issues. Only thing I can think of is if operations were happening concurrently that maybe something will take slightly longer in series but can't imagine that being much of an issue.

PR #3316, opened 2026-07-10
TextField and FormTextField show error on first focus, instead of on "blur"

### LINKED ISSUE #1718: Make TextInput consistent with component design

knipec:
## Description
**Motivation**: The `errorTiming = afterBlur` option didn't work well. Different devices and forms vs standalone TextFields seemed to behave inconsistently (like, Android considered tapping out of the field to be 'onBlur', but on iOS it would only count if you tapped to another text field). 

**Decision**: Courtney and I decided to show the error to the user on `afterFirstFocus` instead of `afterBlur`. This is more consistent throughout scenarios and slightly simplifies the code.

**Changes**:
- Introduce and default to `afterFirstFocus`
- Remove `afterBlur`. (Not sure if `immediate` is reaaally needed anymore, but kept it in for one hypothetical case, and in case we want to extend errorTiming logic options)
- Also updated a few minor things in specific usages:
  - Manual steepness degree error text showed wrong string
  - Create Site should show character count

Also we should update the Figma and props documentation

### Related Issues
Fixes https://github.com/techmatters/terraso-mobile-client/issues/1718 and several of the "Implement" tasks (https://github.com/techmatters/terraso-mobile-client/issues/3240, etc)


knipec:
The branch was made from another feature branch, but that makes the PR confusing. Cherry-picked the two commits and moving the PR over to https://github.com/techmatters/terraso-mobile-client/pull/3317 instead.

PR #3318, opened 2026-07-14
fix: validate coordinates before moving the map camera

johannesparty:
## Problem

A Sentry crash — `EXC_BAD_ACCESS` in Hermes — traced to Mapbox's native `setCamera` throwing *"coordinates must contain numbers"* when handed a non-finite `centerCoordinate`, which Hermes then crashes on while converting the native exception to a JS error.

https://techmatters.sentry.io/issues/7609074277/?referrer=alert_email&alert_type=email&alert_timestamp=1783988264473&alert_rule_id=14823299&notification_uuid=cb5ccee8-4cf1-45b1-bd7b-16b32533836c&environment=production

User-location coords flow from `@rnmapbox/maps`' `locationManager` into Redux (`BottomTabsScreen`), then to `moveToPoint` → `coordsToPosition` → `setCamera`. The only guard along the way is `currentUserCoords !== null` (`SitesScreen.tsx:113`), which does **not** catch `NaN`/`undefined` latitude or longitude. A device that returns a non-null but invalid location (per the report, a Wi-Fi-only iPad surfacing a CoreLocation fallback) sends non-finite coords straight to native Mapbox.

## Fix

Add an `isValidCoords` type guard in `util.ts`, reusing the existing `isValidLatitude`/`isValidLongitude` (which already reject `NaN`/`undefined`, since `NaN >= -90` is `false`). Apply it at two choke points:

- **`BottomTabsScreen`** — don't dispatch invalid coords into the store, so garbage GPS never enters Redux (both `getLastKnownLocation` and the movement listener).
- **`SiteMap.moveToPoint`** — skip `setCamera` for invalid coords, protecting every caller (initial camera, search, site pins) as a backstop.

Adds unit tests for `isValidCoords` (finite/in-range, null, `NaN`, `undefined`, out-of-range).

## User-facing behavior

Instead of crashing, the app falls back gracefully and self-corrects:

| Situation | Before | After |
|---|---|---|
| Non-null but non-finite coords on load | **Crash (SIGABRT)** | Default US view (`DEFAULT_LOCATION`, zoom 3), then auto-centers when the first valid fix arrives |
| Tap "locate me" with no valid fix | crash risk | Silent no-op |

The initial-camera effect doesn't latch as "done" on invalid coords, so it still auto-centers once a valid fix lands. The native user-location puck is unaffected (it reads device location directly, not Redux).

## Notes

- The upstream question of *why* a device returns a non-finite location remains a theory; this makes it a non-event either way.
- Verified: `check-ts`, eslint/prettier, and unit tests (23 passed) all green.

### DIFF
--- a/dev-client/src/util.ts
+++ b/dev-client/src/util.ts
@@ -172,6 +172,17 @@ export const isValidCoordinates = (input: string) => {
   return isValidLatitude(latitude) && isValidLongitude(longitude);
 };
 
+// Guards against non-finite coordinates (NaN/undefined/Infinity) reaching
+// native map APIs. Mapbox's setCamera throws "coordinates must contain numbers"
+// on a non-numeric centerCoordinate, which crashes the app via Hermes. Some
+// devices (e.g. Wi-Fi-only iPads using a CoreLocation fallback) can surface a
+// non-null location whose latitude/longitude are not finite numbers.
+// isValidLatitude/isValidLongitude already reject these (NaN >= -90 is false).
+export const isValidCoords = (coords: Coords | null): coords is Coords =>
+  coords !== null &&
+  isValidLatitude(coords.latitude) &&
+  isValidLongitude(coords.longitude);
+
 export const getSoilWebUrl = (coords: Coords) => {
   return `https://casoilresource.lawr.ucdavis.edu/gmap/?loc=${coords.latitude},${coords.longitude}`;
 };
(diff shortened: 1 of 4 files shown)


dev-client/src/util.ts
```diff
@@ -172,6 +172,17 @@ export const isValidCoordinates = (input: string) => {
   return isValidLatitude(latitude) && isValidLongitude(longitude);
 };
 
+// Guards against non-finite coordinates (NaN/undefined/Infinity) reaching
+// native map APIs. Mapbox's setCamera throws "coordinates must contain numbers"
+// on a non-numeric centerCoordinate, which crashes the app via Hermes. Some
+// devices (e.g. Wi-Fi-only iPads using a CoreLocation fallback) can surface a
+// non-null location whose latitude/longitude are not finite numbers.
+// isValidLatitude/isValidLongitude already reject these (NaN >= -90 is false).
+export const isValidCoords = (coords: Coords | null): coords is Coords =>
```
knipec, line 181: Could rename this or `isValidCoordinates` to disambiguate?
Though maybe it's already clear from the function signatures

knipec (approved):

dev-client/src/util.ts
johannesparty: good idea but lets skip.

PR #3319, opened 2026-07-19
fix: add "Unaffiliated (no project)" option to sites list filter

### LINKED ISSUE #2665: Filter Sites List by project needs "unaffiliated" option

johannesparty:
## Summary
- Closes #2665
- Adds an "Unaffiliated (no project)" option as the second entry of the project dropdown in the Sites screen filter modal (after the "None" option that clears the filter)
- Selecting it filters the list to sites whose `projectId` is undefined

## Implementation notes
- New `UNAFFILIATED_PROJECT_FILTER_VALUE` sentinel constant in `sitesScreenFilters.ts`
- Project filter's function now special-cases the sentinel — matches sites with no project — and otherwise falls back to strict equality (behavior unchanged for real project IDs)
- New `general.filter.unaffiliated` translation key added to all locales. English/Spanish/French have real translations; ka/sw/uk carry English placeholders for POEditor to localize

## Not in scope
Issue comment thread also raised the question of renaming/hiding the "None" (clear-filter) label. That is separate and untouched here — this PR only adds the missing option.

## Test plan
- [ ] Open Sites screen → open filter modal → open the project dropdown
- [ ] Confirm order: `None`, `Unaffiliated (no project)`, then user's projects
- [ ] Pick `Unaffiliated (no project)` → Apply → only sites with no project remain
- [ ] Pick a real project → Apply → only that project's sites remain (regression check)
- [ ] Pick `None` → Apply → filter clears
- [ ] Switch language to es/fr → label localized; ka/sw/uk fall back to English

### DIFF
--- a/dev-client/src/screens/SitesScreen/utils/sitesScreenFilters.ts
+++ b/dev-client/src/screens/SitesScreen/utils/sitesScreenFilters.ts
@@ -22,6 +22,12 @@ import {normalizeText} from 'terraso-client-shared/utils';
 import {SortingOption} from 'terraso-mobile-client/components/ListFilter';
 import {equals, searchText} from 'terraso-mobile-client/util';
 
+// Sentinel filter value for the "Unaffiliated (no project)" option in the
+// project filter. Sites without a project have `projectId === undefined`,
+// which ListFilter's default lookup stringifies to "undefined" — the custom
+// filter function below matches on that.
+export const UNAFFILIATED_PROJECT_FILTER_VALUE = '__unaffiliated__';
+
 export const getSitesScreenFilters = (
   siteDistances: Record<string, number> | null,
   siteProjectRoles: {
@@ -64,7 +70,10 @@ export const getSitesScreenFilters = (
     },
     project: {
       kind: 'filter',
-      f: equals,
+      f: (val: string) => (comp: string | undefined) =>
+        val === UNAFFILIATED_PROJECT_FILTER_VALUE
+          ? comp === undefined || comp === 'undefined'
+          : val === comp,
       lookup: {
         key: 'projectId',
       },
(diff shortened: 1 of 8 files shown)


dev-client/src/screens/SitesScreen/utils/sitesScreenFilters.ts
```diff
@@ -64,7 +70,10 @@ export const getSitesScreenFilters = (
     },
     project: {
       kind: 'filter',
-      f: equals,
+      f: (val: string) => (comp: string | undefined) =>
+        val === UNAFFILIATED_PROJECT_FILTER_VALUE
+          ? comp === undefined || comp === 'undefined'
```
knipec: Huh, are there sites with a project that is `'undefined'` as a string?

dev-client/src/translations/ka.json
```diff
@@ -376,7 +376,8 @@
         "email_label": "ელექტრონული ფოსტის მისამართი",
         "filter": {
             "no_role": "არცერთი",
-            "no_project": "არცერთი"
+            "no_project": "არცერთი",
+            "unaffiliated": "Unaffiliated (no project)"
```
knipec, line 380: Typically we have added english strings with "TK" so we can search on that to make sure nothing is untranslated

knipec (approved):

dev-client/src/screens/SitesScreen/utils/sitesScreenFilters.ts
johannesparty: turns out we can get 'undefined' as a string, but not as actually undefined.
I've eliminated the undefined case because it can never happen.

dev-client/src/translations/ka.json
johannesparty: I checked them all and I think the two-way merge still marks new ones as needing review.  but we should do the poeditor merge too at some point.

Issue #3314, opened 2026-07-08
Deleting a project should immediately delete its sites


knipec:
## Repro
1. Put a site in a project
2. Delete the project
3. Go to the Sites screen
Expected: The site that was in the project should not appear in the Sites list
Actual: It does
4. Click on the site
Actual, without the `project === undefined` check in selectors.ts: crash
Actual, otherwise: Wait for a pull or make a site observation: sync conflict dialog

## TODO
1. Fix it
2. Take some time to look into why the delete user changes made this crash... is there some underlying issue there with how the project deletion cascades?
3. Update comment & possibly remove the `project === undefined` check in selectors.ts 
4. Write tests

DerekCaelin:
Tested this with [TC-019](https://docs.google.com/spreadsheets/d/1ydTm3Dc8eL7qi66m08-znV8R1DWuIT0XlcWOTK_xCLc/edit?gid=1019335581#gid=1019335581&range=24:24) on version 1.4.7 (352) staging

I created a site, added it to a project, then deleted the project. The project has disappeared, but the site remains on the map and in the site list. I can click on the site and view site screens without crashing. If I create a new project and try to transfer it into the site, the site does not show up in the transfer list. After sync, the site disappears and now sync error arrises.

The fix eliminated the crash and sync error - was the expectation to address the site being visible after being deleted?

knipec:
@DerekCaelin Sorry this is merged to main but it's not in a build yet! I expect it'll be in ~~353~~ 354 once we make that

DerekCaelin:
This is resolved in build 359

PR #3321, opened 2026-07-22
fix: deleting a project should delete its sites on the client

knipec:
## Description
It seems we previously weren't deleting a project's sites on the client when we deleted the project. We also weren't doing anything to update the list of existing sites on the client based on the backend until the subsequent pull. This allowed users to open a site on the client that should have been deleted, which would lead to a crash, which we fixed with a workaround existence check. 

Now that the sites do get deleted on the client (and continue to be deleted on the backend), that workaround should no longer be needed.

### Related Issues
Addresses https://github.com/techmatters/terraso-mobile-client/issues/3314 

### Verification steps
Delete a project with sites. Make sure its sites are deleted in the app. Make sure when a pull happens the sites stay deleted.)


johannesparty (approved):
update backend to be careful if project delete mutation changes, you may need to change client code too.

PR #3333, opened 2026-07-28
fix: site note shows localized deleted user string

knipec:
## Description
- Show site note author attribute string as localized to the correct language
- Make "deleted user" lowercase
- Add deleted-user-specific strings because some languages don't do word ordering like english. Claude did the translations.

### Related Issues
https://github.com/techmatters/terraso-mobile-client/issues/3308#issuecomment-5106974719


dev-client/package.json
knipec, line 1: Reminder to update this after merging client-shared PR!

johannesparty (approved):
looks good (checks pending).

PR #3341, opened 2026-08-05
fix: open the site callout when a site is created from a temp location soil ID screen

### LINKED ISSUE #1428: Create Site button from Soil ID screen screen sends user to temporary location popover instead of site popover
## Description
The create Site button from Soil ID screen sends user to temporary location popover instead of site popover.

## Steps To Reproduce
1. Create a temporary site
2. On popover, tap "learn more"
3. Tap "explore the data"
4. Tap a soil, and then return to the soil id screen
5. Tap "Create site"
6. Fill out form and tap "Create"
7. See issue

## Expected behavior
See site location popover

## Actual behavior
Site is created, but see temporary location popover. 

https://github.com/user-attachments/assets/dbf8e981-6809-4766-8856-c6a522440008

Site appears not to be created?

knipec:
## Description
1. `pendingCalloutSiteId`: Keep track of the new site ID in app-level context (feels weird but couldn't find a better place to put it). SiteScreen now consumes it and opens the site callout if set.

Detail: In the buggy case, the new site was handed to SitesScreen by calling a method it publishes into a context-held ref, but react-freeze suspends any screen more than one level below the focused one and React detaches refs from suspended subtrees — so from three screens deep the ref was null and the optional-chained call quietly did nothing.
Now we hand the site id over through state that lives above the navigator instead, so it survives freezing, and let SitesScreen act on it once it renders again. Nothing needs to reach SitesScreen imperatively across screens now, so that
method comes off the ref entirely.

2. `collapseBottomSheet`: Refactor the other piece out of the old context-held ref

### Related Issues
Fixes https://github.com/techmatters/terraso-product/issues/1428


johannesparty (approved):
approved but only because you've tested too.

PR #3342, opened 2026-08-05
chore: update client-shared to ESM only

garobrik:
## Description
Updates `client-shared` to ESM-only, and adds necessary entries to Jest's config. This required mirroring the list from the integration test config into the unit test config. As far as I've researched this shouldn't cause any issues or need any extra config for the actual app build/runtime, but I will kick off a beta build to confirm before merging!

### Related Issues
Needed for https://github.com/techmatters/terraso-client-shared/pull/1556

### Verification steps
Everything should work the same!

### DIFF
--- a/dev-client/jest.integration.config.js
+++ b/dev-client/jest.integration.config.js
@@ -24,7 +24,7 @@ module.exports = {
     '<rootDir>/jest/integration/setup.ts',
   ],
   transformIgnorePatterns: [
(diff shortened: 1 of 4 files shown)


johannesparty (approved):
ok I approve, but note  you are pointing to feat/esm-module-output in terraso-client-shared.  I assume you'll land that and then adjust this to a named tag before merging, right?

PR #3344, opened 2026-08-06
fix: open site callout after creating site (approach 2)

### LINKED ISSUE #1428: Create Site button from Soil ID screen screen sends user to temporary location popover instead of site popover

knipec:
.. by routing the new site's id through navigation params to the SitesScreen.

This is an alternative to the context variables approach for the same bug as https://github.com/techmatters/terraso-mobile-client/pull/3341. Creating a site from the soil ID screen left the temporary-location callout on the map instead of the new site's card, because SitesScreen was reached through a ref that React detaches while the screen is frozen.

Carry the site id in the pop that returns to the map, so the tab navigator hands it to SitesScreen as a route param. Nothing crosses screens imperatively, so showSiteOnMap comes off the ref, and the pop now also selects the SITES tab for free.

The generated ParamList can't express optional params or a parent route that forwards {screen, params} into a nested navigator, so BOTTOM_TABS and SITES are declared by hand. That buys real type checking at the call site.

### Related Issues
Fixes https://github.com/techmatters/terraso-product/issues/1428

### DIFF
--- a/dev-client/src/context/SitesScreenContext.tsx
+++ b/dev-client/src/context/SitesScreenContext.tsx
@@ -17,10 +17,7 @@
 
 import {createContext, memo, RefObject, useContext, useRef} from 'react';
 
-import {Site} from 'terraso-client-shared/site/siteTypes';
-
 type SitesScreenRef = {
-  showSiteOnMap: (site: Site) => void;
   collapseBottomSheet: () => void;
 };
 
--- a/dev-client/src/navigation/types.ts
+++ b/dev-client/src/navigation/types.ts
@@ -16,6 +16,7 @@
  */
 import React from 'react';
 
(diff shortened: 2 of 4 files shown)


knipec:
Eh, both approaches have some weirdness to them. Closing this one because the other was more complete.

